PR #2, opened 2019-11-13
Change dependencies to use new repo path

JAORMX:
They were still pointing to jhrozek's repo. Lets switch it to the
OpenShift one.

PR #3, opened 2019-11-13
Use an init container to deliver the content

jhrozek:
Instead of baking the content into the scanner container, let's use a
separate init container with the content instead. This would allow us to
separate the scanner lifecycle from the content lifecycle and also allow
users to provide their own content.

The init container expects the content to be at the root of the
filesystem and all files matching *.xml would be copied from the init
container's / to emptyDir volume mounted at /content by the scanner.

The init container might be overriden with a new attribute of the CR:
    contentImage: string identifying the container image


jhrozek:
I rebased the PR to atop #2 so that we don't get conflicts when we push #2.

PR #8, opened 2019-11-14
Add verify make target (which also runs gosec)

JAORMX:
The verify target will run on CI and will do format checks, import
checks, as well as security static analysis on our code.


JAORMX:
/assign jhrozek

can you test this as well?

jhrozek:
LGTM, but can you run go mod tidy? Does it change go.mod for you in any way?

btw I tested this by running all the targets, so far they are silent (good).

JAORMX:
funky, go mod tidy removes gosec, what the hell

PR #10, opened 2019-11-15
Extend `clean` target to also clean go cache

JAORMX:
It was only cleaning the binaries, but now this cleans up the go cache
as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ OC?=oc
 SRC = $(shell find . -type f -name '*.go' -not -path "./vendor/*" -not -path "./_output/*")
 
 #.PHONY: all build clean install uninstall fmt simplify check run
-.PHONY: all operator-sdk build clean fmt simplify verify vet mod-verify gosec gendeepcopy test-unit run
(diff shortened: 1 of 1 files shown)


jhrozek:
/lgtm

I wonder if we should have separate Makefile targets for just cleaning the build artifact, but we can always split the target later..

JAORMX:
@jhrozek what do you think about leaving `clean` as is, and refactor. So there would also be `clean-build` and `clean-go`

jhrozek:
On Fri, Nov 15, 2019 at 06:23:43AM -0800, Juan Osorio Robles wrote:
> @jhrozek what do you think about leaving `clean` as is, and refactor. So there would also be `clean-build` and `clean-go`
> 

That sounds good to me. Would you prefer to have the PR merged or
refactored before merging?

JAORMX:
I'm almost done for the day, so, if you'll work a little longer today, feel free to add that bit. Else, we can wa

JAORMX:
@jhrozek up to you

PR #11, opened 2019-11-15
Run `go mod tidy` and upgrade ginkgo

JAORMX:
Ginkgo and gomega were updated in the go.mod file, so this reclects
that.

I also ran `go mod tidy`, which we observed removes the gosec
requirements.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,12 +6,16 @@ require (
 	github.com/go-logr/logr v0.1.0
 	github.com/go-logr/zapr v0.1.1
 	github.com/go-openapi/spec v0.19.0
-	github.com/onsi/ginkgo v1.8.0
-	github.com/onsi/gomega v1.5.0
+	github.com/onsi/ginkgo v1.10.3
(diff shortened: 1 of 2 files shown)


jhrozek:
I'm fine merging this, because `go mod tidy` does essentially the same thing on my machine. Just one question, how did gosec get into go mod in the first place? I thought it was correct to require it since it's used in the Makefile and also imported in `tools.go`.

JAORMX:
every time you run `make gosec` or `make verify` it gets added. Then if you build the artifact, or if you do `go mod tidy` it gets removed :/ I'm not sure why to be honest.

jhrozek:
On Fri, Nov 15, 2019 at 06:24:26AM -0800, Juan Osorio Robles wrote:
> every time you run `make gosec` or `make verify` it gets added. Then if you build the artifact, or if you do `go mod tidy` it gets removed :/ I'm not sure why to be honest.

alright, let's have go.mod in repo tidy..

PR #14, opened 2019-11-18
operator-sdk make

mrogers950:


Makefile
```diff
@@ -21,6 +21,8 @@ TEST_OPTIONS?=
 
 OC?=oc
 
+OPERATOR_SDK_URL=https://github.com/operator-framework/operator-sdk/releases/download/v0.12.0/operator-sdk-v0.12.0-x86_64-linux-gnu
```
jhrozek: I know this is WIP, but could we have the version in a separate variable so that we have a single place to change when we bump the version?

Makefile
```diff
@@ -36,14 +38,20 @@ SRC = $(shell find . -type f -name '*.go' -not -path "./vendor/*" -not -path "./
...
 all: build #check install
 
-build: fmt
-	operator-sdk build $(IMAGE_PATH)
+build: fmt operator-sdk
+	$(GOPATH)/bin/operator-sdk build $(IMAGE_PATH)
+
+operator-sdk:
+ifeq ("$(wildcard $(GOPATH)/bin/operator-sdk)","")
+	wget $(OPERATOR_SDK_URL) -O $(GOPATH)/bin/operator-sdk
```
jhrozek: Should we add an OR after the wget that tells the admin to install operator-sdk themselves and retry if the command fails for whatever reason?

jhrozek:
btw either this PR or PR#13 would have to be rebased depending on which of them merges first as PR#13 adds another invocation of operator-sdk.

Makefile
JAORMX: this would be nice

mrogers950:
@jhrozek @JAORMX should be good now.

PR #15, opened 2019-11-19
Generate OLM-related packages

JAORMX:
I basically ran:

    operator-sdk olm-catalog gen-csv --csv-version 0.0.1 --update-crds

An got this result.


jhrozek:
The patch looks non-intrusive and I have no objection merging it. As long as we're OK that it would change, the API as alpha after all and we're not "shipping" the operator yet.

PR #21, opened 2019-11-22
Use relevant images in e2e tests

JAORMX:
This allows us to take into use the operator image that was built from
CI. It leverages the already existing `$IMAGE_FORMAT` variable, in order
to get the appropriate path of the image to use.

On the other hand, when running locally this also enables the capability of building the image and pushing it to the cluster we're working on. Thus making running the e2e tests with a relevant image easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ OPERATOR_SDK_URL=https://github.com/operator-framework/operator-sdk/releases/dow
 SRC = $(shell find . -type f -name '*.go' -not -path "./vendor/*" -not -path "./_output/*")
 
 #.PHONY: all build clean install uninstall fmt simplify check run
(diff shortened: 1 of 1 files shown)


JAORMX:
Note that we need https://github.com/openshift/release/pull/6005 to merge to test this out.

jhrozek:
Even though the PR is on hold until we merge the CI plumbing, the new target can still be tested with:
```
make e2e IMAGE_FORMAT="foo/bar:\${component}"
```
which seems to do what it should do:
```
/home/jhrozek/.gvm/pkgsets/go1.13.4/global/bin/operator-sdk test local ./tests/e2e --image "foo/bar:compliance-operator" --namespace "openshift-compliance" --go-test-flags "-v"
```
-> ACK

JAORMX:
Failed to set up openshift cluster

JAORMX:
@jhrozek wooo!!! works!!

PR #28, opened 2019-11-27
e2e tests: Be resilient to temporary unavailability of k8s

JAORMX:
The tests are flaky due to leader re-elections in etcd. The following
error is logged every once in a while:

     single_scan_test.go:109: rpc error: code = Unavailable desc = etcdserver: leader changed

So lets try to be more resilient in our tests.


JAORMX:
/test e2e-aws

etcd failed again... but this time it was on the cleanup which is in the operator-sdk code.

JAORMX:
pushed https://github.com/operator-framework/operator-sdk/pull/2277 to avoid issues like this in the future.

PR #33, opened 2019-11-28
Use golang.org's gosumdb and proxy

JAORMX:
This speeds up the module fetching, which fails if we go to github
directly.


jhrozek:
As discussed on slack, relying on a proxy we don't control is not ideal, but it beats failing CI builds.
/lgtm

PR #35, opened 2019-11-29
Improve coverage of unit tests

JAORMX:
This adds two nodes to the unit tests to be taken into account, as well
as more cases.

### DIFF
--- a/pkg/controller/compliancescan/compliancescan_controller_test.go
+++ b/pkg/controller/compliancescan/compliancescan_controller_test.go
@@ -1,11 +1,16 @@
 package compliancescan
 
 import (
+	"context"
+	"fmt"
+
 	"github.com/go-logr/logr"
 	"github.com/go-logr/zapr"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


pkg/controller/compliancescan/compliancescan_controller_test.go
```diff
@@ -51,15 +70,49 @@ var _ = Describe("Testing compliancescan controller phases", func() {
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name: fmt.Sprintf("%s-%s-pod", compliancescaninstance.Name, nodeinstance2.Name),
+					},
+				})
+				// Set state to LAUNCHING
+				compliancescaninstance.Status.Phase = complianceoperatorv1alpha1.PhaseLaunching
+				reconciler.client.Status().Update(context.TODO(), compliancescaninstance)
+			})
+
+			It("should stay in LAUNCHING state", func() {
```
jhrozek: I actually wonder if you found a bug here. Because in `isPodRunningInNode` we only check if the pod exists based on the name and if the pod exists, we treat it as running. But, shouldn't we also check the owner reference and if it doesn't point to a scan, just go to an error state?

jhrozek:
Hey, let me know if you'd like me to merge this PR or if you think the owner reference should be checked.. Either way I'd not block this PR, but I thought agreeing on whether it's a bug or not might be a good idea before we merge the PR or else we forget.

JAORMX:
@jhrozek I think the namespace and predictable names for the pods are enough asurance about the ownership of the pods. Maybe checking the ownership is overkill in this case.

jhrozek:
Yup, as discussed on Slack, the namespace gives enough assurance.
/lgtm

PR #36, opened 2019-11-29
e2e: Test ComplianceScan filtered with nodeSelector

JAORMX:
Adds an e2e test that filters the compliance scan with a node selector. The test checks that the number of generated configmaps are the same as the number of nodes that are selected with the filter (in this case, the worker nodes).


jhrozek:
Looks good and passes CI.
/lgtm

PR #37, opened 2019-11-29
Clean up Makefile and add help target

JAORMX:
This makes our Makefile friendlier by organizing the variables, moving
the `.PHONY` declarations to be besides the target and adding a help
target.


jhrozek:
Looks good, CI passes and the make targets I tried work fine.
/lgtm

PR #38, opened 2019-11-29
e2e: Add test for scan with unexistent profile

JAORMX:


jhrozek:
Looks good and passes. Thank you.
/lgtm

PR #39, opened 2019-12-02
Add compliance scan result to status

JAORMX:
The result represents the result of the OpenSCAP scan. Which can be:
    
* non-compliance
* error
* compliant
    
This adds this as an extra value to the Status in the CRD.

### DIFF
--- a/deploy/crds/complianceoperator.compliance.openshift.io_compliancescans_crd.yaml
+++ b/deploy/crds/complianceoperator.compliance.openshift.io_compliancescans_crd.yaml
@@ -34,10 +34,6 @@ spec:
             content:
               type: string
             contentImage:
-              description: 'INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-                Important: Run "operator-sdk generate k8s" to regenerate code after
-                modifying this file Add custom validation using kubebuilder tags:
-                https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html'
               type: string
             nodeSelector:
               additionalProperties:
@@ -51,11 +47,13 @@ spec:
         status:
           description: ComplianceScanStatus defines the observed state of ComplianceScan
           properties:
+            errormsg:
+              type: string
             phase:
-              description: 'INSERT ADDITIONAL STATUS FIELD - define observed state
-                of cluster Important: Run "operator-sdk generate k8s" to regenerate
-                code after modifying this file Add custom validation using kubebuilder
-                tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html'
+              description: Represents the status of the compliance scan run.
+              type: string
+            result:
+              description: Represents the result of the compliance scan
               type: string
           type: object
       type: object
--- a/pkg/apis/complianceoperator/v1alpha1/compliancescan_types.go
+++ b/pkg/apis/complianceoperator/v1alpha1/compliancescan_types.go
@@ -6,7 +6,7 @@ import (
 
 // +genclient
 
-// ComplianceScanStatusPhase represents the status of the compliancescan run.
+// Represents the status of the compliance scan run.
 type ComplianceScanStatusPhase string
 
 const (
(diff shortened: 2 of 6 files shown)


JAORMX:
@jhrozek @mrogers950 What do you think?

deploy/crds/complianceoperator.compliance.openshift.io_compliancescans_crd.yaml
```diff
@@ -51,11 +47,13 @@ spec:
...
           description: ComplianceScanStatus defines the observed state of ComplianceScan
           properties:
+            faliurereason:
+              type: string
             phase:
-              description: 'INSERT ADDITIONAL STATUS FIELD - define observed state
-                of cluster Important: Run "operator-sdk generate k8s" to regenerate
-                code after modifying this file Add custom validation using kubebuilder
-                tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html'
+              description: Represents the status of the compliancescan run.
```
shawndwells:
```suggestion
              description: Represents the status of the compliance scan run.
```

pkg/apis/complianceoperator/v1alpha1/compliancescan_types.go
```diff
@@ -6,21 +6,35 @@ import (
...
+)
+
+// Represents the result of the compliance scan
+type ComplianceScanStatusResult string
+
+const (
+	// ResultCompliant represents the compliance scan having succeeded
+	ResultCompliant ComplianceScanStatusResult = "COMPLIANT"
+	// ResultFailure represents a compliance scan pod having failed to run the scan
+	ResultFailure ComplianceScanStatusResult = "FAILURE"
```
shawndwells: To indicate that an error occurred during the evaluation, could a different phrase than ``FAILURE`` be used?

The status result of ``FAILURE`` may lead many users/administrators/auditors to believe the system failed the compliance scan.

JAORMX: Thanks for the feedback! You're right, FAILURE might be confusing. Would ERROR be a better status?

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -276,6 +295,45 @@ func aContainerHasFailed(statuses []corev1.ContainerStatus) bool {
...
+	return complianceoperatorv1alpha1.ResultFailure, fmt.Errorf("Couldn't find '%s' container in '%s' pod status", OpenSCAPScanContainerName, p.Name)
+}
+
+func gatherResults(r *ReconcileComplianceScan, instance *complianceoperatorv1alpha1.ComplianceScan, nodes corev1.NodeList) (complianceoperatorv1alpha1.ComplianceScanStatusResult, error) {
+	var err error
+	var result complianceoperatorv1alpha1.ComplianceScanStatusResult
+	for _, node := range nodes.Items {
+		result, err = getScanResult(r, instance, &node)
+		if result != complianceoperatorv1alpha1.ResultCompliant {
+			// we output the last result if it was a failure
```
jhrozek: Would it be more defensive if we, instead of using the last non-compliant status used the 'strongest' one si that a failure would trump non-compliance?

JAORMX: Right, that would be a good idea. However, how do you output the strongest one? Another thing to note is that this iterates until it finds a non-compliant check. Which really would be all of the nodes, given that they should all have similar configurations (if we do a profile per node). So, there might really be no use in going through all of them, since one failure would in theory represent the whole set of nodes.

deploy/crds/complianceoperator.compliance.openshift.io_compliancescans_crd.yaml
JAORMX: Done, thanks for the suggestion!

pkg/apis/complianceoperator/v1alpha1/compliancescan_types.go
JAORMX: Changed it now. What do you think?

pkg/controller/compliancescan/compliancescan_controller.go
jhrozek: By the strongest one, I meant that a failure would override non-compliance. I agree that in theory the results should all be the same, though. So I'm not sure if it is needed to cherry-pick the status.

JAORMX: ah! I understand now... I'm honestly not sure. I'm open on this one. What do you think is best in this case? Error is should take precedence over non-compliance for the status?

JAORMX: I made errors take precedence. Any thoughts?

pkg/apis/complianceoperator/v1alpha1/compliancescan_types.go
shawndwells: @JAORMX yes, ``ERROR`` would be perfect.

jhrozek: Hmm, the constant name says `ResultError` but the constant value still says `FAILURE`, did you mean to change both maybe?

jhrozek:
Thank you, the new version looks very good to me. Just see the question about the constant value ERROR vs FAILURE, once we discuss that we can merge this PR.

pkg/apis/complianceoperator/v1alpha1/compliancescan_types.go
JAORMX: lol. oops!

JAORMX:
@jhrozek done

PR #40, opened 2019-12-03
Re-use same operator cluster to run e2e tests

JAORMX:
This tries to optimize the e2e test run by using the same operator for
each test run, instead of tearing everything up and down with each test.

This takes the form of wrapping each test to run the setup and teardown
as part of the helper code, and creating a structure to define the tests
in.


JAORMX:
Note that I tried making the tests parallel too, but that just made the whole thing slightly slower.

jhrozek:
very nice refactor again, thank you.
/lgtm

PR #45, opened 2019-12-04
Set podman as default runtime

JAORMX:
The current code uses flags that are only available in podman... so
setting that as the default.


jhrozek:
I agree that if our default actions use podman-specific options, the default runtime should be podman, too.
As a follow-up, we could warn that if RUNTIME==docker some actions might fail or are not supported,but given docker's uncertain future, I wouldn't even bother.
/lgtm

PR #46, opened 2019-12-06
Redirect output of `oc whoami`

JAORMX:
This is meant to check to if we'll need to rename kubeadmin. However,
given that it's positioned on the `ifeq` call. It gets called every
time.

This doesn't actually cause an error but it can be very distracting. So
lets get rid of the standard error of this command.


JAORMX:
/retest

aws test flake

PR #49, opened 2019-12-06
Move to vendor directory

JAORMX:
Seems folks are not too keen on using `go modules`, and vendor
directories are preferred in OpenShift.


mrogers950:
Just so we don't have to bother with periodic vendor bumps as we add stuff until our official release, we can hold off until then.

JAORMX:
@mrogers950 you're right. I'll abandon this and we can re-propose this then.

PR #51, opened 2019-12-09
While in RUNNING state Return to LAUNCHING state if pod is mising

JAORMX:
This will effectively attempt to launch the pods again if a pod is
missing. This ensures that all the nodes get a pod be scanned.


JAORMX:
@jhrozek @mrogers950 what do you think?

jhrozek:
/lgtm
The updates will be converted to deepcopy in a subsequent PR.

JAORMX:
/test e2e-aws
installer flake

PR #53, opened 2019-12-09
Do Makefile `all` target only build the code

JAORMX:
Currently the `all` target will build, verify and run the unit tests for
the project. We probably don't want that, since we use this target for
the go-build test. So lets make it only do builds.

We already have other tests that run the unit tests and verifications.


jhrozek:
Ack. It is easier to be able to iterate on code w/o disabling tests or such
/lgtm

PR #55, opened 2019-12-10
bump(*): Vendor dependencies

tnozicka:
This ensures offline builds and ensures the project requirements are met.

Also the project level golang builder is about to be switched to force vendoring mode.

/priority critical-urgent
/assign jhrozek JAORMX mrogers950


jhrozek:
/override aws-e2e
(Per @tnozicka 's request)

PR #56, opened 2019-12-10
Add ability to pass parameters to the e2e test run

JAORMX:
This allows us to pass flags to the `go test` that is called by the e2e
tests. This way, you can effectively call a specific test instead of
always running them all.

For instance, if you wanna run the test `TestScanWithNodeSelectorFiltersCorrectly`
you'd call:

    make e2e E2E_GO_TEST_FLAGS="-v -run TestE2E/TestScanWithNodeSelectorFiltersCorrectly"


jhrozek:
Does what it says on the box.
/lgtm

PR #60, opened 2019-12-12
Prepare makefile for vendor directory usage

JAORMX:
This adds the missing pieces for us to use the vendor directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ TAG?=latest
 # ===============
 CURPATH=$(PWD)
 TARGET_DIR=$(CURPATH)/build/_output
-GO=GO111MODULE=on go
+GO=GOFLAGS=-mod=vendor GO111MODULE=auto go
 GOBUILD=$(GO) build
 BUILD_GOPATH=$(TARGET_DIR):$(CURPATH)/cmd
 TARGET=$(TARGET_DIR)/bin/$(APP_NAME)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -69,7 +69,7 @@ image: fmt operator-sdk ## Build the compliance-operator container image
 	$(GOPATH)/bin/operator-sdk build $(IMAGE_PATH) --image-builder $(RUNTIME)
 
 .PHONY: build
-build: fmt ## Build the compliance-operator binary
```
jhrozek: Why was fmt removed?

build/Dockerfile
```diff
@@ -3,11 +3,7 @@ FROM registry.access.redhat.com/ubi8/go-toolset as builder
 
 WORKDIR /go/src/github.com/openshift/compliance-operator
 
-ARG ARG_GOPROXY="https://proxy.golang.org"
-ARG ARG_GOSUMDB="sum.golang.org"
-
-ENV GOPROXY=$ARG_GOPROXY
-ENV GOSUMDB=$ARG_GOSUMDB
+ENV GOFLAGS=-mod=vendor
```
jhrozek: Don't we also need GO111MODULE=auto here?

Makefile
JAORMX: because for some reason fmt is not taking into account the GOFLAGS... and it keeps trying to use the modules coming from the sources instead of taking into account the vendor dir.

build/Dockerfile
JAORMX: GO111MODULE=auto is kind of an overkill really. It already is the default.

Makefile
jhrozek: cool, because that's exactly the issue I was seeing with my scapresults patch :)

JAORMX:
/test e2e-aws
failed to launch cluster. unrelated error.

PR #67, opened 2019-12-16
Add interface for non-retriable errors

JAORMX:
This creates an error wrapper called `NonRetriableCtrlError` which
represents an error that's fatal and should not be retried on the
controller.

This changes the logic in the controllers to use this specific
NonRetriableCtrlError as opposed to the wrapper that was implemented
originally.

It also adds the logic of retry/no-retry for errors in the
compliancesuite controller.

### DIFF
--- a/pkg/controller/common/errors.go
+++ b/pkg/controller/common/errors.go
@@ -1,0 +1,54 @@
+package common
+
+import (
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// NonRetriableCtrlError wraps errors with the addition of having
+// the information for the error being non-retriable
+type NonRetriableCtrlError struct {
+	err      error
+	canRetry bool
+}
+
+func (cerr NonRetriableCtrlError) Error() string {
+	return cerr.err.Error()
+}
+
+// blank assignment to verify that RetriableCtrlError implements error
+var _ error = &NonRetriableCtrlError{}
+
+// IsRetriable exposes if the error is retriable or not
+func (cerr NonRetriableCtrlError) IsRetriable() bool {
+	return cerr.canRetry
+}
+
+// WrapNonRetriableCtrlError wraps an error with the RetriableCtrlError interface
+func WrapNonRetriableCtrlError(err error) *NonRetriableCtrlError {
+	return &NonRetriableCtrlError{
+		canRetry: false,
+		err:      err,
+	}
+}
+
+// IsRetriable returns whether the error is retriable or not using the
+// NonRetriableCtrlError interface
+func IsRetriable(err error) bool {
+	ccErr, ok := err.(*NonRetriableCtrlError)
+	if ok {
+		return ccErr.IsRetriable()
+	}
+	return true
+}
+
+// ReturnWithRetriableError will check if the error is retriable we return it.
+// If it's not retriable, we return nil so the reconciler doesn't keep looping.
+func ReturnWithRetriableError(log logr.Logger, err error) (reconcile.Result, error) {
+	if IsRetriable(err) {
+		log.Error(err, "Retriable error")
+		return reconcile.Result{}, err
+	}
+	log.Error(err, "Non-retriable error")
+	return reconcile.Result{}, nil
+}
--- a/pkg/controller/complianceremediation/complianceremediation_controller.go
+++ b/pkg/controller/complianceremediation/complianceremediation_controller.go
@@ -3,14 +3,10 @@ package complianceremediation
 import (
 	"context"
 	"fmt"
-	"github.com/go-logr/logr"
 	"sort"
 
 	ign "github.com/coreos/ignition/config/v2_2"
-
-	complianceoperatorv1alpha1 "github.com/openshift/compliance-operator/pkg/apis/complianceoperator/v1alpha1"
-	mcfgv1 "github.com/openshift/compliance-operator/pkg/apis/machineconfiguration/v1"
-	mcfgClient "github.com/openshift/compliance-operator/pkg/generated/clientset/versioned/typed/machineconfiguration/v1"
+	"github.com/go-logr/logr"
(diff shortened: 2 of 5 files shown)


pkg/controller/common/errors.go
```diff
@@ -0,0 +1,54 @@
+package common
+
+import (
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// NonRetriableCtrlError wraps errors with the addition of having
+// the information for the error being non-retriable
+type NonRetriableCtrlError struct {
```
jhrozek: Because everyone's favourite topic in software is naming things...I'm not super fond of `NonRetriableCtrlError`. It really sounds to me like the error can't ever be retried. Would something neutral like `ComplianceCtrlError` work?

Sorry for nitpicking.

jhrozek: btw all the method names sound good to me.

JAORMX: That was exactly the intent though. The error is non reyriable. And I only wrap errors that can't be retried.

pkg/controller/common/errors.go
```diff
@@ -0,0 +1,54 @@
...
+	return &NonRetriableCtrlError{
+		canRetry: false,
+		err:      err,
+	}
+}
+
+// IsRetriable returns whether the error is retriable or not using the
+// NonRetriableCtrlError interface
+func IsRetriable(err error) bool {
+	ccErr, ok := err.(*NonRetriableCtrlError)
```
jhrozek: Do you know if we could switch to go 1.13 in the project? This would allow us to simplify the code he a little. But it's really not a big deal and probably should be done in a follow-up patch (to also wrap errors propagating up with %w).

pkg/controller/common/errors.go
```diff
@@ -0,0 +1,54 @@
...
+	if ok {
+		return ccErr.IsRetriable()
+	}
+	return true
+}
+
+// ReturnWithRetriableError will check if the error is retriable we return it.
+// If it's not retriable, we return nil so the reconciler doesn't keep looping.
+func ReturnWithRetriableError(log logr.Logger, err error) (reconcile.Result, error) {
```
jhrozek: Good idea to put this into a function.

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -280,7 +278,9 @@ func (r *ReconcileComplianceSuite) reconcileScanRemediations(suite *complianceop
 			// All remediation lists in the scan must be equal
 			ok := diffRemediationList(scanRemediations, resultRemediations)
 			if !ok {
-				return fmt.Errorf("the list of remediations differs")
+				// TODO(jaosorior): Update suite status to reflect that there's an issue
+				// either with MCO or a compromise on a host.
```
jhrozek: Should we file a Jira or an Issue? (Yes, I am very guilty of just putting FIXMEs and TODOs in the code, too)

pkg/controller/common/errors.go
JAORMX: I tried but I didn't figure out what image to use. Go-toolset from red hat is still using 1.12

pkg/controller/compliancesuite/compliancesuite_controller.go
JAORMX: Let's open a jira

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -254,7 +251,8 @@ func (r *ReconcileComplianceSuite) reconcileScanRemediations(suite *complianceop
 		resultRemediations, err := parseResultRemediations(r, scan.Name, scan.Namespace, &cm, logger)
 		if err != nil {
 			logger.Error(err, "cannot parse scan result")
-			return err
+			// If the results are not parseable, we cannot recover from this.
+			return common.WrapNonRetriableCtrlError(err)
```
jhrozek: So earlier I said that in a future patch we might want to wrap errors with %w. This is one place where I think it would be nice so that the context is propagated up the stack.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -344,6 +314,7 @@ func deleteMachineConfig(r *ReconcileComplianceRemediation, name string, logger
 		logger.Info("MC to be deleted was already deleted")
 		return nil
 	} else if err != nil {
+		// delete error should be retried
```
jhrozek: Thank you for keeping and adding the comments. It shows that someone thought about the logic and it's not just copy-paste code.

pkg/controller/common/errors.go
jhrozek: OK

JAORMX:
@jhrozek answered to the name nit. What do you think?

pkg/controller/common/errors.go
jhrozek: Yeah, we talked on Slack about how you switched the logic to signaling that some errors can't be retried and I still think it's a good change. So with the naming you meant "Take an error and wrap it as non-retriable?" That makes more sense and also makes sense together with `IsRetriable` returning true on any other type than `NonRetriableCtrlError`.

Now I /really/ with we could use the new error mechanism because the code would be clearer if it said:
```
if errors.Is(err, NonRetriableCtrlError)  {
 // fail hard
}
```

JAORMX: True... But I didn't figure out how to use an image with go 1.13. At least not with go-toolset. Maybe if we switched to Fedora for the build image.

pkg/controller/compliancesuite/compliancesuite_controller.go
jhrozek: I amended the acceptance criteria of 306

jhrozek:
Looks good now, thanks.

PR #69, opened 2019-12-18
Include MachineConfigPool types, CRDs and generate the API

jhrozek:
This is the first step in adding better e2e tests for remediations. Because
we need to keep track of whether remediations were already applied and nodes
have been updated with MCs, we'll need to get MachineConfigPool types


jhrozek:
So this patch doesn't add the needed RBAC rules yet. I'm not sure it's needed because at first, until we use the API in the operator as well, this code would only be used in tests.

jhrozek:
I have pushed a patch that reports the scan status to the wrong branch by accident. The `remediations_demo` tag has been rebuilt, so either this run or the following should succeed.

PR #70, opened 2019-12-18
Two fixes in e2e tests so that we can fix our scanner image

jhrozek:
Our scanner image is currently broken in the sense that it does not report
non-compliance. Before we fix the scanner image, we need to make sure the
tests will keep passing.

The second patch fixes an embarassing issue where checks were not actually
checked..

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -231,12 +231,28 @@ func TestE2E(t *testing.T) {
 				}
 
 				// At this point, both scans should be non-compliant given our current content
-				scanResultIsExpected(f, namespace, workerScanName, complianceoperatorv1alpha1.ResultNonCompliant)
-				scanResultIsExpected(f, namespace, masterScanName, complianceoperatorv1alpha1.ResultNonCompliant)
+				err = scanResultIsExpected(f, namespace, workerScanName, complianceoperatorv1alpha1.ResultNonCompliant)
+				if err != nil {
+					return err
+				}
+
+				err = scanResultIsExpected(f, namespace, masterScanName, complianceoperatorv1alpha1.ResultNonCompliant)
+				if err != nil {
+					return err
+				}
+
 
 				// Each scan should produce two remediations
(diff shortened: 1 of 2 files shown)


jhrozek:
oops, testSuiteScan expects to find some issues, let me fix that

jhrozek:
Actually, let me just merge the scanner patch and regenerate the scanner. We need to find /some/ issues in that test.

What could be debated is whether it's good idea to check for exact name of remediations or if it would be better to check for a list of names we expect?

So now we have:
```
err = assertNumRemediations(f, suiteName, workerScanName, "worker", 2)
```
Maybe it would be better to have:
```
err = assertHasRemediations(f, suiteName, workerScanName, "worker", "remediation1")
err = assertHasRemediations(f, suiteName, workerScanName, "worker", "remediation2")
```
WDYT?

JAORMX:
> Actually, let me just merge the scanner patch and regenerate the scanner. We need to find /some/ issues in that test.
> 
> What could be debated is whether it's good idea to check for exact name of remediations or if it would be better to check for a list of names we expect?
> 
> So now we have:
> 
> ```
> err = assertNumRemediations(f, suiteName, workerScanName, "worker", 2)
> ```
> 
> Maybe it would be better to have:
> 
> ```
> err = assertHasRemediations(f, suiteName, workerScanName, "worker", "remediation1")
> err = assertHasRemediations(f, suiteName, workerScanName, "worker", "remediation2")
> ```
> 
> WDYT?
 
Right, I prefer testing for the specific remediations.

jhrozek:
OK, added another patch that checks the two expected remediations, any extra remediations are OK. This should be hopefully stable.

JAORMX:
helpers.go:45: The ComplianceScan Result wasn't what we expected. Got 'COMPLIANT', expected 'NON-COMPLIANT' 

huh

jhrozek:
I didn't merge the retval patch to the remediations_demo branch of the scanner I guess. Another reason to get rid of the script have all the logic in the operator.

PR #71, opened 2019-12-19
E2E tests: Add tests for remediations

jhrozek:
Adds a test that runs a scan, applies a remediation, checks the MC has
been created, runs the scan again to make sure that one finding is gone,
removes the MC to not leave things around and finally waits for the
nodes to come back up.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -3,12 +3,15 @@ package e2e
 import (
 	goctx "context"
 	"fmt"
-	"math/rand"
-	"testing"
-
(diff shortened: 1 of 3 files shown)


tests/e2e/e2e_test.go
```diff
@@ -235,32 +238,194 @@ func TestE2E(t *testing.T) {
...
+				t.Logf("Remediation found")
 
+				applyRemediation := func() error {
+					rem.Spec.Apply = true
+					err = f.Client.Update(goctx.TODO(), rem)
+					if err != nil {
+						t.Errorf("Cannot apply remediation")
+						return err
+					}
+					t.Errorf("Remediation applied")
```
JAORMX: you want to do Logf here.

PR #72, opened 2020-01-07
Consistently use constants across the codebase

jhrozek:
We copy-pasted string labels on some places and used constants with the
same value on other places. Let's consolidate the mess by:
    - defining the machineconfiguration.openshift.io/role label in the
      MC types go file
    - consistently using complianceoperatorv1alpha1.ScanLabel and
      complianceoperatorv1alpha1.SuiteLabel instead of their string
      values


jhrozek:
The new PR fixes a missing import

PR #74, opened 2020-01-10
Use deep-compare when checking if all remediations for a scan are the same

jhrozek:
We used to only check if the lenght of remediations for a scan (which
should map to a pool) is the same for all scan results. With this patch,
we also reflect and deep compare the remediations and their
machineconfig Specs and error out if one of the machines is out of step
with the rest.

### DIFF
--- a/pkg/controller/compliancesuite/compliancesuite_controller.go
+++ b/pkg/controller/compliancesuite/compliancesuite_controller.go
@@ -7,6 +7,8 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


jhrozek:
When I started working on this, I also wanted to populate the remediations with some real-world data based on scans. But then I realised, that it would be tricky to get different results in an e2e tests, so I opted for a unit test. On the other hand, in a unit test, I had no idea how to parse the results except for creating configMaps based on two huge XML files that would live in the source tree along with the tests..so in the end, the test is sort of minimal and does not check that the suite status is set to error.

Can anyone see a way to mock out a test for `reconcileScanRemediations` ?

jhrozek:
Added the WIP label because I forgot to report the error in the status

jhrozek:
Ugh, timeout:
```
Installing from release registry.svc.ci.openshift.org/ci-op-z0kpd3tk/release@sha256:c61ae2dc5c1e08a44c83b7987fc7b01688ccfbfb00498ed39aca4596e3f2e17c
level=info msg="Credentials loaded from the \"default\" profile in file \"/etc/openshift-installer/.awscred\""
level=warning msg="Found override for release image. Please be warned, this is not advised"
level=info msg="Consuming Install Config from target directory"
level=info msg="Creating infrastructure resources..."
level=error
level=error msg="Error: Error waiting for AMI (ami-034e58ff6181c7acf) to be ready: timeout while waiting for state to become 'available' (last state: 'pending', timeout: 40m0s)"
level=error
level=error msg="  on ../tmp/openshift-install-822946104/main.tf line 104, in resource \"aws_ami_copy\" \"main\":"
level=error msg=" 104: resource \"aws_ami_copy\" \"main\" {"
level=error
level=error
level=error
level=error msg="Error: [ERR]: Error building changeset: timeout while waiting for state to become 'accepted' (timeout: 5m0s)"
level=error
level=error msg="  on ../tmp/openshift-install-822946104/route53/base.tf line 33, in resource \"aws_route53_record\" \"api_external\":"
level=error msg="  33: resource \"aws_route53_record\" \"api_external\" {"
level=error
level=error
level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"
```

jhrozek:
same timeout again

PR #77, opened 2020-01-16
Use a configMap as the script source for the openscap container, not the container's entrypoint

jhrozek:
The openscap container uses a script baked into the container image that
represents the container's entry point. This patch instead uses a
configmap mounted into the container as an entrypoint and another
configmap with environment variables that fine-tune the entrypoint
script.

This has two advantages:
    1) All the logic is in one place, we no longer need to synchronize
    changes between the container image logic and the controller logic
    2) The openscap container image can be really dumb, just installing
    the scap scanner.

A follow-up patch will add another Dockerfile to the project so that we
can stop relying on the openscap container from a third party repository
and instead build another image from the operator repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 # Operator variables
 # ==================
 export APP_NAME=compliance-operator
+OPENSCAP_IMAGE_NAME=openscap-ocp
 
 # Container image variables
 # =========================
@@ -11,6 +12,8 @@ RUNTIME?=podman
 # or your e2e tests. This is overwritten if we bulid the image and push it to
 # the cluster or if we're on CI.
 IMAGE_PATH?=$(IMAGE_REPO)/$(APP_NAME)
+OPENSCAP_IMAGE_PATH=$(IMAGE_REPO)/$(OPENSCAP_IMAGE_NAME)
+OPENSCAP_DOCKERFILE_PATH=./images/openscap/Dockerfile
 
 # Image tag to use. Set this if you want to use a specific tag for building
 # or your e2e tests.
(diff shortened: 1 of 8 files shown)


jhrozek:
Marking as WIP because there would be another patch that builds the container image from the same repo and I think it makes sense if the two go together. At the same time, submitting now in case any of the other developers already has feedback.

jhrozek:
Looks like a flake:
level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"

jhrozek:
again the same:
```
level=error msg="Error: Error creating VPC Endpoint: VpcEndpointLimitExceeded: The maximum number of VPC endpoints has been reached."
level=error msg="\tstatus code: 400, request id: 8f83fd66-5a11-4266-9575-8a4fe8d0ea5a"
level=error
level=error msg="  on ../tmp/openshift-install-519943756/vpc/vpc.tf line 21, in resource \"aws_vpc_endpoint\" \"s3\":"
level=error msg="  21: resource \"aws_vpc_endpoint\" \"s3\" {"
level=error
level=error
level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"
```

jhrozek:
Huh, timed out again, this time waiting for a condition in the tests:
```
         e2e_test.go:349: Machines updated with remediation
        e2e_test.go:380: Second scan launched
        helpers.go:154: Waiting until scan test-recheck-remediations-workers-scan is done
        helpers.go:154: Waiting until scan test-recheck-remediations-workers-scan is done
        helpers.go:154: Waiting until scan test-recheck-remediations-workers-scan is done 
...
         helpers.go:49: timed out waiting for the condition 
```
This is the second suite the TestRemediate e2e test launches to check that remediations were actually taken into effect.
I'll just reschedule the test again as it works locally for me, let's see if the second run will finish.

It's also weird that this happens even though we have increased the test timeout to 60 minutes precisely to avoid issues like this.

JAORMX:
@jhrozek what timeout is failing? the overall timeout? or the request timeout? those are different. The overall timeout is 60min, which encompasses the whole test suite. The request timeout is the time per request of action is waited for; this one is less... 10 min I think.

jhrozek:
Rebased and slightly reworked. Now the patchset includes an additional patch to lookup the pod names through labels as well, which was causing yet another intermittent e2e test failure.
There's also a dockerfile added to build the openscap container. The dockerfile still relies on my copr repo, but I've verified that the RH dist-git branch for 8.2 already contains the binaries we need.

jhrozek:
oops, looking into the unit test failures

jhrozek:
Fixed. The unit tests were amended to also store the pod-Node mappings in a scan label.

PR #81, opened 2020-01-28
Add csv for OLM installation

JAORMX:
This adds the csv manifest for us to take the OLM into use to install
the operator. This way we can start pushing this for OperatorHub.


JAORMX:
@jhrozek @mrogers950 even if there are bits missing, we could merge this to start asking for guidance on how to include it in operatorhub.

PR #83, opened 2020-01-28
Move scapresults code into this repo

JAORMX:
This moves the scapresults code into the repo, renaming it to a more
general "resultscollector". It also pulls in the dependencies into the
vendor directory.


Makefile
```diff
@@ -1,7 +1,9 @@
 # Operator variables
 # ==================
 export APP_NAME=compliance-operator
+RESULTCOLLECTORBIN=resultcollector
 OPENSCAP_IMAGE_NAME=openscap-ocp
+RESULTSCOLLECTOR_IMAGE_NAME=$(RESULTCOLLECTORBIN)
```
mrogers950: having both RESULTCOLLECTOR and RESULTSCOLLECTOR var names makes it a little hard to read, lets use one of them for all related vars

PR #85, opened 2020-01-28
Use urn:xccdf:fix:script:ignition remediations instead of pretend ignition in bash

jhrozek:
There's been a bunch of commits recently to the ComplianceAsCode repo
that were either merged:
    https://github.com/ComplianceAsCode/content/commit/85587c4159fa316b404e74b0b1c77ac642a3d2b0
or are in progress:
    https://github.com/ComplianceAsCode/content/pull/5142

Those commits adds a new type of remediation that uses the type URN
"urn:xccdf:fix:script:ignition". This commits updates our ARF parser to
look for remediations of this type and points to a new content image for
e2e tests that require remediations.

### DIFF
--- a/pkg/utils/parse_arf_result.go
+++ b/pkg/utils/parse_arf_result.go
@@ -14,9 +14,7 @@ import (
 )
 
 const (
-	// FIXME: we pretend that shell is MC..
-	machineConfigFixType   = "urn:xccdf:fix:script:sh"
-	machineConfigFixPrefix = "apiVersion: machineconfiguration.openshift.io/v1"
+	machineConfigFixType = "urn:xccdf:fix:script:ignition"
 )
 
 func ParseRemediationsFromArf(scheme *runtime.Scheme, scanName string, namespace string, arf string) ([]*complianceoperatorv1alpha1.ComplianceRemediation, error) {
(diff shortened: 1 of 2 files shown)


isimluk:
LGTM AFAICT

jhrozek:
/hold
Putting this on hold as I need to regenerate the content used for the tests to use the new regex.

jhrozek:
huh, looks like the content is still wrong? Let me check everything locally and come back to this one

jhrozek:
Rebased atop the recent test changes so that we hopefully get a passing CI again..

JAORMX:
@jhrozek could you update this one? you need to update the references to the xml files.

jhrozek:
oops, earlier I forgot to change the big 18MB file to also use ignition, it was still using the pretend bash remediations. That's why the unit tests were failing.

jhrozek:
We had yet another race in the tests it seems. At the end of TestRemediate, we remove the MC to clean up after the test and in order for the other tests to be able to find remediations, but we don't wait for the nodes to reboot.

So I pushed a patch that does that, if it succeeds, I'll change the functions in the e2e tests so that we also have a helper for removing a MC as we do for applying and unapplying remediations.

jhrozek:
huh, same issue? Trying locally first..

jhrozek:
of course it didn't work, the last test doesn't use the proper content image *facepalm*

jhrozek:
Still quite slow parsing:
```
       Expected
          <float64>: 3.330466953
      to be <
          <float64>: 2
      /go/src/github.com/openshift/compliance-operator/pkg/utils/parse_arf_result_test.go:108 
```

jhrozek:
/test unit
I wonder if we need to bump the timeout yet again before we minimize the ARF?

PR #89, opened 2020-01-30
Add prefixes to the scans created by the compliancesuite

### LINKED ISSUE #84: Creating multiple compliancesuite CRs doesn't trigger a scan
If you trigger a scan by creating a compliancesuite CR, but leave it around, it prevents further scans from running with you create another compliancesuite CR.  It seems that this is caused by the pod names already existing for each node.

If you delete the original compliancescan CR first, it will then work as expected when you attempt to run another scan.  We should allow multiple scans to run, as keeping the results of previous scans for historical purposes may be desired.

JAORMX:
In order to make the names unique, this patch adds a prefix to the name
of the compliancescans created by the compliancesuite controller. This
way, we can have multiple suites that create scans with similar names
without conflict.

This closes #84

### DIFF
--- a/pkg/apis/complianceoperator/v1alpha1/complianceremediation_types.go
+++ b/pkg/apis/complianceoperator/v1alpha1/complianceremediation_types.go
@@ -3,8 +3,9 @@ package v1alpha1
 import (
 	"fmt"
 
-	mcfgv1 "github.com/openshift/compliance-operator/pkg/apis/machineconfiguration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	mcfgv1 "github.com/openshift/compliance-operator/pkg/apis/machineconfiguration/v1"
 )
 
 type RemediationApplicationState string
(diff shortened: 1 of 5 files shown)


JAORMX:
Closing this temporarily to avoid more rechecks

jhrozek:
This should now pass CI..
/retest

JAORMX:
I think there's still issues on this one. May have to fix the tests.

jhrozek:
According to the logs, we are creating scans with the following names:
```
{"level":"info","ts":1581358143.0307467,"logger":"controller_compliancescan","msg":"Phase: Done","Request.Namespace":"openshift-compliance","Request.Name":"test-suite-two-scans-test-suite-two-scans-workers-scan","ComplianceScan scan":"test-suite-two-scans-test-suite-two-scans-workers-scan"}
{"level":"info","ts":1581358153.6362312,"logger":"controller_compliancesuite","msg":"Reconciling ComplianceSuite","Request.Namespace":"openshift-compliance","Request.Name":"test-remediate"}
```

JAORMX:
@jhrozek that's the intent. The actual scans will have unique names, with the suite names prefixed. But we'll still display the status with the scan names the admin gave.

JAORMX:
/hold

Will address this later

PR #90, opened 2020-02-03
e2e: Check that the generation changed, not that it increased

JAORMX:
For MachineConfigPools we currently check that the generation always
increses; this assumption is not accurate, since, when we delete
MachineConfigs, it might go back to a previous version, so it would
decrease the generation.

Instead, lets check that the number of the generation changes.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -3,17 +3,18 @@ package e2e
 import (
 	goctx "context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


jhrozek:
This looks good, but would you like to also change the same issue at `e2e_test.go:466`?

jhrozek:
This is the code:
		```
		if mcOne.Generation != mcBoth.Generation {
					t.Errorf("Expected that the MC generation increases")
				}
```

jhrozek:
I've built and pushed content with the fixed regex. Let's get this past CI now...
/test e2e-aws

PR #91, opened 2020-02-04
Prepare to unbundle

isimluk:
I am trying to unbundle MachineConfig API from the Compliance Operator as these two started to diverge.

First roadblock, I have hit was the McRoleKey not available in the original MachineConfig. I did submit the variable on the MachineConfig side at openshift/machine-config-operator#1418 and here I rename the bundled variable in Compliance Operator so we get closer to the original.

We can wait for openshift/machine-config-operator#1418 being merged by we don't need to.

### DIFF
--- a/cmd/remediation-aggregator/main.go
+++ b/cmd/remediation-aggregator/main.go
@@ -252,8 +252,8 @@ func createRemediations(crClient *complianceCrClient, scan *complianceoperatorv1
 		}
 		rem.Labels[complianceoperatorv1alpha1.ScanLabel] = scan.Name
 		rem.Labels[complianceoperatorv1alpha1.SuiteLabel] = scan.Labels["compliancesuite"]
-		rem.Labels[mcfgv1.McRoleKey] = getScanRoleLabel(scan.Spec.NodeSelector)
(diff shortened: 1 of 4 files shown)


isimluk:
@jhrozek, @JAORMX, this can be now merged as https://github.com/openshift/machine-config-operator/pull/1418 was just merged and thus there are no doubts how the variable is going to be named in the MCO.

jhrozek:
Hi @isimluk sorry it took so long to get back to this PR. In the meantime, the code had changed a bit underneath. In particular, the logic that parsed and created the remediations was removed from the compliancesuite and into a new process called remediation-aggregator. So the code that uses McRoleKey can be just removed from the suite component and you can substitute the old constant for the new one in `cmd/remediation-aggregator/main.go` instead.

The commits that implemented the change are `72489333008208f13a7de918419cf410cdc57c74` and `8040a45091338b99e9ca08749f2ba2d845ac2bf5` mainly and some minor follow-up ones.

isimluk:
Amended. Please re-review. Thank You!

JAORMX:
thanks for getting back to this!

/lgtm

PR #93, opened 2020-02-05
Temporarily use bzip2 to decompress configmaps

jhrozek:
Because even gzipped configmaps can get over 1MB we would switch the
logcollector to use bzip2 temporarily. This is the matching change in
the operator. Also, the operator points to the image of the logcollector
that uses bzip2 for now.


jhrozek:
Last time it really looked like a super slow test run, I added some more log messages and bumped the timeout waiting for the remediation, let's see now..

PR #94, opened 2020-02-06
Add benchmark test

JAORMX:
This adds a benchmark test to help us identify the bottleneck in the XML parser.
    
Currently, the only package ran by the test is the pkg/utils package, since that's where the bottleneck lives. This is because benchmark tests can only be run on one package.
    
This also introduces the `make test-benchmark` target, which runs the aforementioned tests.

This does a couple of extra optimizations as well:

* Removing the test data from the packages and moving it to their own directory.
* Excluding the test code from the container builds
* Making the container build only build relevant binaries.

### DIFF
--- a/pkg/utils/parse_arf_result.go
+++ b/pkg/utils/parse_arf_result.go
@@ -2,13 +2,15 @@ package utils
 
 import (
 	"fmt"
-	complianceoperatorv1alpha1 "github.com/openshift/compliance-operator/pkg/apis/complianceoperator/v1alpha1"
-	mcfgv1 "github.com/openshift/compliance-operator/pkg/apis/machineconfiguration/v1"
+	"strings"
+
 	"github.com/subchen/go-xmldom"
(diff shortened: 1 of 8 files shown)


pkg/utils/parse_arf_result.go
```diff
@@ -29,15 +31,18 @@ func ParseRemediationsFromArf(scheme *runtime.Scheme, scanName string, namespace
 	// Get the checks that had failed
 	failedRuleResults := filterFailedResults(dom.Root.Query("//TestResult/rule-result"))
 
+	// Get group that contains remediations
+	relevantDom := dom.Root.Query("/report-requests/report-request/content/data-stream-collection/component/Benchmark")
```
jhrozek: This is a very nice improvement, but I would suggest to use dom.Root.QueryOne instead

JAORMX: apparently the regular query is more performant than QueryOne :/

PR #95, opened 2020-02-06
Make assertRemediations use global timeout

JAORMX:
Instead of using a hardcoded timeout variable, lets make that function
use our constants instead...

This also ups the timeout to 15 min


JAORMX:
/test e2e-aws

Installer error

jhrozek:
retesting so I can watch the test interactively in CI..
/test e2e-aws

jhrozek:
retesting so I can watch the test interactively in CI..
/test e2e-aws

PR #88, opened 2020-01-29
[WIP] Check for the node status as well

jhrozek:
It appears we might have a race in the e2e tests. Let's see if checking for the node status in addition to the MCP status helps here.

Also adds a forgotten error check

PR #99, opened 2020-02-07
Several patches to fix our CI: wait for CMs to appear, check for node status, fix predicates

jhrozek:
Let's see if our theory is correct

### DIFF
--- a/pkg/controller/compliancesuite/compliancesuite_controller.go
+++ b/pkg/controller/compliancesuite/compliancesuite_controller.go
@@ -6,9 +6,11 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"reflect"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/dsnet/compress/bzip2"
 	"github.com/go-logr/logr"
@@ -241,13 +243,27 @@ func (r *ReconcileComplianceSuite) reconcileScanRemediations(suite *complianceop
 	var scanRemediations []*complianceoperatorv1alpha1.ComplianceRemediation
 
 	// Look for configMap with this scan label
-	listOpts := client.ListOptions{
-		LabelSelector: labels.SelectorFromSet(labels.Set{"compliance-scan": scan.Name}),
-	}
+	err := wait.PollImmediate(5*time.Second, 10*time.Minute, func() (bool, error) {
+		listOpts := client.ListOptions{
+			LabelSelector: labels.SelectorFromSet(labels.Set{"compliance-scan": scan.Name}),
+		}
 
-	if err := r.client.List(context.TODO(), &cMapList, &listOpts); err != nil {
+		if err := r.client.List(context.TODO(), &cMapList, &listOpts); err != nil {
+			return false, err
+		}
+
+		if len(cMapList.Items) == 0 {
+			logger.Info("Scan has no results yet", "scan", scan.Name)
+			return false, nil
+		}
+		return true, nil
+	})
+
+	if err != nil {
+		logger.Error(err, "Error waiting for CMs to appear")
 		return err
 	}
+
 	logger.Info("Scan has results", "scan", scan.Name)
 
 	for _, cm := range cMapList.Items {
--- a/tests/e2e/helpers.go
+++ b/tests/e2e/helpers.go
@@ -313,7 +313,7 @@ func waitForMachinePoolUpdate(t *testing.T, mcClient *mcfgClient.Machineconfigur
 	}
 
 	// Should we make this configurable? Maybe 5 minutes is not enough time for slower clusters?
-	wait.PollImmediate(5*time.Second, 5*time.Minute, func() (bool, error) {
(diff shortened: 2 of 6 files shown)


jhrozek:
In case it's not clear from the title, this is just a way to prove a point. I don't think we should loop in the operator, I think the logcollector should only exit when the resources are created. But this is faster than touching the logcollector.

jhrozek:
Failed because:
```
could not copy stable imagestreamtag: Timeout: request did not complete within allowed duration
```

jhrozek:
OK, so this has helped somewhat, but we're not hitting the issue that we tried to avoid in PR #88. In the logs, I see that the scan that checks if remediations were applied ran and failed already here:
```
{"level":"info","ts":1581108752.170705,"logger":"controller_compliancesuite","msg":"The remediations differ between machines, this should never happen as the machines in a pool should be identical","Request.Namespace":"openshift-compliance","Request.Name":"test-recheck-remediations"}
```
as the message says, some nodes report different set of results than others. Looking at the timestap we see:
```
$ date -d @1581108752 --utc 
Fri 07 Feb 2020 08:52:32 PM UTC
```
But, in the MCO logs, we see that the rendered config that is the result of applying the remediation is this:
```
I0207 20:47:14.203986       1 render_controller.go:497] Generated machineconfig rendered-worker-9d773dd54c9dc531e608dc0ac8b53743 from 6 configs: [{MachineConfig  00-worker  machineconfiguration.openshift.io/v1  } {MachineConfig  01-worker-container-runtime  machineconfiguration.openshift.io/v1  } {MachineConfig  01-worker-kubelet  machineconfiguration.openshift.io/v1  } {MachineConfig  75-test-remediate-workers-scan-test-remediate  machineconfiguration.openshift.io/v1  } {MachineConfig  99-worker-d899ef7e-8f4f-40ad-8c09-200a8b845015-registries  machineconfiguration.openshift.io/v1  } {MachineConfig  99-worker-ssh  machineconfiguration.openshift.io/v1  }]
I0207 20:47:14.211639       1 render_controller.go:516] Pool worker: now targeting: rendered-worker-9d773dd54c9dc531e608dc0ac8b53743
```
But one of the worker nodes doesn't reach the rendered config until after the test had failed:
```
I0207 20:54:55.752495       1 node_controller.go:442] Pool worker: node ip-10-0-138-53.us-west-1.compute.internal has completed update to rendered-worker-9d773dd54c9dc531e608dc0ac8b53743
```

jhrozek:
I pushed the combined set of patches from this PR and #88, let's see..

JAORMX:
Could you rebase? a bunch of patches already merged

jhrozek:
Rebased and improved the log messages, logging the whole node object was too much.

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -241,13 +243,27 @@ func (r *ReconcileComplianceSuite) reconcileScanRemediations(suite *complianceop
 	var scanRemediations []*complianceoperatorv1alpha1.ComplianceRemediation
 
 	// Look for configMap with this scan label
-	listOpts := client.ListOptions{
-		LabelSelector: labels.SelectorFromSet(labels.Set{"compliance-scan": scan.Name}),
-	}
+	err := wait.PollImmediate(5*time.Second, 10*time.Minute, func() (bool, error) {
```
JAORMX: lets use the global timeout variables here.

tests/e2e/helpers.go
```diff
@@ -341,6 +340,41 @@ func waitForMachinePoolUpdate(t *testing.T, mcClient *mcfgClient.Machineconfigur
...
 
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func waitForNodesToBeReady(t *testing.T, f *framework.Framework) error {
+	err := wait.PollImmediate(10*time.Second, 15*time.Minute, func() (bool, error) {
```
JAORMX: the global variables are prefered.

jhrozek: done

pkg/controller/compliancesuite/compliancesuite_controller.go
jhrozek: this is the operator code, the purpose of the patch is to just check if it would help, in 'real' code I would fix this is the logcollector

jhrozek:
Looks like we didn't even get a cluster the last time around:
```
could not wait for template instance to be ready: failed to create objects: object is being deleted: pods "e2e-aws" already exists
```

jhrozek:
```
error: unable to read image registry.svc.ci.openshift.org/ci-op-ns6s5ii5/stable@sha256:e468f0ef83e6eca2366ec32b582eade7214b55d78ba60bc63e907173562fbba5: received unexpected HTTP status: 504 Gateway Time-out
```

jhrozek:
oops, s/labels/annotations/ and let's try yet again

jhrozek:
Heh:
```
xpected
          <float64>: 1.7403373850000001
      to be <
          <float64>: 1.7 
```
..so close :-)

jhrozek:
slow ARF again
/retest

jhrozek:
/test unit
arf parsing is slow again

PR #100, opened 2020-02-11
Port over bzip compression from the standalone scapresults repo

jhrozek:
This is an equivalent of:
    https://github.com/openshift/scapresults/pull/9


jhrozek:
btw I'm not sure why doesn't `go mod vendor` do anything?

JAORMX:
@jhrozek maybe it doesn't need to do anything. We already have most imports in the vendor directory, from before this patch.

PR #101, opened 2020-02-11
Add toleration to pod spec so we can scan master nodes too

JAORMX:
This toleration is needed and as missing.


jhrozek:
@JAORMX can you rebase atop the patch that ups the unit test timeout?

jhrozek:
tests passed, merging.

jhrozek:
/lgtm
for having a nice paper trail :)

PR #105, opened 2020-02-13
Move resultscollector Dockerfile to the top-level directory

jhrozek:
This fixes the "make resultscollector-image" target work. Previously, the
Dockerfile was executed in the context of the
"images/resultscollector" directory, so the COPY directive only copied what
was in that directory. And COPY doesn't support navigating outside the
build context.

The reason the target worked at all was that there was a typo in the 
RESULTSCOLLECTOR_DOCKERFILE_PATH variable, it was pointing to the openscap
scanner Dockerfile where we don't have the same issue as we only need the
files from the context to build the image.


jhrozek:
slow arf parser again, this time more than 6 seconds
/test-unit

JAORMX:
@jhrozek lets do https://github.com/openshift/compliance-operator/pull/106 instead.

jhrozek:
On Thu, Feb 13, 2020 at 06:29:49AM -0800, Juan Osorio Robles wrote:
> @jhrozek lets do https://github.com/openshift/compliance-operator/pull/106 instead.

I agree, that's better.

PR #108, opened 2020-02-14
ARF report server

JAORMX:
This is a small server that will listen for requests from all log
collectors and store the results in a PVC. this way, the results won't
take as much space as they used to.

### DIFF
--- a/cmd/resultscollector/main.go
+++ b/cmd/resultscollector/main.go
@@ -20,6 +20,8 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"net/http"
+	"net/http/httputil"
 	"os"
 	"path/filepath"
 	"time"
@@ -44,12 +46,13 @@ const (
 )
 
 type scapresultsConfig struct {
-	File          string
-	ScanName      string
-	ConfigMapName string
-	Namespace     string
-	Timeout       int64
-	Compress      bool
+	File            string
+	ScanName        string
+	ConfigMapName   string
+	Namespace       string
+	ResultServerURI string
+	Timeout         int64
+	Compress        bool
 }
 
 func defineFlags(cmd *cobra.Command) {
@@ -59,6 +62,7 @@ func defineFlags(cmd *cobra.Command) {
 	cmd.Flags().String("namespace", "Running pod namespace.", ".")
 	cmd.Flags().Int64("timeout", 3600, "How long to wait for the file.")
 	cmd.Flags().Bool("compress", false, "Always compress the results.")
+	cmd.Flags().String("resultserveruri", "", "The resultserver URI name.")
 }
 
 func parseConfig(cmd *cobra.Command) *scapresultsConfig {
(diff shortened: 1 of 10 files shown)


cmd/resultserver/main.go
```diff
@@ -0,0 +1,116 @@
...
+
+	if err := srvCmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func server(c *config) {
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		filename := r.Header.Get("X-Report-Name")
```
mrogers950: @JAORMX we might want to make this server handle both the pod payloads and the view requests. The pod payloads could be under a subpath, and we could even gate that separately with oauth-proxy if needed. Then we will only need one Service. 
The alternative to that is to keep the receiver and server services seperately, in which this part of the code should be named something like "result-receiver"

cmd/resultscollector/main.go
```diff
@@ -245,6 +255,39 @@ func main() {
...
 				os.Exit(1)
 			}
+			fmt.Println("Uploaded ConfigMap")
+
+			err = backoff.Retry(func() error {
+				url := scapresultsconf.ResultServerURI
+				fmt.Printf("Trying to upload to resultserver: %s\n", url)
+				reader := bytes.NewReader(contents)
+				client := &http.Client{}
+				req, err := http.NewRequest("POST", url, reader)
```
jhrozek: Does it make sense to check the error here?

cmd/resultserver/main.go
```diff
@@ -0,0 +1,116 @@
...
+	"os"
+	"path"
+
+	"github.com/spf13/cobra"
+)
+
+func defineFlags(cmd *cobra.Command) {
+	cmd.Flags().String("address", "127.0.0.1", "Server address")
+	cmd.Flags().String("port", "443", "Server port")
+	cmd.Flags().String("volume", "", "Volume name ??")
```
jhrozek: Is this currently unused?

cmd/resultscollector/main.go
JAORMX: uhm... perhaps not. The error would only be returned if we gave the wrong parameters.

cmd/resultserver/main.go
JAORMX: yep, leftover from the original implementation.

cmd/resultscollector/main.go
JAORMX: Fixed.

cmd/resultserver/main.go
JAORMX: Fixed.

cmd/resultserver/main.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		encoding := r.Header.Get("Content-Encoding")
+		extraExtension := encoding
+		if encoding != "" && encoding != "bzip2" {
+			log.Println("Rejecting. Invalid \"Content-Encoding\" header given.")
+			http.Error(w, "invalid content encoding header", 400)
+			return
+		} else if encoding == "bzip2" {
+			extraExtension = "." + extraExtension
+		}
```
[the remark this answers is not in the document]
JAORMX: The thing is, that an empty string is also valid. So I would need three if statements here to handle that.

deploy/operator.yaml
```diff
@@ -34,6 +34,10 @@ spec:
               value: "compliance-operator"
             - name: OPENSCAP_IMAGE
               value: "quay.io/jhrozek/openscap-ocp:remediations_demo"
+            - name: LOG_COLLECTOR_IMAGE
+              value: "quay.io/compliance-operator/resultscollector:latest"
+            - name: RESULT_SERVER_IMAGE
+              value: "quay.io/compliance-operator/resultserver:latest"
```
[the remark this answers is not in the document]
JAORMX: No, @mrogers950 was looking at that.

JAORMX:
retriggered CI, bootstrapping failed.

PR #111, opened 2020-02-17
Wait for ConfigMaps in compliancescan controller

JAORMX:
We used to just move to the DONE phase after getting the results from
the openscap container... this wasn't very practical for the components
depending on the results (e,g, the compliancesuite controller). So we
just wait for all of the configmaps to be ready in the AGGREGATING phase
of the compliancescan controller.

### DIFF
--- a/pkg/controller/compliancescan/compliancescan_controller.go
+++ b/pkg/controller/compliancescan/compliancescan_controller.go
@@ -3,6 +3,7 @@ package compliancescan
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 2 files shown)


JAORMX:
bootstrap failed

JAORMX:
> Just one question: Do you think that we still need to wait for the configMaps in the Suite controller after this patch? 
Probably not, let me remove that.
> We also seem to be polling for the configmaps in the collector, which seems more logical to me as there we want to retry for some transient errors,
can you elaborate?
> but perhaps the Suite controller could just requeue the reconcile loop?

jhrozek:
> > We also seem to be polling for the configmaps in the collector, which seems more logical to me as there we want to retry for some transient errors,
> 
> can you elaborate?

It seems like the controller should not give up on the first error, especially since we have RestartPolicyNever. If there was a network error or whatnot, we probably don't want to just fail the collector hard. So in the collector, I would personally leave the backoff-retry logic.

JAORMX:
> > > We also seem to be polling for the configmaps in the collector, which seems more logical to me as there we want to retry for some transient errors,
> > 
> > 
> > can you elaborate?
> 
> It seems like the controller should not give up on the first error, especially since we have RestartPolicyNever. If there was a network error or whatnot, we probably don't want to just fail the collector hard. So in the collector, I would personally leave the backoff-retry logic.

Got it. So that code is good as it is.

PR #113, opened 2020-02-18
Don't upload the ARF to ConfigMaps, instead use the DS and the XCCDF results separately

jhrozek:
The commit messages should hopefully provide enough context.

1) Add a new component, the CM aggregator

We've been using the full ARF report as produced by the openscap-chroot 
tool because it contains both the input DS and the test results. However,
as the DS gets too big, the ARF report would no longer fit into the 1MB
ConfigMap limit.

To address this, we're going to switch to only generating the XCCFD report
out of the ARF report and then generate the remediations based on a
(DataStream, XCCDF) pair. The openscap container would upload not the ARF,
but only the XCCDF which is significantly smaller.

Because the DataStream comes from the init container and generally can be 
user-provided, we need a new pod that processes the XCCDF reports together 
with the Data Stream. This patch adds the aggregator binary that would be 
running in this new pod, but the binary is not used yet as of this patch.

2) Make use of the aggregator

The previous patch added the aggregator binary, this one actually starts
using it. As of this patch, still the full ARF report is used, the only
functional change is that the reports are now generated in the scan
controller not the suite controller.

3) Don't upload the ARF to ConfigMaps, instead use the DS and the XCCDF results separately

Changes the shell script in the openscap component so that after the ARF is
generatd, it is also split into the XCCDF and the DS parts. The full ARF is
still being uploaded to the resultsserver, by the logcollector, but the
XCCDF is what's now being uploaded to the configmaps.

Provides a new utility function to parse the XCCDF,DataStream pair and
return remediations out of that. Removes the ARF parsing as it's no longer
used, along with the huge ARF test files.

### DIFF
--- a/cmd/remediation-aggregator/main.go
+++ b/cmd/remediation-aggregator/main.go
@@ -1,0 +1,405 @@
+/*
+Copyright © 2020 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"fmt"
+	"github.com/dsnet/compress/bzip2"
+	complianceoperatorv1alpha1 "github.com/openshift/compliance-operator/pkg/apis/complianceoperator/v1alpha1"
+	mcfgv1 "github.com/openshift/compliance-operator/pkg/apis/machineconfiguration/v1"
+	"github.com/openshift/compliance-operator/pkg/utils"
+	"github.com/spf13/cobra"
+	"io/ioutil"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"os"
+	"path/filepath"
+	"reflect"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sort"
+	"strings"
+	"time"
+)
+
+const (
+	configMapRemediationsProcessed = "compliance-remediations/processed"
+	configMapCompressed            = "openscap-scan-result/compressed"
+	nodeRolePrefix                 = "node-role.kubernetes.io/"
+)
+
+type aggregatorConfig struct {
+	Content   string
+	ScanName  string
+	Namespace string
+}
+
+type complianceCrClient struct {
+	client runtimeclient.Client
+	scheme *runtime.Scheme
+}
+
+func defineFlags(cmd *cobra.Command) {
+	cmd.Flags().String("content", "", "The path to the OpenScap content")
+	cmd.Flags().String("scan", "", "The compliance scan that owns the configMap objects.")
+	cmd.Flags().String("namespace", "openshift-compliance", "Running pod namespace.")
+}
+
+func parseConfig(cmd *cobra.Command) *aggregatorConfig {
+	var conf aggregatorConfig
+	conf.Content = getValidStringArg(cmd, "content")
+	conf.ScanName = getValidStringArg(cmd, "scan")
+	conf.Namespace = getValidStringArg(cmd, "namespace")
+	return &conf
+}
+
+func getValidStringArg(cmd *cobra.Command, name string) string {
+	val, _ := cmd.Flags().GetString(name)
+	if val == "" {
+		fmt.Fprintf(os.Stderr, "The command line argument '%s' is mandatory.\n", name)
+		os.Exit(1)
+	}
+	return val
+}
+
+func createCrClient(config *rest.Config) (*complianceCrClient, error) {
+	scheme := runtime.NewScheme()
+
+	v1.AddToScheme(scheme)
+	mcfgv1.AddToScheme(scheme)
+
+	scheme.AddKnownTypes(complianceoperatorv1alpha1.SchemeGroupVersion,
+		&complianceoperatorv1alpha1.ComplianceRemediation{})
+	scheme.AddKnownTypes(complianceoperatorv1alpha1.SchemeGroupVersion,
+		&complianceoperatorv1alpha1.ComplianceScan{})
+	scheme.AddKnownTypes(complianceoperatorv1alpha1.SchemeGroupVersion,
+		&metav1.CreateOptions{})
+
+	client, err := runtimeclient.New(config, runtimeclient.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return &complianceCrClient{
+		client: client,
+		scheme: scheme,
+	}, nil
+}
+
+func getScanRoleLabel(nodeSelector map[string]string) string {
+	if nodeSelector == nil {
+		return ""
+	}
+
+	// FIXME: should we protect against multiple labels and return
+	// an empty string if there are multiple?
+	for k := range nodeSelector {
+		if strings.HasPrefix(k, nodeRolePrefix) {
+			return strings.TrimPrefix(k, nodeRolePrefix)
+		}
+	}
+
+	return ""
+}
+
+func getScanConfigMaps(clientset *kubernetes.Clientset, scan, namespace string) ([]v1.ConfigMap, error) {
+	var cMapList *v1.ConfigMapList
+	var err error
+
+	// Look for configMap with this scan label
+	err = wait.PollImmediate(5*time.Second, 10*time.Minute, func() (bool, error) {
+		listOpts := metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("compliance-scan=%s", scan),
+		}
+
+		cMapList, err = clientset.CoreV1().ConfigMaps(namespace).List(listOpts)
+		if err != nil {
+			return false, err
+		}
+
+		if len(cMapList.Items) == 0 {
+			fmt.Printf("Scan %s has no results yet\n", scan)
+			return false, nil
+		}
+		return true, nil
+	})
+
+	if err != nil {
+		fmt.Printf("Error waiting for CMs of scan %s: %v\n", scan, err)
+		return nil, err
+	}
+
+	fmt.Printf("Scan %s has %d results\n", scan, len(cMapList.Items))
+	return cMapList.Items, nil
+}
+
+func readCompressedData(compressed string) (string, error) {
+	decoded, err := base64.StdEncoding.DecodeString(compressed)
+	if err != nil {
+		return "", err
+	}
+
+	compressedReader := bytes.NewReader(decoded)
+	bzReader, err := bzip2.NewReader(compressedReader, &bzip2.ReaderConfig{})
+	if err != nil {
+		return "", err
+	}
+	defer bzReader.Close()
+
+	// FIXME: probably unsafe, see https://haisum.github.io/2017/09/11/golang-ioutil-readall/
+	b, err := ioutil.ReadAll(bzReader)
+	if err != nil {
+		return "", err
+	}
+
+	return string(b), nil
+}
+
+func parseResultRemediations(scheme *runtime.Scheme, scanName string, namespace string, cm *v1.ConfigMap) ([]*complianceoperatorv1alpha1.ComplianceRemediation, error) {
+	var scanResult string
+	var err error
+
+	_, ok := cm.Annotations[configMapRemediationsProcessed]
+	if ok {
+		fmt.Printf("ConfigMap %s already processed\n", cm.Name)
+		return nil, nil
+	}
+
+	scanResult, ok = cm.Data["results"]
+	if !ok {
+		return nil, fmt.Errorf("no results in configmap %s", cm.Name)
+	}
+
+	_, ok = cm.Annotations[configMapCompressed]
+	if ok {
+		fmt.Printf("Results are compressed\n")
+		scanResult, err = readCompressedData(scanResult)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return utils.ParseRemediationsFromArf(scheme, scanName, namespace, scanResult)
+}
+
+// returns true if the lists are the same, false if they differ
+func diffRemediationList(oldList, newList []*complianceoperatorv1alpha1.ComplianceRemediation) bool {
+	if newList == nil {
+		return oldList == nil
+	}
+
+	if len(newList) != len(oldList) {
+		return false
+	}
+
+	sortMcSlice := func(mcSlice []*complianceoperatorv1alpha1.ComplianceRemediation) {
+		sort.SliceStable(mcSlice, func(i, j int) bool { return mcSlice[i].Name < mcSlice[j].Name })
+	}
+
+	sortMcSlice(oldList)
+	sortMcSlice(newList)
+
+	for i := range oldList {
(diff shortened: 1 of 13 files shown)


jhrozek:
btw the gosec warning I had to suppress in the aggregator is weird. I wouldn't think gosec would complain after we sanitize the patch? Nonetheless, as I wrote in the comment, I don't think it's an issue because the paths are not provided directly by users.

jhrozek:
/hold
the component is being renamed and I need to rebase Makefile changed atop a recent change

jhrozek:
/hold cancel
I ended up just renaming the aggregator to remediation-aggregator, I'll amend the Makefile when the image lands in the compliance-operator quay.io organisation.

cmd/remediation-aggregator/main.go
```diff
@@ -0,0 +1,426 @@
...
+		if err != nil {
+			fmt.Printf("Cannot parse CM %s into remediations, err: %v\n", cm.Name, err)
+		} else if cmRemediations == nil {
+			fmt.Println("Either no remediations found in result or result already processed")
+			continue
+		}
+		fmt.Printf("CM %s contained %d remediations\n", cm.Name, len(cmRemediations))
+
+		// Annotate the configmap, we want to avoid parsing it next time the reconcile
+		// loop hits
```
JAORMX: We wouldn't need this comment with a better function name. e.g. annotateParsedConfigMap

cmd/remediation-aggregator/main.go
```diff
@@ -0,0 +1,426 @@
...
+		if envvar := os.Getenv("KUBECONFIG"); len(envvar) > 0 {
+			kubeconfig = envvar
+		}
+		fmt.Printf("Using KUBECONFIG %s\n", kubeconfig)
+		config, err = clientcmd.BuildConfigFromFlags("", kubeconfig)
+		if err != nil {
+			fmt.Printf("The kubeconfig cannot be loaded: %v\n", err)
+			os.Exit(1)
+		}
+	}
```
JAORMX: Will this ever run outside the cluster though? I think you only need the InClusterconfig call, and fail if that doesn't work.

cmd/remediation-aggregator/main.go
```diff
@@ -0,0 +1,426 @@
...
+	cmCopy.Annotations[configMapRemediationsProcessed] = ""
+
+	_, err := clientset.CoreV1().ConfigMaps(cmCopy.Namespace).Update(cmCopy)
+	return err
+}
+
+func createRemediations(crClient *complianceCrClient, scan *complianceoperatorv1alpha1.ComplianceScan, remediations []*complianceoperatorv1alpha1.ComplianceRemediation) error {
+	for _, rem := range remediations {
+		fmt.Printf("Creating remediation %s\n", rem.Name)
+		if err := controllerutil.SetControllerReference(scan, rem, crClient.scheme); err != nil {
```
JAORMX: should the scan own the remediations? or should it be the suite? (legit question, I don't know).

cmd/remediation-aggregator/main.go
```diff
@@ -0,0 +1,426 @@
...
+
+	compressedReader := bytes.NewReader(decoded)
+	bzReader, err := bzip2.NewReader(compressedReader, &bzip2.ReaderConfig{})
+	if err != nil {
+		return "", err
+	}
+	defer bzReader.Close()
+
+	// FIXME: probably unsafe, see https://haisum.github.io/2017/09/11/golang-ioutil-readall/
+	b, err := ioutil.ReadAll(bzReader)
```
JAORMX: Do we need to read this already though? Why not pass the reader itself and just use it in the parsing function? that way we don't load the results in memory until they're needed.

pkg/utils/parse_arf_result.go
```diff
@@ -17,30 +17,34 @@ const (
 	machineConfigFixType = "urn:xccdf:fix:script:ignition"
 )
 
-func ParseRemediationsFromArf(scheme *runtime.Scheme, scanName string, namespace string, arf string) ([]*complianceoperatorv1alpha1.ComplianceRemediation, error) {
+func ParseRemediationFromContentAndResults(scheme *runtime.Scheme, scanName string, namespace string, content, results string) ([]*complianceoperatorv1alpha1.ComplianceRemediation, error) {
 	remediations := make([]*complianceoperatorv1alpha1.ComplianceRemediation, 0)
 
-	reader := strings.NewReader(arf)
-	dom, err := xmldom.Parse(reader)
+	resultsReader := strings.NewReader(results)
```
JAORMX: we wouldn't need this function if we would pass the reader directly. We could just pass this as the Reader interface, and not load the results in memory in the function that calls this.

cmd/remediation-aggregator/main.go
```diff
@@ -0,0 +1,426 @@
...
+	}, nil
+}
+
+func getScanRoleLabel(nodeSelector map[string]string) string {
+	if nodeSelector == nil {
+		return ""
+	}
+
+	// FIXME: should we protect against multiple labels and return
+	// an empty string if there are multiple?
```
JAORMX: This is an interesting question... let's leave it for now and discuss it later.

cmd/remediation-aggregator/main.go
```diff
@@ -0,0 +1,426 @@
...
+
+	return ""
+}
+
+func getScanConfigMaps(clientset *kubernetes.Clientset, scan, namespace string) ([]v1.ConfigMap, error) {
+	var cMapList *v1.ConfigMapList
+	var err error
+
+	// Look for configMap with this scan label
+	err = wait.PollImmediate(5*time.Second, 10*time.Minute, func() (bool, error) {
```
JAORMX: We probably don't need this Poll anymore, do we?

pkg/utils/parse_arf_result.go
```diff
@@ -17,30 +17,34 @@ const (
...
 		return nil, err
 	}
 
 	// Get the checks that had failed
-	failedRuleResults := filterFailedResults(dom.Root.Query("//TestResult/rule-result"))
+	failedRuleResults := filterFailedResults(resultsDom.Root.Query("//rule-result"))
 
-	// Get group that contains remediations
-	relevantDom := dom.Root.Query("/report-requests/report-request/content/data-stream-collection/component/Benchmark")
+	dsReader := strings.NewReader(content)
```
JAORMX: Same applies here. We could pass a Reader interface for the content.

cmd/remediation-aggregator/main.go
```diff
@@ -0,0 +1,426 @@
...
+	// this is not a security issue because none of the files are user-provided
+	cleanFileName := filepath.Clean(filename)
+	// #nosec G304
+	file, err := os.Open(cleanFileName)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+
+	b, err := ioutil.ReadAll(file)
```
JAORMX: Same applies here. We could merely pass the content Reader interface, instead of reading this all and loading the 20+MB in memory

cmd/resultscollector/main.go
```diff
@@ -219,36 +253,22 @@ func main() {
...
-				}
-				compressed = true
-				fmt.Printf("Compressed results are %d bytes in size\n", len(contents))
-			}
-
 			err = backoff.Retry(func() error {
 				url := scapresultsconf.ResultServerURI
 				fmt.Printf("Trying to upload to resultserver: %s\n", url)
-				reader := bytes.NewReader(contents)
+				reader := bytes.NewReader(arfContents.contents)
```
JAORMX: why not use the Reader directly instead of reading the ARF and then creating a Reader out of the string?

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -818,3 +843,101 @@ func getInitContainerImage(scanSpec *complianceoperatorv1alpha1.ComplianceScanSp
...
+					},
+					VolumeMounts: []corev1.VolumeMount{
+						{
+							Name:      "content-dir",
+							MountPath: "/content",
+						},
+					},
+				},
+			},
+			RestartPolicy: corev1.RestartPolicyNever,
```
JAORMX: perhaps this could restart if there's an error?

pkg/controller/compliancescan/config.go
```diff
@@ -57,8 +57,7 @@ cmd=(
     oscap-chroot $HOSTROOT xccdf eval \
     --fetch-remote-resources \
     --profile $PROFILE \
-    --report /tmp/scan-report.xml \
-    --results-arf /tmp/report.xml
+    --results-arf $ARF_REPORT
```
JAORMX: stupid question, what's the difference between the report and the results-arf?

jhrozek: --report is the XCCDF file that contains the selected tests and their results. For our clusters and the content we have it's typically less than a megabyte. OTOH --results-arf is the data stream with all the checks and remediations and the XCCDF together.
What would be better names? Would it be better to call them `--test-results` and `--arf-report`?

pkg/controller/compliancescan/compliancescan_controller.go
jhrozek: Yes, I was thinking about letting k8s handle errors as well. The reason I left PolicyNever was to not let k8s restart the pod indefinitely in case there was an unrecoverable error.

In the end we should think about what kinds of unrecoverable errors we might have. Of course that could be bugs (can't find the scan because it's mislabeled, wrong arguments passed to the aggregator), we can't do much about those and if we had a bug and a restartPolicy, then the pod would just loop. Then there are issues like can't create remediations, can't read the files. We could make them recoverable by logging an error and exiting the container gracefully (`os.Exit(0)`) so that the pod is spawned again. Then finally there is the case where remediations differ. We could also handle this in the aggregator by setting the scan status to done/error ourselves, but then I'm not sure if we want multiple components (not just the scan controller) to update the scan status.

tl;dr I'm not sure either, I'd welcome another opinion.

JAORMX: I think a more relevant thing is: can we detect if we have an unrecoverable error? If so, lets always restart, and if we detect that things are just not moving forward... we kill the pod and surface the error on the CR's status.

The status should only be updated by the controllers. If we get the workloads updating status, we would really get into a mess.

cmd/resultscollector/main.go
jhrozek: Because of the compression. I'm not sure there is a way to 'stream' the compression so that you would just have a reader that compresses what you read on the fly.
But if you think the interface would have been nicer if we had a reader returned (I do think it would be a bit nicer), then of course I can change the semantics to populate a reader in the `resultFileContents` struct. But, in case the file needs to be compressed, I don't think we'll get away with not reading the file.

JAORMX: Right, if it needs to be compressed we do need to read the file, that's true. But that would still reduce the number of reads we have to do in total. Feel free to leave this as-is, and we can optimize this later.

cmd/remediation-aggregator/main.go
jhrozek: OK, for the content it would work. But I'm afraid not for the results that we read from the configmap, there we just get a string from the CM. Anyway, passing a Reader makes for a nicer API. Fixed locally so far.

pkg/utils/parse_arf_result.go
jhrozek: Done locally.

cmd/remediation-aggregator/main.go
jhrozek: I guess not since one of your last patches, fixed locally

cmd/remediation-aggregator/main.go
jhrozek: We could always return an error from the aggregator and let the admin sort things out. Maybe failing deterministically is better than using a label based on the order.

pkg/utils/parse_arf_result.go
jhrozek: Fixed locally

cmd/remediation-aggregator/main.go
jhrozek: you're right, fixed locally

cmd/remediation-aggregator/main.go
jhrozek: I think a better thing to ask is: Should we support creating Remediations out of Scans or should we support going through the Suite only?

cmd/remediation-aggregator/main.go
jhrozek: I added this code for easier development initially so I can just create some objects without running all the scans and then run the binary locally. I removed the code now.

cmd/remediation-aggregator/main.go
jhrozek: OK, fixed

jhrozek:
I pushed new patches. Because there was quite a few review requests and github doesn't show interdiff, I pushed them so far as a separate patch. We can squash the changes later. I'll close those review comments that I think were addressed, I guess you can still click at the show outdated button..

jhrozek:
btw only now am I going to schedule a new local test run, hopefully I didn't screw up during the changes :)

pkg/controller/compliancescan/config.go
jhrozek: Would you like me to rename the command line switches?

jhrozek:
On Wed, Feb 19, 2020 at 01:37:35AM -0800, Juan Osorio Robles wrote:
> JAORMX requested changes on this pull request.
> 
> 
> 
> > +					Args: []string{
> +						"--content=" + absContentPath(scanInstance.Spec.Content),
> +						"--scan=" + scanInstance.Name,
> +					},
> +					SecurityContext: &corev1.SecurityContext{
> +						Privileged: &trueVal,
> +					},
> +					VolumeMounts: []corev1.VolumeMount{
> +						{
> +							Name:      "content-dir",
> +							MountPath: "/content",
> +						},
> +					},
> +				},
> +			},
> +			RestartPolicy: corev1.RestartPolicyNever,
> 
> I think a more relevant thing is: can we detect if we have an unrecoverable
> error? If so, lets always restart, and if we detect that things are just not
> moving forward... we kill the pod and surface the error on the CR's status.
> 

Hmm, is there a pattern for killing a pod after some timeout? I guess
the controller could spawn a goroutine, but is that a good idea?

> The status should only be updated by the controllers. If we get the workloads updating status, we would really get into a mess.

Right.

pkg/controller/compliancescan/config.go
JAORMX: up to you, I don't have a strong opinion about this.

cmd/remediation-aggregator/main.go
```diff
@@ -0,0 +1,389 @@
...
+
+	return nil
+}
+
+func readContent(filename string) (*os.File, error) {
+	// gosec complains that the file is passed through an evironment variable. But
+	// this is not a security issue because none of the files are user-provided
+	cleanFileName := filepath.Clean(filename)
+	// #nosec G304
+	return os.Open(cleanFileName)
```
JAORMX: By returning a file from `os.Open` straight, wouldn't this be reading from the file directly? Perhaps we need to wrap this into a buffered reader. So subsequent reads of the content would not have to interact with disk: https://golang.org/pkg/bufio/

cmd/remediation-aggregator/main.go
```diff
@@ -0,0 +1,389 @@
...
+		fmt.Printf("getScanConfigMaps failed: %v\n", err)
+		os.Exit(1)
+	}
+
+	contentFile, err := readContent(aggregatorConf.Content)
+	if err != nil {
+		fmt.Printf("Cannot read the content: %v\n", err)
+		os.Exit(1)
+	}
+	defer contentFile.Close()
```
JAORMX: Would you need to Seek(0) after each iteration?

pkg/controller/compliancescan/config.go
jhrozek: Ooh, only now did I notice that this is about the arguments for openscap, not for the pod. OK, in that case we can't remove the arguments :-) As per why didn't I use the output provided by `--report` in the first place and why I used the ARF -- the report still includes some data from the DS which we don't need. The rds-split was something Simon suggested on chat and only includes the rule results.

cmd/remediation-aggregator/main.go
jhrozek: Fixed

pkg/controller/compliancescan/compliancescan_controller.go
jhrozek: Actually, could we just always restart the pod and in the controller abort the scan if the pod reaches CrashLoopBackoff?

JAORMX: that sounds reasonable.

JAORMX:
/lgtm
/approve

This looks good as it is. Let's address the rest of the comments in subsequent PRs.

PR #114, opened 2020-02-18
Delete resultserver on DONE phase

JAORMX:
After gathering the results, we know that everything has been uploaded
(the configMaps and the ARF files), so we can safely delete the
resultserver.


jhrozek:
/lgtm
Although my cluster just expired so I won't be able to do any testing until later this evening.

jhrozek:
/lgtm
Tested manually with the help of a YAML snippet provided on Slack. I was still able to mount the PVC and ls the files.

PR #115, opened 2020-02-18
resultcollector: Run ConfigMap upload and ARF upload in parallel

JAORMX:
this uses goroutines to run the ConfigMap and ARF upload in parallel.


jhrozek:
This seems to work well even when I tried to inject an artificial error as if it was returned from backoff.Retry. The collector just printed the error text, there was no strange traceback complaining about the other goroutine (I /did/ place an error into both of them, just to make whichever finishes first error out..).

Nice work, we should merge this PR.

/lgtm

PR #122, opened 2020-02-21
Update the documentation

jhrozek:
Updates the README and the compliancesuite CR to match the current state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,25 +1,27 @@
 # compliance-operator
 
-The compliance-operator is a OpenShift/Kubernetes Operator that runs an
-openscap container in a pod on nodes in the cluster.
+The compliance-operator is a OpenShift Operator that allows an administrator
+to run compliance scans and provide remediations for the issues found. The
+operator leverages OpenSCAP under the hood to perform the scans.
 
-This repo is a POC for host openshift compliance detection and remediation
-and is work-in-progress.
+The operator runs in the `openshift-compliance` namespace, so make sure
+all namespaced resources like the deployment or the custom resources the
+operator consumes are created there.
 
-### Deploying:
+## Deploying the operator
+Before you can actually use the operator, you need to make sure it is
+deployed in the cluster.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,25 +1,27 @@
 # compliance-operator
 
-The compliance-operator is a OpenShift/Kubernetes Operator that runs an
-openscap container in a pod on nodes in the cluster.
+The compliance-operator is a OpenShift/Kubernetes Operator that allows and
```
mrogers950: Considering that we use the MCO I think it would be more accurate to only say OpenShift operator

README.md
```diff
@@ -28,88 +30,115 @@ run:
...
+and create the Kubernetes object:
+```
+# edit the Suite definition to your liking. You can also copy the file and edit the copy.
+$ vim deploy/crds/complianceoperator.compliance.openshift.io_v1alpha1_compliancesuite_cr.yaml
+# make sure to switch to the openshift-compliance namespace
+$ oc project openshift-compliance
+$ oc create -f deploy/crds/complianceoperator.compliance.openshift.io_v1alpha1_compliancesuite_cr.yaml
+```
+
+At this point the operator would pick up the ComplianceSuite custom resource,
```
mrogers950: nit: s/would pick up/ reconciles

README.md
```diff
@@ -28,88 +30,115 @@ run:
...
+create the scans for the suite, the scan would create a pod for every node
```
mrogers950: I might suggest different wording here.. "creates scan pods that run on each node in the cluster"

README.md
```diff
@@ -28,88 +30,115 @@ run:
...
+in the cluster, execute `openscap-chroot` on every node and eventually
+report the results. The scan would take several minutes to complete.
```
mrogers950: Any of the "would"s in the text we should change to present tense so we're describing the current behavior;  s/would take/takes

PR #124, opened 2020-02-24
Secure log-collector and result-server traffic with mTLS

mrogers950:
The traffic between the logcollector and result-server was operating over http. Configure both utilities to use https with mutual-TLS, to ensure that these two components may only communicate exclusively per-scan.


pkg/utils/x509.go
```diff
@@ -0,0 +1,28 @@
...
+func ComplianceOperatorRootCA(certname string, expire int) ([]byte, []byte, error) {
+	config, err := crypto.MakeSelfSignedCAConfig(certname, expire)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return config.GetPEMBytes()
+}
+
+func NewServerCert(caCert, caKey []byte, certname string, expire int) ([]byte, []byte, error) {
```
jhrozek: Do you plan on using the expire timeout? If a constant is to be used, do you think it makes sense to define it in the code?

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -144,6 +145,92 @@ func (r *ReconcileComplianceScan) Reconcile(request reconcile.Request) (reconcil
...
+	secret = &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "compliance-operator-root-ca",
+			Namespace: "openshift-compliance",
+		},
+		Data: map[string][]byte{
+			"ca.crt": cert,
+			"ca.key": key,
+		},
+	}
```
jhrozek: Do you think it would make sense to wrap creating the Secret in a function so that we don't have to leak the key and cert parameters to the main controller code?

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -144,6 +145,92 @@ func (r *ReconcileComplianceScan) Reconcile(request reconcile.Request) (reconcil
...
+	// Create the CA.
+	cert, key, err := utils.ComplianceOperatorRootCA("compliance-operator-root-ca", 100)
+	if err != nil {
+		return err
+	}
+
+	secret = &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "compliance-operator-root-ca",
+			Namespace: "openshift-compliance",
```
jhrozek: I know the namespace is probably not going to change, ever, but do you think it should be either defined globally or passed from the scan instance?

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -144,6 +145,92 @@ func (r *ReconcileComplianceScan) Reconcile(request reconcile.Request) (reconcil
...
+	}
+
+	// Create the CA secret.
+	err = r.client.Create(context.TODO(), secret)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+	// Do not create controlle references because for now, we want the CA to be created first scan, and then
+	// remain for subsequent scans. Might help with race conditions for server/client certs.
+	// Force rotate with a secret deletion.
```
jhrozek: So the rotation would be up to the admin?

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: I'd dig a global variable or a function.

... we already have a function for this: https://github.com/openshift/compliance-operator/blob/master/cmd/manager/main.go#L165

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: This is a good point to bring up. Rotation should be fairly simple (as Matt put out) and should be a matter of deleting the Secret. Should we have a process that does this daily or weekly as long as there isn't a scan running?

The certs that the CA gives out are only relevant in the context of scans (to upload results to the resultserver) which is deleted when the scan is in phase DONE.

jhrozek:
On Tue, Feb 25, 2020 at 04:16:47AM -0800, Juan Osorio Robles wrote:
> This is a good point to bring up. Rotation should be fairly simple (as Matt put out) and should be a matter of deleting the Secret. Should we have a process that does this daily or weekly as long as there isn't a scan running?
> 
> The certs that the CA gives out are only relevant in the context of scans (to upload results to the resultserver) which is deleted when the scan is in phase DONE.

Would it make sense to delete the secret once it's past, say 75% percent
of its lifetime and create a new one? I guess we'd have to either
reissue the certs for the scans of the old CA or just kill the old scans
or keep the CA around provisionally..

JAORMX:
> On Tue, Feb 25, 2020 at 04:16:47AM -0800, Juan Osorio Robles wrote: This is a good point to bring up. Rotation should be fairly simple (as Matt put out) and should be a matter of deleting the Secret. Should we have a process that does this daily or weekly as long as there isn't a scan running? The certs that the CA gives out are only relevant in the context of scans (to upload results to the resultserver) which is deleted when the scan is in phase DONE.
> Would it make sense to delete the secret once it's past, say 75% percent of its lifetime and create a new one? I guess we'd have to either reissue the certs for the scans of the old CA or just kill the old scans or keep the CA around provisionally..

We would kill the pods from old scans as soon as possible. So refreshing secrets wouldn't be necessary.

pkg/controller/compliancescan/compliancescan_controller.go
mrogers950: We can have the controller check the expiration and regenerate the CA, right now I just have a check for if it exists.  We could then use something short-lived, like around what we think might be the maximum for a single scan (1hr?). Then the admin doesn't need to know anything.

mrogers950:
The reason I thought to keep the same CA around for multiple runs is if there were ever a situation where a new scan's logcollector would be sending to an old scan's resultserver (the re-issued client would still authenticate to the CA of the server). If this should never be the case, then doing a brand new CA for each scan makes the most sense.

If it's _sometimes_ the case that it happens because of a race condition or whatever then we can make CAs with a validity that spans two or three runs, and auto-rotate with a validity check.

pkg/controller/compliancescan/compliancescan_controller.go
jhrozek: That sounds good to me. Regenerating the CA past some percentage would still probably be a good idea, because otherwise you might have a second scans scheduled 59mins after the first one and run out of the lifetime..

pkg/controller/compliancescan/compliancescan_controller.go
mrogers950: Addressed this

pkg/controller/compliancescan/compliancescan_controller.go
mrogers950: done, I moved these to helper functions in a different file.

pkg/controller/compliancescan/compliancescan_controller.go
mrogers950: I'm not sure if the new CA per scan or a limited use one would be better. For the moment, I updated to make it so there is now one CA per scan (cleaned up by CR reference) and they are valid for a day.  (The library-go helpers only give me a day as minimum) .

pkg/utils/x509.go
mrogers950: Done

mrogers950:
@JAORMX @jhrozek I've rebased and updated the PR, please take another look

mrogers950:
Made a mistake with the CA secret data keys, that's fixed, and the tests should be OK now.

pkg/controller/common/constants.go
```diff
@@ -0,0 +1,18 @@
...
+func init() {
+	ns, err := k8sutil.GetOperatorNamespace()
+	if err == nil {
+		ComplianceOperatorNamespace = ns
+	}
+}
+
+func GetComplianceOperatorNamespace() string {
+	return ComplianceOperatorNamespace
+}
```
JAORMX: if we have a function that gets this variable. Why is the variable exported? Let's make it private to this package and force folks t use the function instead.

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -144,6 +145,94 @@ func (r *ReconcileComplianceScan) Reconcile(request reconcile.Request) (reconcil
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileComplianceScan) handleRootCASecret(instance *complianceoperatorv1alpha1.ComplianceScan, logger logr.Logger) error {
+	caSecret := &corev1.Secret{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: RootCAPrefix + instance.Name, Namespace: common.GetComplianceOperatorNamespace()}, caSecret)
```
JAORMX: Let's make this a function.

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -144,6 +145,94 @@ func (r *ReconcileComplianceScan) Reconcile(request reconcile.Request) (reconcil
...
+	if err == nil {
+		// Already exists, so return.
+		// TODO: Check for expiration and re-create.
```
JAORMX: This depends... how long are we planning to keep the CA around?

We might want to consider deleting it in the DONE stage of the compliancescan. What do you think?

pkg/utils/x509.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"time"
+
+	"github.com/openshift/library-go/pkg/crypto"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+func ComplianceOperatorRootCA(certname string, expire int) ([]byte, []byte, error) {
+	config, err := crypto.MakeSelfSignedCAConfig(certname, expire)
```
JAORMX: Let's reach out to the crypto team... when using FIPS mode, is this function also forked to have usable defaults?

pkg/controller/compliancescan/compliancescan_controller.go
mrogers950: done

pkg/controller/common/constants.go
mrogers950: done.

PR #127, opened 2020-02-26
Make minor OLM release and enhance publishing workflow

JAORMX:
This releases version v0.1.1, which takes into use the new quay.io organization for the operator, as well as the latest code.

This also enhances the publish workflow to also generate the csv for the desired release. Note that the relevant versions need to be set up in the Makefile.


mrogers950:
@JAORMX excellent!
/lgtm

PR #129, opened 2020-02-27
Allow the restart of the scan/logcollector pod

JAORMX:
This makes it so that it's possible to restart the logcollector/scan pod in case there's an issue with the logcollector, the scan, or even an unforeseen error such as an OOM.

This is done by:

* Adding an exit_status file to the scan that skips doing subsequent scans if it exists.
* Marking "AlreadyExists" as a non-error for the logcollector (this ensures we can run it again).
* Pushing the result to the ConfigMap isntead of reading it from the container itself

### DIFF
--- a/cmd/resultscollector/main.go
+++ b/cmd/resultscollector/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/dsnet/compress/bzip2"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -49,6 +50,8 @@ const (
 type scapresultsConfig struct {
 	ArfFile         string
 	XccdfFile       string
+	ExitCodeFile    string
+	CmdOutputFile   string
 	ScanName        string
 	ConfigMapName   string
 	Namespace       string
@@ -60,6 +63,8 @@ type scapresultsConfig struct {
 func defineFlags(cmd *cobra.Command) {
 	cmd.Flags().String("arf-file", "", "The ARF file to watch.")
 	cmd.Flags().String("results-file", "", "The XCCDF results file to watch.")
+	cmd.Flags().String("exit-code-file", "", "A file containing the oscap command's exit code.")
+	cmd.Flags().String("oscap-output-file", "", "A file containing the oscap command's output.")
 	cmd.Flags().String("owner", "", "The compliance scan that owns the configMap objects.")
 	cmd.Flags().String("config-map-name", "", "The configMap to upload to, typically the podname.")
 	cmd.Flags().String("namespace", "openshift-compliance", "Running pod namespace.")
@@ -72,6 +77,8 @@ func parseConfig(cmd *cobra.Command) *scapresultsConfig {
 	var conf scapresultsConfig
 	conf.ArfFile = getValidStringArg(cmd, "arf-file")
 	conf.XccdfFile = getValidStringArg(cmd, "results-file")
+	conf.ExitCodeFile = getValidStringArg(cmd, "exit-code-file")
+	conf.CmdOutputFile = getValidStringArg(cmd, "oscap-output-file")
 	conf.ScanName = getValidStringArg(cmd, "owner")
 	conf.ConfigMapName = getValidStringArg(cmd, "config-map-name")
 	conf.Namespace = getValidStringArg(cmd, "namespace")
@@ -197,7 +204,7 @@ func readResultsFile(filename string, timeout int64, doCompress bool) (*resultFi
 	return &rfContents, nil
 }
 
-func getConfigMap(owner *unstructured.Unstructured, configMapName, filename string, contents []byte, compressed bool) *corev1.ConfigMap {
+func getConfigMap(owner *unstructured.Unstructured, configMapName, filename string, contents []byte, compressed bool, exitcode string) *corev1.ConfigMap {
 	annotations := map[string]string{}
 	if compressed {
 		annotations = map[string]string{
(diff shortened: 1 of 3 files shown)


cmd/resultscollector/main.go
```diff
@@ -255,64 +263,103 @@ func main() {
 				os.Exit(1)
 			}
 
-			arfContents, err := readResultsFile(scapresultsconf.ArfFile, scapresultsconf.Timeout, scapresultsconf.Compress)
-			if err != nil {
-				fmt.Println(err)
-				os.Exit(1)
-			}
+			exitcode, err := readResultsFile(scapresultsconf.ExitCodeFile, scapresultsconf.Timeout, false)
```
jhrozek: unchecked err?

cmd/resultscollector/main.go
```diff
@@ -255,64 +263,103 @@ func main() {
...
 					}
-					fmt.Println(string(bytesresp))
-					return nil
-				}, backoff.WithMaxRetries(backoff.NewExponentialBackOff(), maxRetries))
-
+					fmt.Println("Uploaded ConfigMap")
+					wg.Done()
+				}()
+				wg.Wait()
+			} else {
```
jhrozek: The logic looks good, but instead of:
```
if ok {
 ok_stuff()
} else {
 error_stuff()
}
```
Could we do:
```
if !ok {
   error_stuff()
}
ok_stuff()
```
This is to keep the usual codepath to the left. Alternatively, if there is some shared code that needs to be executed in both cases, it would at least be nice to extract bits of code into functions. The current code is getting long and hard to read.

pkg/controller/compliancescan/config.go
```diff
@@ -70,9 +76,10 @@ cmd+=($CONTENT)
 # move the results file when the command is done so the log collector
 # picks up the whole thing and not a partial file
 echo "Running oscap-chroot as ${cmd[@]}"
-"${cmd[@]}"
+"${cmd[@]}" > $REPORT_DIR/cmd_output
 rv=$?
 echo "The scanner returned $rv"
+echo "$rv" > $REPORT_DIR/exit_code
```
jhrozek: Are you sure it's safe to add the exit code after the scanner has exited? At this point, we haven't split the ARF and the XCCDF result yet..

mrogers950:
One thing I noticed is a race between the logcollector pods and result server. I had a result server for a scan sit in Pending for longer than the logcollecgtor allowed time, it still resulted in a DONE status and erroring scan pods. (The move to DONE state kills the result server eventually, the logcollector pods are left timing out). Will this help?

pkg/controller/compliancescan/config.go
JAORMX: Right, gotta move this till later then. Good call

JAORMX:
@mrogers950 yep! This would help sinc eot now waits till all the pods are done.

mrogers950:
I tested this fixes the race by merging in #124 and introduced a tls error. The erroring pods restarted and retried the log-collector upload, while the scanStatuses remained RUNNING. The resultserver pods waited around as well which is good. However, with this hard error the pods continued to restart and status remained in RUNNING indefinitely. So we should allow for restarts but be able to limit them and eventually fail out

JAORMX:
> I tested this fixes the race by merging in #124 and introduced a tls error. The erroring pods restarted and retried the log-collector upload, while the scanStatuses remained RUNNING. The resultserver pods waited around as well which is good. However, with this hard error the pods continued to restart and status remained in RUNNING indefinitely. So we should allow for restarts but be able to limit them and eventually fail out

so, was this an unrecoverable error?

cmd/resultscollector/main.go
```diff
@@ -226,105 +233,187 @@ func getConfigMap(owner *unstructured.Unstructured, configMapName, filename stri
...
 	}
 }
 
+func uploadToResultServer(arfContents *resultFileContents, scapresultsconf *scapresultsConfig) error {
+	return backoff.Retry(func() error {
+		url := scapresultsconf.ResultServerURI
+		fmt.Printf("Trying to upload to resultserver: %s\n", url)
+		reader := bytes.NewReader(arfContents.contents)
+		client := &http.Client{}
+		req, _ := http.NewRequest("POST", url, reader)
```
jhrozek: Why do we ignore the error here?

cmd/resultscollector/main.go
```diff
@@ -226,105 +233,187 @@ func getConfigMap(owner *unstructured.Unstructured, configMapName, filename stri
...
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	if len(exitcodeContent.contents) < 1 {
+		fmt.Println("exitcode file was empty")
+		os.Exit(1)
+	}
+
+	return string(exitcodeContent.contents[0])
```
jhrozek: Why are you trimming the results? I would understand trimming whitespace, but we check the results for correctness anyway..

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -549,6 +510,19 @@ func gatherResults(r *ReconcileComplianceScan, instance *complianceoperatorv1alp
 		if err != nil {
 			isReady = false
 		}
+
+		result, err = getScanResult(r, foundCM)
```
jhrozek: Could you please add a comment to either the function definition or here to the call that err can only ever be set if result is ResultError? Alternatively, we could check for err and if it is set, then we could always return ResultError.

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -463,79 +461,42 @@ func isPodRunning(r *ReconcileComplianceScan, podName, namespace string, logger
...
 
-func aContainerHasFailed(statuses []corev1.ContainerStatus, logger logr.Logger, podname string) bool {
-	for _, status := range statuses {
-		if status.State.Terminated != nil {
-			if status.State.Terminated.ExitCode != 0 {
-				logger.Info("container failed in pod",
-					"pod", podname, "container", status.Name,
-					"exit-code", status.State.Terminated.ExitCode)
-				return true
+func getScanResult(r *ReconcileComplianceScan, cm *corev1.ConfigMap) (complianceoperatorv1alpha1.ComplianceScanStatusResult, error) {
```
jhrozek: The `r` parameter is unused.

cmd/resultscollector/main.go
JAORMX: Cause this would only return an error in case we would set up an invalid url name, an invalid method or an invalid reader. And those are hardcoded values here.

The real request is done in line 254.

cmd/resultscollector/main.go
JAORMX: I merely am removing the newline. Given that what we care for in the exitcode is just a number (which always fits in a byte, since it's UTF-8), I just get that byte, turn it into a string, and use it.

This code had issues before because the value was, for instance, "1\n" instead of just "1"

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: nice! Let me fix it.

jhrozek:
I tested the logcollector restart by just modifying the logcollector code to always return 1. The pods then went into a CrashLoopBackOff. I wonder if we should, in some subsequent PR, count the number of restarts of the containers and if they get above some threshold just set the scan status to error?

JAORMX:
> I tested the logcollector restart by just modifying the logcollector code to always return 1. The pods then went into a CrashLoopBackOff. I wonder if we should, in some subsequent PR, count the number of restarts of the containers and if they get above some threshold just set the scan status to error?

@mrogers950 @jhrozek I don't think we should have special handling of unrecoverable errors. We should merely try to converge to the state that's marked in the Custom Resource and that's it. If the error is temporary, we'll eventually converge, and the workload will eventually work. If the error is unrecoverable, the issue will manifest itself in more critical workloads than this.

I think the best we can and should do is surface such errors in the status in an earlier stage. Perhaps we should try to do that in the RUNNING phase. But that's it. This keeps the logic simple and reduces tech-debt.

jhrozek:
Ozz would create a card to discuss the unrecoverable errors. In the meantime, let's merge this PR and iterate on it
/lgtm

PR #131, opened 2020-02-28
e2e: Test for errormsg being set in the scan status

JAORMX:
We were just testing for the error result, but never noticed if the
errormsg was being set or not. This simply checks if the errormsg is
empty or not, and if it isn't (and we were expecting an error), it'll
fail the test.


JAORMX:
uhm... the /lgtm tag wasn't added by the bot. That's weird.

jhrozek:
Oi, bot!

/lgtm

PR #135, opened 2020-03-02
Add build directories to .dockerignore

JAORMX:
This ensures that we don't copy any binaries that we might have
previously built. Thus making the COPY operation lighter.


jhrozek:
Images still build and the patch makes sense in general, so..

/lgtm

PR #139, opened 2020-03-04
Delete scan/logcollector pods & PKI secrets when the scan is done

JAORMX:
This removes the pods & secrets to clean up the namespace once the scan is done.

As a side-effect of deleting the PKI secrets, it means that we don't have to deal with cert-rotation, given that we generate a new PKI per-scan.

### DIFF
--- a/pkg/controller/compliancescan/compliancescan_controller.go
+++ b/pkg/controller/compliancescan/compliancescan_controller.go
@@ -145,94 +145,6 @@ func (r *ReconcileComplianceScan) Reconcile(request reconcile.Request) (reconcil
 	return reconcile.Result{}, nil
 }
 
-func (r *ReconcileComplianceScan) handleRootCASecret(instance *complianceoperatorv1alpha1.ComplianceScan, logger logr.Logger) error {
-	exist, err := secretExists(r.client, RootCAPrefix+instance.Name, common.GetComplianceOperatorNamespace())
-	if err != nil {
-		return err
-	}
-	// TODO: Maybe check for expiration and re-create.
-	if exist {
-		return nil
-	}
-
-	logger.Info("creating CA", "instance", instance.Name)
-	secret, err := makeCASecret(instance, common.GetComplianceOperatorNamespace())
-	if err != nil {
-		return err
-	}
-
-	// Create the CA secret.
-	err = r.client.Create(context.TODO(), secret)
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return err
-	}
-
-	// TODO: Rather than a controller reference, delete during DONE phase.
-	if err = controllerutil.SetControllerReference(instance, secret, r.scheme); err != nil {
-		return err
-	}
-
-	return nil
-}
-
-func (r *ReconcileComplianceScan) handleResultServerSecret(instance *complianceoperatorv1alpha1.ComplianceScan, logger logr.Logger) error {
-	exist, err := secretExists(r.client, ServerCertPrefix+instance.Name, common.GetComplianceOperatorNamespace())
-	if err != nil {
-		return err
-	}
-	if exist {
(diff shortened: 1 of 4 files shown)


pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -443,13 +330,171 @@ func (r *ReconcileComplianceScan) phaseAggregatingHandler(instance *complianceop
...
+	if err = controllerutil.SetControllerReference(instance, secret, r.scheme); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *ReconcileComplianceScan) deleteRootCASecret(instance *complianceoperatorv1alpha1.ComplianceScan, logger logr.Logger) error {
+	logger.Info("deleting CA", "instance", instance.Name)
+	secret, err := makeCASecret(instance, common.GetComplianceOperatorNamespace())
```
mrogers950: These functions do all the key generation and stuff, it would be better to just get the names

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -443,13 +330,171 @@ func (r *ReconcileComplianceScan) phaseAggregatingHandler(instance *complianceop
...
+		return err
+	}
+
+	// Create the CA secret.
+	err = r.client.Create(context.TODO(), secret)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+
+	if err = controllerutil.SetControllerReference(instance, secret, r.scheme); err != nil {
```
mrogers950: If we're deleting them should we still set ownership?

JAORMX: We shouldn't. But this function creates it.

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: Right. It would have been better to have a function to get the names. I'll fix that.

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -443,13 +330,155 @@ func (r *ReconcileComplianceScan) phaseAggregatingHandler(instance *complianceop
...
+	logger.Info("deleting CA", "instance", instance.Name)
+	ns := common.GetComplianceOperatorNamespace()
+	secret := certSecret(getCASecretName(instance), ns, []byte{}, []byte{}, []byte{})
+	return r.deleteSecret(secret)
+}
+
+func (r *ReconcileComplianceScan) deleteResultServerSecret(instance *complianceoperatorv1alpha1.ComplianceScan, logger logr.Logger) error {
+	logger.Info("deleting server cert", "instance", instance.Name)
+	ns := common.GetComplianceOperatorNamespace()
+	secret := certSecret(getClientSecretName(instance), ns, []byte{}, []byte{}, []byte{})
```
mrogers950: Should be using getServerSecreName()

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -443,13 +330,155 @@ func (r *ReconcileComplianceScan) phaseAggregatingHandler(instance *complianceop
...
+	return r.deleteSecret(secret)
+}
+
+func (r *ReconcileComplianceScan) deleteResultClientSecret(instance *complianceoperatorv1alpha1.ComplianceScan, logger logr.Logger) error {
+	logger.Info("deleting client cert", "instance", instance.Name)
+	ns := common.GetComplianceOperatorNamespace()
+	secret := certSecret(getCASecretName(instance), ns, []byte{}, []byte{}, []byte{})
```
mrogers950: and getClientSecretName here

PR #143, opened 2020-03-05
complianceremediation: Don't reuse loop variable

jhrozek:
The go spec says in:
    https://golang.org/ref/spec#For_range
that:
    The iteration variables may be declared by the "range" clause using a
    form of short variable declaration (:=). In this case their types are
    set to the types of the respective iteration values and their scope is
    the block of the "for" statement; they are re-used in each iteration.

What was unclear to me when I wrote the original code was that also the
address of the range variable is reused. This lead to the
appliedRemediations slice being populated with copies of the last
iteration of the range variable.

### DIFF
--- a/pkg/controller/complianceremediation/complianceremediation_controller.go
+++ b/pkg/controller/complianceremediation/complianceremediation_controller.go
@@ -224,25 +224,25 @@ func getAppliedMcRemediations(r *ReconcileComplianceRemediation, rem *compliance
 	}
 
 	appliedRemediations := make([]*mcfgv1.MachineConfig, 0, len(scanSuiteRemediations.Items))
-	for _, candidate := range scanSuiteRemediations.Items {
-		if candidate.Spec.Type != complianceoperatorv1alpha1.McRemediation {
+	for i := range scanSuiteRemediations.Items {
+		if scanSuiteRemediations.Items[i].Spec.Type != complianceoperatorv1alpha1.McRemediation {
 			// We'll only merge MachineConfigs
 			continue
 		}
-		if candidate.Status.ApplicationState != complianceoperatorv1alpha1.RemediationApplied {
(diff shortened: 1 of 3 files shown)


JAORMX:
@jhrozek do you think it would be relevant to add a test for this?

jhrozek:
Yes, but let me check how many rules we have currently in the content. I found this while working on adding new rules to the content and hit this when mass-applying ~40 rules at the same time. Our current tests only rely on two rules which is why we haven't hit the issue.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -242,7 +242,8 @@ func getAppliedMcRemediations(r *ReconcileComplianceRemediation, rem *compliance
 		}
 
 		// OK, we've got an applied MC, add it to the list
-		appliedRemediations = append(appliedRemediations, &candidate.Spec.MachineConfigContents)
+		ar := candidate.Spec.MachineConfigContents
```
mrogers950: Alternately you can use the indexing form of range to avoid reuse, eg: https://play.golang.org/p/xd7rBklWFaZ . (there's a good argument in favor of this form across the project to avoid accidents)

jhrozek: I like this better, thank you for the suggestion.

jhrozek:
/hold

I'm marking this as on hold because my cluster expired and it's too late in the day to wait for a new one :-) but more importantly because I'd like to add a test.

jhrozek:
Taken Matt's suggestion about the loop variable into account and added a test. Going forward, the test could be extended to test more that the function we touched so that we could have better unit test coverage for the whole remediations controller, but for now, we have covered what the patch had changed..

PR #144, opened 2020-03-06
Get updated CRDs

JAORMX:
Now that the MCO API is imported via a go module, we can generate the
CRDs with the current APIs. This is the result.


jhrozek:
/test images
I this this must be fluke..

jhrozek:
Actually, before this merges, don't you need to also run "make csv" to also update the OLM CRDs?

jhrozek:
As discussed on Slack, the OLM files would be regenerated when we release.
/lgtm

PR #145, opened 2020-03-06
Add `debug` flag to compliancescan

JAORMX:
This disables cleanup of resources when in the DONE phase. This way, you
can go back to the pods and check what happened.


jhrozek:
This looks good, but do you want to also run "make csv" to propagate the change to the OLM CRDs?

jhrozek:
Works as expected now.
/lgtm

PR #146, opened 2020-03-06
Replace generate openapi command with crds

JAORMX:
The openapi command was deprecated in operator-sdk; instead, lets
replace it with the generate crds call, which updates the CRDs.


jhrozek:
/lgtm
As discussed on Slack, there is an issue due to a bug in MCO, but we can either work around it locally or commit the change to vendor/. Either way, nothing that concerns this PR.

PR #147, opened 2020-03-06
compliancescan: Make debug=true imply running openscap with --verbose INFO

jhrozek:
It might be handy to get some introspection to what the scanner is
doing. Therefore when the scan is set in debugging mode, the
oscap-chroot command would also use --verbose=INFO. The scan output is
also tee(1)-d to stdout and the cmd_output file.

### DIFF
--- a/pkg/controller/compliancescan/config.go
+++ b/pkg/controller/compliancescan/config.go
@@ -21,11 +21,12 @@ const (
 	OpenScapEnvConfigMapName = "openscap-env-map"
 
 	// environment variables the default script consumes
-	OpenScapHostRootEnvName  = "HOSTROOT"
-	OpenScapProfileEnvName   = "PROFILE"
-	OpenScapContentEnvName   = "CONTENT"
-	OpenScapReportDirEnvName = "REPORT_DIR"
-	OpenScapRuleEnvName      = "RULE"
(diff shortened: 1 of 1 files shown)


jhrozek:
There's two questions with this PR:
1. If it's OK to hardcode the INFO level. I think yes, because the DEBUG output is overly verbose.
2. The output would now be part of the cmd_output which is then uploaded to the configmap unconditionally uncompressed. This could in theory overflow the configmap, because running the whole suite with increased verbosity could produce a lot of data. But since the cmd_output is only uploaded if the openscap command fails, I'm not sure if this is a big issue.

JAORMX:
We should really address cmd_output compression before meeting this

jhrozek:
Welp, this really does break `TestE2E/TestScanWithInvalidContentFails`

jhrozek:
OK, so the problem was with the `| tee`. This had caused the return value to be the return value of the whole pipe and because tee almost always succeeds, we would always return 0 as if the scanner succeeded. Let's try it again with `set -o pipefail`.

jhrozek:
So the pipefail was also a bad idea since the tee wouldn't have been executed in case the first part of the pipe failed. So it's back to the low-tech solution of just cat-ing the command output..

PR #148, opened 2020-03-07
Update description of API types

JAORMX:
This makes sure that the appropraite descriptions are set in the CRDs,
and will eventually show up in the `explain` command's output.

### DIFF
--- a/pkg/apis/complianceoperator/v1alpha1/compliancescan_types.go
+++ b/pkg/apis/complianceoperator/v1alpha1/compliancescan_types.go
@@ -37,10 +37,24 @@ const (
 // ComplianceScanSpec defines the desired state of ComplianceScan
 // +k8s:openapi-gen=true
 type ComplianceScanSpec struct {
-	ContentImage string            `json:"contentImage,omitempty"`
-	Profile      string            `json:"profile,omitempty"`
-	Rule         string            `json:"rule,omitempty"`
-	Content      string            `json:"content,omitempty"`
+	// Is the image with the content (Data Stream), that will be used to run
+	// OpenSCAP.
+	ContentImage string `json:"contentImage,omitempty"`
+	// Is the profile in the data stream to be used. This is the collection of
+	// rules that will be checked for.
+	Profile string `json:"profile,omitempty"`
+	// A Rule can be specified if the scan should check only for a specific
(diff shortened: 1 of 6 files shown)


pkg/apis/complianceoperator/v1alpha1/compliancescan_types.go
```diff
@@ -37,10 +37,22 @@ const (
...
+	// OpenSCAP.
+	ContentImage string `json:"contentImage,omitempty"`
+	// Is the profile in the data stream to be used. This is the collection of
+	// rules that will be checked for.
+	Profile string `json:"profile,omitempty"`
+	// A Rule can be specified if the scan should check only for a specific
+	// rule. Note that when leaving this empty, the scan will check for all the
+	// rules for a specific profile.
+	Rule string `json:"rule,omitempty"`
+	// Is the path to the file that contains the content (the data stream)
```
jhrozek: Should we say that the init container by default copies the content to `/content`? Should we say that if the path is relative, then the scan pod would produce a path based at `/content`?

jhrozek:
Actually I guess I can answer to myself I guess, it looks like `https://github.com/openshift/origin/blob/master/test/extended/cli/explain.go` has the CRDs hardcoded?

JAORMX:
> Actually I guess I can answer to myself I guess, it looks like `https://github.com/openshift/origin/blob/master/test/extended/cli/explain.go` has the CRDs hardcoded?

I actually didn't know this. Let me make the change manually there and see if it works now.

pkg/apis/complianceoperator/v1alpha1/compliancescan_types.go
JAORMX: Yes, we should mention the specifics of this. Specially if we have requirements. IMO we don't need to mention anything about the `/content` directory, but we should instead mention that the path is relative to the container's `/` directory.

jhrozek: OK, that works for me. I was actually thinking that we might want to drop the `if path is relative then path := /content+path` logic as it might be confusing for people unless it's documented.

jhrozek: Resolved in PR #151

PR #149, opened 2020-03-07
legal: add Apache 2.0 license

JAORMX:
Similar to k8s and other openshift projects, this adds the Apache 2.0 license to the project.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
jhrozek: Should we fill this up?

jhrozek:
Another semi-random though: The fact that Go creates static binaries means that we need to make sure our dependencies are MIT licensed or under a similar permissive license. I did a quick check:
```
find vendor -name LICENSE | xargs grep -i gnu 
```
and it seems that none are e.g. pure GPL libraries which would have been a problem. But it's something to keep in mind for the future..

LICENSE
JAORMX: lol, I actually let github do this commit automatically, seems I gotta modify it.

JAORMX: Seems other operators have this same text... I wonder if that's an issue.

jhrozek:
Thank you, this works for me.
/lgtm

PR #162, opened 2020-03-11
Remove hardcoded namespace references where possible

JAORMX:
This makes it so that we can deploy the compliance operator in any namespace when doing it from a CSV.

Only leaving the one required reference.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -2,7 +2,6 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: compliance-operator
-  namespace: openshift-compliance
 spec:
   replicas: 1
   selector:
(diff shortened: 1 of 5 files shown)


jhrozek:
The content in these patches is good in itself, but what do you think about the usage of `GetComplianceOperatorNamespace()`? There we always return a constant unconditionally..

jhrozek:
By the way, if you do end up doing the change, then it might make sense to wait until PR#157 merges as it would conflict with those changes. I just lgtm-ed #157.

jhrozek:
> > The content in these patches is good in itself, but what do you think about the usage of `GetComplianceOperatorNamespace()`? There we always return a constant unconditionally..
> 
> It should return whatever namespace the operator is running on: https://github.com/openshift/compliance-operator/blob/master/pkg/controller/common/constants.go#L10-L12
> 
> So it's not hardcoded, it merely defaults to `openshift-compliance`

Thank you, I missed the init() function.

/lgtm

JAORMX:
Had to rebase.

PR #166, opened 2020-03-12
Make csv target idempotent

JAORMX:
As a Makefile would usually have as a good practice... if a directory or
file already exists, you probably don't need to try the command again
and again. So this makes the csv command idempotent by checking that the
directory of the target version already exists.

This also adds docs for the target, so it appears in the help target
too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -325,7 +325,9 @@ ifndef COURIER_PACKAGE_VERSION
 endif
 
 .PHONY: csv
-csv: check-package-version operator-sdk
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -325,7 +325,9 @@ ifndef COURIER_PACKAGE_VERSION
 endif
 
 .PHONY: csv
-csv: check-package-version operator-sdk
+csv: deploy/olm-catalog/compliance-operator/$(COURIER_PACKAGE_VERSION) check-package-version operator-sdk ## Generate the CSV and packaging for the specific version (NOTE: Gotta specify the version with the COURIER_PACKAGE_VERSION environment variable)
+
+deploy/olm-catalog/compliance-operator/$(COURIER_PACKAGE_VERSION):
```
jhrozek: Would it make sense to amend the clean target so that it removes this artifact?

JAORMX: Not really. Doing that would make our API releases dissappear. Ultimately we do need a policy to clean up older CSVs.

PR #167, opened 2020-03-12
Rely on predictable pod names instead of labels

JAORMX:
Given that we're using sha-1 to generate unique pod names for the scan
instances, we can rely on the pod name to be predictable. So we don't
need to store this in a label as we used to.


JAORMX:
Closing in favor of https://github.com/openshift/compliance-operator/pull/171

PR #170, opened 2020-03-13
Don't persist the ConfigMaps in labels

JAORMX:
We are using predictable names, so we don't need to store these values
in labels. The downside of doing this is that labels are user-facing
parameters, and we don't want to expose inner functionality here.


JAORMX:
Closing in favor of https://github.com/openshift/compliance-operator/pull/171

JAORMX:
Closing in favor of https://github.com/openshift/compliance-operator/pull/171

PR #174, opened 2020-03-13
Fix getting previous csv version

JAORMX:
It used to base the previous version on the last touched file in the
CSV versions directory... This might very between people's computers (I
might have temporarily modified a CSV) so this is not reliable. Instead,
lets just use alphabetical order.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ COURIER_PACKAGE_NAME=compliance-operator-bundle
 COURIER_OPERATOR_DIR=deploy/olm-catalog/compliance-operator
 COURIER_QUAY_NAMESPACE=compliance-operator
 COURIER_PACKAGE_VERSION?=
-OLD_COURIER_PACKAGE_VERSION=$(shell ls -t deploy/olm-catalog/compliance-operator/ | grep -v package.yaml | head -1)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -82,7 +82,7 @@ COURIER_PACKAGE_NAME=compliance-operator-bundle
 COURIER_OPERATOR_DIR=deploy/olm-catalog/compliance-operator
 COURIER_QUAY_NAMESPACE=compliance-operator
 COURIER_PACKAGE_VERSION?=
-OLD_COURIER_PACKAGE_VERSION=$(shell ls -t deploy/olm-catalog/compliance-operator/ | grep -v package.yaml | head -1)
+OLD_COURIER_PACKAGE_VERSION=$(shell ls deploy/olm-catalog/compliance-operator/ | grep -v package.yaml | sort -r | head -1)
```
jhrozek: Even better:
```
find deploy/olm-catalog/compliance-operator/ -mindepth 1 -maxdepth 1  -type d -printf '%f\n' | sort -r
```
that way we get rid of the grep and protect against adding some other file later that might turn up in the ls output and make sure we only list the directories which is what we care about.

mrogers950: +1

JAORMX: Done! Thanks for the suggestion!

PR #178, opened 2020-03-16
Enable JUnit output for unit tests

JAORMX:
This enables us to output JUnit reports which are readable in CI. This
will show up better in the prow UI.

Note that this uses a tool called gotest2junit[1] which lives in the tools
from the openshift/release repo.

[1] https://github.com/openshift/release/tree/master/tools/gotest2junit


JAORMX:
Note that this still requires us to pass the JUNITFILE parameter in CI. So we need to make a commit in the openshift/release repo

PR #188, opened 2020-03-20
Implement remediation auto-apply

JAORMX:
This implements the feature to auto apply remediations by checking if
the scans are all done and then starting to set the "apply" for all of
the remediations belonging to the suite.

Before applying, this actually gets the MachineConfigPool that the
remediation will apply to, and it pauses it. Once all of the "apply"
have been set, it unpauses the pool.

Note that this is only done for Remediations that are of type
MachineConfig.

This also removes the check for nodeSelectors in the suite. This is
instead done on the remediation controller and only for MachineConfig
remediation types

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/openshift/compliance-operator/pkg/apis"
 	"github.com/openshift/compliance-operator/pkg/controller"
+	mcfgapi "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 )
 
@@ -101,8 +102,13 @@ func main() {
 
 	log.Info("Registering Components.")
 
+	mgrscheme := mgr.GetScheme()
 	// Setup Scheme for all resources
-	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
+	if err := apis.AddToScheme(mgrscheme); err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+	if err := mcfgapi.Install(mgrscheme); err != nil {
 		log.Error(err, "")
 		os.Exit(1)
 	}
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -115,6 +115,7 @@ rules:
   - machineconfiguration.openshift.io
   resources:
   - machineconfigs  # The compliance remediation controller operates on MachineConfigs
+  - machineconfigpools
   verbs:
   - list
   - get
--- a/pkg/controller/complianceremediation/complianceremediation_controller.go
+++ b/pkg/controller/complianceremediation/complianceremediation_controller.go
@@ -11,6 +11,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -21,6 +22,7 @@ import (
 
 	compv1alpha1 "github.com/openshift/compliance-operator/pkg/apis/compliance/v1alpha1"
 	"github.com/openshift/compliance-operator/pkg/controller/common"
+	"github.com/openshift/compliance-operator/pkg/utils"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	mcfgClient "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
 )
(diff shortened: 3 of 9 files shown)


jhrozek:
Conceptually looks OK. It would be nice to add a test for the pool matching.

jhrozek:
I'm sorry I didn't notice this during the first round of review, but looking at the patches more carefully I wonder if it wouldn't be more systematic to pause and unpause the pool in the aggregator as opposed to the suite controller, because that's where the remediations are actually being created. The `reconcileRemediations` method in the suite controller doesn't actually touch the remediations themselves, just exposes the remediations to the suite object as an overview.

I have a similar comment about the check for pool existing being done in the remediation controller. Wouldn't it be better to have the check in the scan controller? The code added in the remediation controller looks up a scan object anyway, so why not make it a part of the scan controller in the first place?

JAORMX:
> I'm sorry I didn't notice this during the first round of review, but looking at the patches more carefully I wonder if it wouldn't be more systematic to pause and unpause the pool in the aggregator as opposed to the suite controller, because that's where the remediations are actually being created. The `reconcileRemediations` method in the suite controller doesn't actually touch the remediations themselves, just exposes the remediations to the suite object as an overview.
> 

Doing it in the compliancesuite controller allows us to change the "apply" flag of the remediations at any point.

Example:
An SRE runs the compliancescan with `autoApplyRemediations: false`. Checks the remediations and decides that they make sense to apply. the SRE can then change the ComplianceSuite resource directly, and the controller will make sure that the remediations get applied. This wouldn't be possible if we would handle it from the aggregator itself, as you would need to run a separate scan.

> I have a similar comment about the check for pool existing being done in the remediation controller. Wouldn't it be better to have the check in the scan controller? The code added in the remediation controller looks up a scan object anyway, so why not make it a part of the scan controller in the first place?
Not sure what we would gain from the check being in the scan. Can you elaborate on this?

jhrozek:
> > I'm sorry I didn't notice this during the first round of review, but looking at the patches more carefully I wonder if it wouldn't be more systematic to pause and unpause the pool in the aggregator as opposed to the suite controller, because that's where the remediations are actually being created. The `reconcileRemediations` method in the suite controller doesn't actually touch the remediations themselves, just exposes the remediations to the suite object as an overview.
> 
> Doing it in the compliancesuite controller allows us to change the "apply" flag of the remediations at any point.
> 
> Example:
> An SRE runs the compliancescan with `autoApplyRemediations: false`. Checks the remediations and decides that they make sense to apply. the SRE can then change the ComplianceSuite resource directly, and the controller will make sure that the remediations get applied. This wouldn't be possible if we would handle it from the aggregator itself, as you would need to run a separate scan.

OK, that's a good point.

> 
> > I have a similar comment about the check for pool existing being done in the remediation controller. Wouldn't it be better to have the check in the scan controller? The code added in the remediation controller looks up a scan object anyway, so why not make it a part of the scan controller in the first place?
> Not sure what we would gain from the check being in the scan. Can you elaborate on this?

Code locality and error reporting. Code locality is mostly that we are checking a property of the scan, so it might look nicer (and be more discoverable later) if it's done in the component that deals with the scan. For error reporting, we might, on error, just set the phase to done and set an error message.

JAORMX:
> > > I'm sorry I didn't notice this during the first round of review, but looking at the patches more carefully I wonder if it wouldn't be more systematic to pause and unpause the pool in the aggregator as opposed to the suite controller, because that's where the remediations are actually being created. The `reconcileRemediations` method in the suite controller doesn't actually touch the remediations themselves, just exposes the remediations to the suite object as an overview.
> > 
> > 
> > Doing it in the compliancesuite controller allows us to change the "apply" flag of the remediations at any point.
> > Example:
> > An SRE runs the compliancescan with `autoApplyRemediations: false`. Checks the remediations and decides that they make sense to apply. the SRE can then change the ComplianceSuite resource directly, and the controller will make sure that the remediations get applied. This wouldn't be possible if we would handle it from the aggregator itself, as you would need to run a separate scan.
> 
> OK, that's a good point.
> 
> > > I have a similar comment about the check for pool existing being done in the remediation controller. Wouldn't it be better to have the check in the scan controller? The code added in the remediation controller looks up a scan object anyway, so why not make it a part of the scan controller in the first place?
> > > Not sure what we would gain from the check being in the scan. Can you elaborate on this?
> 
> Code locality and error reporting. Code locality is mostly that we are checking a property of the scan, so it might look nicer (and be more discoverable later) if it's done in the component that deals with the scan. For error reporting, we might, on error, just set the phase to done and set an error message.

would it be an error though? The scan can be fine but the operator just can't remediate.

jhrozek:
On Mon, Mar 23, 2020 at 07:29:27AM -0700, Juan Osorio Robles wrote:
> > > > I'm sorry I didn't notice this during the first round of review, but looking at the patches more carefully I wonder if it wouldn't be more systematic to pause and unpause the pool in the aggregator as opposed to the suite controller, because that's where the remediations are actually being created. The `reconcileRemediations` method in the suite controller doesn't actually touch the remediations themselves, just exposes the remediations to the suite object as an overview.
> > > 
> > > 
> > > Doing it in the compliancesuite controller allows us to change the "apply" flag of the remediations at any point.
> > > Example:
> > > An SRE runs the compliancescan with `autoApplyRemediations: false`. Checks the remediations and decides that they make sense to apply. the SRE can then change the ComplianceSuite resource directly, and the controller will make sure that the remediations get applied. This wouldn't be possible if we would handle it from the aggregator itself, as you would need to run a separate scan.
> > 
> > OK, that's a good point.
> > 
> > > > I have a similar comment about the check for pool existing being done in the remediation controller. Wouldn't it be better to have the check in the scan controller? The code added in the remediation controller looks up a scan object anyway, so why not make it a part of the scan controller in the first place?
> > > > Not sure what we would gain from the check being in the scan. Can you elaborate on this?
> > 
> > Code locality and error reporting. Code locality is mostly that we are checking a property of the scan, so it might look nicer (and be more discoverable later) if it's done in the component that deals with the scan. For error reporting, we might, on error, just set the phase to done and set an error message.
> 
> would it be an error though? The scan can be fine but the operator just can't remediate.

OK, maybe not an error, but a warning. But wouldn't it still be on the
suite level?

jhrozek:
Sorry, a scan, not a suite. So yeah, I guess it boils down to where and how do you want the warning to surface.

JAORMX:
> Sorry, a scan, not a suite. So yeah, I guess it boils down to where and how do you want the warning to surface.

We've decided to keep this as-is for now and introduce warnings to the Scan in a subsequent iteration.

PR #191, opened 2020-03-23
Remove duplicate definition of diffRemediationList

jhrozek:
There was a duplicate definition of diffRemediationList in the codebase.
One that was present since before we introduced the aggregator and since
introducing the aggregator was only called from suite unit tests, but
otherwise unused.

There was also a copy in the aggregatpr, that was not tested. This
commit moves the diff functions to the utils/ package and moves the
tests of these functions from the controller to utils so the functions
are now outside a main package and being tested.


JAORMX:
yay, cleanups!

/lgtm

PR #200, opened 2020-03-31
Display more metadata about complianceremediations

jhrozek:
It was currently quite hard for an administrator to know what a
remediation is doing and why. This patch extends the
complianceremediation CR with a unique ID, a title and a description.

The ID is typically the identifier of the openscap rule. The title is a
one-sentence summary of the remediation and the rationale is a longer,
free-form text.

With this enhancement, the admin can inspect the complianceremediation
object to see what would applying the remediation do to their system and
why.

### DIFF
--- a/deploy/crds/compliance.openshift.io_complianceremediations_crd.yaml
+++ b/deploy/crds/compliance.openshift.io_complianceremediations_crd.yaml
@@ -40,6 +40,9 @@ spec:
               description: Whether the remediation should be picked up and applied
                 by the operator
               type: boolean
+            id:
+              description: A unique identifier of the remediation
+              type: string
             machineConfigContents:
               description: The actual remediation payload
               properties:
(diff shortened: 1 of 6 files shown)


deploy/crds/compliance.openshift.io_complianceremediations_crd.yaml
```diff
@@ -40,6 +40,9 @@ spec:
               description: Whether the remediation should be picked up and applied
                 by the operator
               type: boolean
+            id:
+              description: A unique identifier of the remediation
```
JAORMX: how is the id useful to administrators in general? Seems to me that this further ties the remediation object to the OpenSCAP schema

jhrozek: In case you need to go to the arf output and learn more about the remediation, you can use the rule name.
If we ever support anything else than OpenScap profiles, we can put other data there. The ID is not a rule name, it just happens to contain the rule name if the remediation is generated by openscap. btw both the ID and the other fields are marked as `omitempty`.

jhrozek:
/hold
we decided to refactor how the remediations and their metadata is presented

jhrozek:
Superseded by #211

PR #201, opened 2020-03-31
Fixes to complianceSuite remediations handling

JAORMX:
* Remove watch for remediations: The suite doesn't create them, so it can't really track them. That code is just confusing folks.
* Don't un-pause the pool if the remediation isn't in an "applied" state: Let's wait until it is before un-pausing.
* This also fixes the e2e test to check that the remediation was applied using a pre-fetched machineconfigpool, instead of fetching it in-place. The issue was that we had a race-condition, since the controller had already started applying the remediation when we checked.

### DIFF
--- a/pkg/controller/compliancesuite/compliancesuite_controller.go
+++ b/pkg/controller/compliancesuite/compliancesuite_controller.go
@@ -60,15 +60,6 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	// Watch for changes to secondary resource ComplianceRemediation and requeue the owner ComplianceSuite
-	err = c.Watch(&source.Kind{Type: &compv1alpha1.ComplianceRemediation{}}, &handler.EnqueueRequestForOwner{
-		IsController: true,
-		OwnerType:    &compv1alpha1.ComplianceSuite{},
-	})
-	if err != nil {
-		return err
-	}
-
 	return nil
 }
 
--- a/pkg/controller/compliancesuite/compliancesuite_controller.go
+++ b/pkg/controller/compliancesuite/compliancesuite_controller.go
@@ -102,11 +102,12 @@ func (r *ReconcileComplianceSuite) Reconcile(request reconcile.Request) (reconci
 		return common.ReturnWithRetriableError(reqLogger, err)
 	}
 
-	if err := r.reconcileRemediations(suiteCopy, reqLogger); err != nil {
+	var res reconcile.Result
+	if res, err = r.reconcileRemediations(suiteCopy, reqLogger); err != nil {
 		return common.ReturnWithRetriableError(reqLogger, err)
 	}
 
-	return reconcile.Result{}, nil
+	return res, nil
 }
 
 func (r *ReconcileComplianceSuite) reconcileScans(suite *compv1alpha1.ComplianceSuite, logger logr.Logger) error {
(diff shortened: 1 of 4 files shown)


pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -273,6 +273,20 @@ func (r *ReconcileComplianceSuite) reconcileRemediations(suite *compv1alpha1.Com
...
+	// refresh remediationList
+	if err := r.client.List(context.TODO(), &remList, &listOpts); err != nil {
+		return err
+	}
+
+	// Check that all remediations have been applied yet. If not, cause an error
+	// and requeue.
+	for _, rem := range remList.Items {
+		if rem.Status.ApplicationState != compv1alpha1.RemediationApplied {
+			return fmt.Errorf("Remediation '%s' not applied yet", rem.Name)
```
jhrozek: Wait, isn't this always going to match unless the remediations are auto-applied?

JAORMX: you're right

JAORMX: Done

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -60,15 +60,6 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	// Watch for changes to secondary resource ComplianceRemediation and requeue the owner ComplianceSuite
-	err = c.Watch(&source.Kind{Type: &compv1alpha1.ComplianceRemediation{}}, &handler.EnqueueRequestForOwner{
-		IsController: true,
-		OwnerType:    &compv1alpha1.ComplianceSuite{},
```
jhrozek: Wow, you're right. With this and our earlier discussion about removing the status completely, this is probably just dead code. But let me explain why I added it in the first place: to propagate the changes from the remediation to the suite status. So we can either go the way of removing the status or attempt to fix the watch. I guess fixing the watch might only make sense if there is a release between now and the time we remove the status and at the same time if we care about the status at all.

JAORMX: I don't think we care much about the status. Especially not if we're trying to trim the status to be smaller. What that reconcile function will end up doing is aggregating the remediations and outputting a number... and the auto-apply.

jhrozek: OK, cool.

JAORMX:
/hold

not working yet

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -273,54 +296,70 @@ func (r *ReconcileComplianceSuite) reconcileRemediations(suite *compv1alpha1.Com
...
+	if err := r.client.List(context.TODO(), &remList, &listOpts); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Check that all remediations have been applied yet. If not, cause an error
+	// and requeue.
+	for _, rem := range remList.Items {
+		if rem.Status.ApplicationState != compv1alpha1.RemediationApplied {
+			logger.Info("Remediation not applied yet. Skipping post-processing", "ComplianceRemediation.Name", rem.Name)
+			return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
JAORMX: That was the original intention, and then I removed it. It's reintroduced here: https://github.com/openshift/compliance-operator/pull/204

JAORMX:
dropping in favor of https://github.com/openshift/compliance-operator/pull/204

PR #204, opened 2020-04-02
Remove remediations from suite status

JAORMX:
This is just causing a lot of unnecessary hassle instead of helping...

There are many remediations to keep track of and could potentially
damage the overall performance, since tracking big objects is not ideal
in k8s.
    
So lets just make use of the labels to get the appropriate remediations.


JAORMX:
> By the way, I like the last patch in this series also. I'll leave it up to you whether you want to merge 201 first and then rebase or deprecate 201 in favour of this one.

Lets keep them separate

PR #212, opened 2020-04-15
Enable OCP API scanning

mrogers950:
This enables the "Platform" scan type. The added api-resource-collector component runs as an init container for the platform scan pod. It fetches API objects using a generic client (The cluster's API URL is found by the in-cluster kubeconfig) and saves them to temporary files to be read by the scanner in an offline mode. The API object paths for a scan are sourced from the provided content image and selected by profile.

### DIFF
--- a/cmd/api-resource-collector/log.go
+++ b/cmd/api-resource-collector/log.go
@@ -1,0 +1,38 @@
+/*
+Copyright © 2020 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+var debugLog bool
+
+func LOG(format string, a ...interface{}) {
+	fmt.Printf(format+"\n", a...)
+}
+
+func DBG(format string, a ...interface{}) {
+	if debugLog {
+		LOG("debug: "+format, a...)
+	}
+}
+
+func FATAL(format string, a ...interface{}) {
+	fmt.Fprintf(os.Stderr, "FATAL:"+format+"\n", a...)
+	os.Exit(1)
+}
--- a/cmd/api-resource-collector/scap.go
+++ b/cmd/api-resource-collector/scap.go
@@ -1,0 +1,290 @@
+/*
+Copyright © 2020 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/subchen/go-xmldom"
+
+	"github.com/ghodss/yaml"
+
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openshift/compliance-operator/pkg/utils"
+)
+
+const (
+	contentFileTimeout = 3600
+)
+
+// For OpenSCAP content as an XML data stream. Implements ResourceFetcher.
+type scapContentDataStream struct {
+	// Client for Gets
+	client *kubernetes.Clientset
+	// Staging objects
+	dataStream *utils.XMLDocument
+	resources  map[string]string
+	found      map[string][]byte
+}
+
+func (c *scapContentDataStream) LoadSource(path string) error {
+	f, err := openNonEmptyFile(path)
+	if err != nil {
+		return err
+	}
+	// #nosec
+	defer f.Close()
+	xml, err := parseContent(f)
+	if err != nil {
+		return err
+	}
+	c.dataStream = xml
+	return nil
+}
+
+func parseContent(f *os.File) (*utils.XMLDocument, error) {
+	return utils.ParseContent(bufio.NewReader(f))
+}
+
+// Returns the file, but only after it has been created by the other init container.
+// This avoids a race.
+func openNonEmptyFile(filename string) (*os.File, error) {
+	readFileTimeoutChan := make(chan *os.File, 1)
+
+	// gosec complains that the file is passed through an evironment variable. But
+	// this is not a security issue because none of the files are user-provided
+	cleanFileName := filepath.Clean(filename)
+
+	go func() {
+		for {
+			// Note that we're cleaning the filename path above.
+			// #nosec
+			file, err := os.Open(cleanFileName)
+			if err == nil {
+				fileinfo, err := file.Stat()
+				// Only try to use the file if it already has contents.
+				if err == nil && fileinfo.Size() > 0 {
+					readFileTimeoutChan <- file
+				}
+			} else if !os.IsNotExist(err) {
+				fmt.Println(err)
+				os.Exit(1)
+			}
+			time.Sleep(1 * time.Second)
+		}
+	}()
+
+	select {
+	case file := <-readFileTimeoutChan:
+		fmt.Printf("File '%s' found, using.\n", filename)
+		return file, nil
+	case <-time.After(time.Duration(contentFileTimeout) * time.Second):
+		fmt.Println("Timeout. Aborting.")
+		os.Exit(1)
+	}
+
+	// We shouldn't get here.
+	return nil, nil
+}
+
+func (c *scapContentDataStream) FigureResources(profile string) error {
+	paths := getResourcePaths(c.dataStream, profile)
+	if len(paths) == 0 {
+		return fmt.Errorf("no checks found in datastream")
+	}
+	c.resources = paths
+	return nil
+}
+
+const (
+	endPointTag    = "ocp-api-endpoint"
+	endPointTagEnd = endPointTag + "\">"
+	dumpTag        = "ocp-dump-location"
+	dumpTagEnd     = dumpTag + "\"><sub idref=\"xccdf_org.ssgproject.content_value_ocp_data_root\" use=\"legacy\" />"
+	codeTag        = "</code>"
+)
+
+// getPathsFromRuleWarning finds the API endpoint and save path from in. The expected structure is:
+//
+//  <warning category="general" lang="en-US"><code class="ocp-api-endpoint">/apis/config.openshift.io/v1/oauths/cluster
+//  </code><code class="ocp-dump-location"><sub idref="xccdf_org.ssgproject.content_value_ocp_data_root" use="legacy" />
+//  /idp.yml</code>file.</warning>
+//
+// We return both values or none at all.
+func getPathsFromWarningXML(in string) (string, string) {
+	var (
+		apiPathOut  string
+		dumpPathOut string
+	)
+
+	DBG("%s", in)
+	apiIndex := strings.Index(in, endPointTag)
+	if apiIndex == -1 {
+		return "", ""
+	}
+
+	apiValueBeginIndex := apiIndex + len(endPointTagEnd)
+	apiValueEndIndex := strings.Index(in[apiValueBeginIndex:], codeTag)
+	if apiValueEndIndex == -1 {
+		return "", ""
+	}
+
+	apiPathOut = in[apiValueBeginIndex : apiValueBeginIndex+apiValueEndIndex]
+
+	dumpIndex := strings.Index(in, dumpTag)
+	if dumpIndex == -1 {
+		return "", ""
+	}
+
+	dumpValueBeginIndex := dumpIndex + len(dumpTagEnd)
+	dumpValueEndIndex := strings.Index(in[dumpValueBeginIndex:], codeTag)
+	if dumpValueEndIndex == -1 {
+		return "", ""
+	}
+
+	dumpPathOut = in[dumpValueBeginIndex : dumpValueBeginIndex+dumpValueEndIndex]
+
+	return apiPathOut, dumpPathOut
+}
+
+// Collect the resource paths for objects that this scan needs to obtain, and their corresponding file paths based on
+// profile. The profile will have a series of "selected" checks that we grab all of the path info from.
+func getResourcePaths(ds *utils.XMLDocument, profile string) map[string]string {
+	out := map[string]string{}
+	selectedChecks := []string{}
+
+	// First we find the Profile node, to locate the enabled checks.
+	DBG("Using profile %s", profile)
+	nodes := ds.Root.Query("//Profile")
+	for _, node := range nodes {
+		profileID := node.GetAttributeValue("id")
+		if profileID != profile {
+			continue
+		}
+
+		checks := node.GetChildren("select")
+		for _, check := range checks {
+			if check.GetAttributeValue("selected") != "true" {
+				continue
+			}
+
+			if idRef := check.GetAttributeValue("idref"); idRef != "" {
+				DBG("selected: %v", idRef)
+				selectedChecks = append(selectedChecks, idRef)
+			}
+		}
+	}
+
+	checkDefinitions := ds.Root.Query("//Rule")
+	if len(checkDefinitions) == 0 {
+		DBG("WARNING: No rules to query (invalid datastream)")
+		return out
+	}
+
+	// For each of our selected checks, collect the required path info.
+	for _, checkID := range selectedChecks {
+		var found *xmldom.Node
+		for _, rule := range checkDefinitions {
+			if rule.GetAttributeValue("id") == checkID {
+				found = rule
+				break
+			}
+		}
+		if found == nil {
+			DBG("WARNING: Couldn't find a check for id %s", checkID)
+			continue
+		}
+
+		// This node is called "warning" and contains the path info. It's not an actual "warning" for us here.
+		warning := found.GetChild("warning")
+		if warning == nil {
+			DBG("Couldn't find 'warning' child of check %s", checkID)
+			continue
+		}
+
+		apiPath, dumpPath := getPathsFromWarningXML(warning.XML())
+		if len(apiPath) == 0 || len(dumpPath) == 0 {
+			continue
+		}
+		out[apiPath] = dumpPath
+	}
+
+	return out
+}
+
+func (c *scapContentDataStream) FetchResources() error {
+	found, err := fetch(c.client, c.resources)
+	if err != nil {
+		return err
+	}
+	c.found = found
+	return nil
+}
+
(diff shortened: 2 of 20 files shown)


cmd/api-resource-collector/testdata/ssg-ocp4-ds-mocked.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0"?>
```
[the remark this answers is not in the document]
mrogers950: @JAORMX good catch, will do.

deploy/crds/compliance.openshift.io_v1alpha1_compliancesuite_cr.yaml
```diff
@@ -7,12 +7,14 @@ spec:
...
       - name: workers-scan
         profile: xccdf_org.ssgproject.content_profile_moderate
+        scanType: Node
         content: ssg-ocp4-ds.xml
         contentImage: quay.io/jhrozek/ocp4-openscap-content:latest
         nodeSelector:
             node-role.kubernetes.io/worker: ""
       - name: masters-scan
         profile: xccdf_org.ssgproject.content_profile_moderate
+        scanType: Node
```
mrogers950, line 17: @JAORMX @jhrozek I thought about making Node the default if you don't specify.

JAORMX: That sounds good to me.

cmd/api-resource-collector/scap.go
```diff
@@ -0,0 +1,235 @@
...
+// 	- use saved ID to find "reference" with ref_id==id
+func getResourcePaths(ds *utils.XMLDocument) map[string]string {
+	out := map[string]string{}
+
+	LOG("Searching for API paths and filenames from datastream..")
+	nodes := ds.Root.Query("//yamlfilecontent_test")
+	for _, node := range nodes {
+		id := node.GetAttributeValue("id")
+		DBG("yamlfilecontent_test id: '%s'", id)
+		if !strings.HasPrefix(id, "oval:ssg-test_") || !strings.HasSuffix(id, ":tst:1") {
```
JAORMX, line 145: what is `:tst:1`? Might wanna comment the code for that.

cmd/api-resource-collector/scap.go
```diff
@@ -0,0 +1,235 @@
...
+//		- if none, error.
+//	- Find each "selected" child.
+//  - If selected=="true" and idref==prefix(xccdf_org.ssgproject.content_rule_)
+//     - strip off xccdf_org.ssgproject.content_rule_ prefix and save that id.
+// 	- use saved ID to find "reference" with ref_id==id
+func getResourcePaths(ds *utils.XMLDocument) map[string]string {
+	out := map[string]string{}
+
+	LOG("Searching for API paths and filenames from datastream..")
+	nodes := ds.Root.Query("//yamlfilecontent_test")
```
JAORMX, line 141: So... as far as I can tell, this goes through every yamlfilecontent_test definition in the content and gather alls the id's with the associated files. Is this so? Should we already introduce filtering by profile? this way we reduce the number of calls to the kube API server.

cmd/api-resource-collector/scap.go
```diff
@@ -0,0 +1,235 @@
...
+			defer stream.Close()
+			body, err := ioutil.ReadAll(stream)
+			if err != nil {
+				return err
+			}
+			if len(body) == 0 {
+				DBG("no data in request body")
+				return nil
+			}
+			yaml, err := yaml.JSONToYAML(body)
```
JAORMX, line 209: Instead of this, you could use the `k8s.io/apimachinery/pkg/util/yaml` package, which contains similar functions. Eventually, you can get that into an `unstructured.Unstructured` object which allows us to have a generic way to access the objects.

I used that for the CaC e2e tests if you want a reference: https://github.com/ComplianceAsCode/content/blob/master/tests/ocp4e2e/helpers.go#L392

This way, we also don't introduce another dependency.

deploy/role_binding.yaml
```diff
@@ -46,3 +46,16 @@ roleRef:
   kind: Role
   name: remediation-aggregator
   apiGroup: rbac.authorization.k8s.io
+---
+kind: ClusterRoleBinding
```
JAORMX, line 50: Does this get used as part of the csv generation? I think you might need to add these capabilities directly into an api-resource-collector role.

pkg/apis/compliance/v1alpha1/compliancescan_types.go
```diff
@@ -48,6 +48,8 @@ const (
 	ResultError ComplianceScanStatusResult = "ERROR"
 	// ResultNonCompliant represents the compliance scan having found a gap
 	ResultNonCompliant ComplianceScanStatusResult = "NON-COMPLIANT"
+	ScanTypeNode       ComplianceScanType         = "Node"
+	ScanTypePlatform   ComplianceScanType         = "Platform"
```
JAORMX, line 52: nice naming. I dig it.

pkg/controller/compliancescan/scan.go
```diff
@@ -213,14 +248,210 @@ func newScanPodForNode(scanInstance *compv1alpha1.ComplianceScan, node *corev1.N
...
+					Command: []string{OpenScapScriptPath},
+					SecurityContext: &corev1.SecurityContext{
+						Privileged: &trueVal,
+					},
+					VolumeMounts: []corev1.VolumeMount{
+						{
+							Name:      "host",
+							MountPath: "/host",
+							ReadOnly:  true,
+						},
```
JAORMX, line 354: will this host mount be needed if the scan type is of platform type?

pkg/controller/compliancescan/scan.go
```diff
@@ -213,14 +248,210 @@ func newScanPodForNode(scanInstance *compv1alpha1.ComplianceScan, node *corev1.N
...
+			Tolerations: []corev1.Toleration{
+				{
+					Key:      "node-role.kubernetes.io/master",
+					Operator: "Exists",
+					Effect:   "NoSchedule",
+				},
+			},
+			RestartPolicy: corev1.RestartPolicyOnFailure,
+			Volumes: []corev1.Volume{
+				// TODO: hostmount not needed for scans
```
JAORMX, line 396: ah! I see

pkg/controller/compliancescan/scan.go
```diff
@@ -213,14 +248,210 @@ func newScanPodForNode(scanInstance *compv1alpha1.ComplianceScan, node *corev1.N
...
 func (r *ReconcileComplianceScan) deleteScanPods(instance *compv1alpha1.ComplianceScan, nodes corev1.NodeList, logger logr.Logger) error {
 	// On each eligible node..
 	for _, node := range nodes.Items {
-		// ..schedule a pod..
-		logger.Info("Creating a pod for node", "Node.Name", node.Name)
+		logger.Info("Deleting a pod on node", "node", node.Name)
 		pod := newScanPodForNode(instance, &node, logger)
 
-		// ..and launch it..
+		// Delete it.
```
JAORMX, line 454: lol, thanks for the cleanup

cmd/api-resource-collector/scap.go
mrogers950: @JAORMX Will do, thanks!

cmd/api-resource-collector/scap.go
mrogers950: That's what I'll be doing when I rewrite this functon (the current one is just for my experimental datastream)

cmd/api-resource-collector/scap.go
mrogers950: This will change

cmd/api-resource-collector/log.go
```diff
@@ -0,0 +1,38 @@
...
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+var debugLog bool
+
+func LOG(format string, a ...interface{}) {
```
jhrozek, line 25: I know we talked about debug logging for a long time and this is as good of a start as any. Do you think it would make sense to move this file to a shared module used by all the helper containers? If not, we should at least note this file in a ticket (I think we already have one?)

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -413,16 +457,18 @@ func gatherResults(r *ReconcileComplianceScan, instance *compv1alpha1.Compliance
 	var result compv1alpha1.ComplianceScanStatusResult
 	compliant := true
 	isReady := true
-	for _, node := range nodes.Items {
+
+	switch instance.Spec.ScanType {
+	case compv1alpha1.ScanTypePlatform:
 		targetCM := types.NamespacedName{
-			Name:      getConfigMapForNodeName(instance.Name, node.Name),
+			Name:      getConfigMapForNodeName(instance.Name, "api-checks"),
```
jhrozek, line 464: I would prefer to avoid using hardcoded constants and define the name in `config.go` (or add `constants.go` if you think it is cleaner)

pkg/controller/compliancescan/scan.go
```diff
@@ -213,14 +248,210 @@ func newScanPodForNode(scanInstance *compv1alpha1.ComplianceScan, node *corev1.N
...
+							ReadOnly:  true,
+						},
+					},
+				},
+				{
+					Name:    OpenSCAPScanContainerName,
+					Image:   GetComponentImage(OPENSCAP),
+					Command: []string{OpenScapScriptPath},
+					SecurityContext: &corev1.SecurityContext{
+						Privileged: &trueVal,
```
jhrozek, line 347: Does the pod still need to run privileged if it doesn't mount the host fs?

pkg/controller/compliancescan/scan.go
```diff
@@ -47,6 +56,32 @@ func (r *ReconcileComplianceScan) createScanPods(instance *compv1alpha1.Complian
...
+		logger.Info("Pod already exists. This is fine.", "pod", pod)
+	} else if err != nil {
+		log.Error(err, "Failed to launch a pod", "pod", pod)
+		return err
+	} else {
+		logger.Info("Launched a pod", "pod", pod)
+	}
+
+	// make sure the instance is updated with the node-pod labels
+	// TODO: Is this needed for node-pod labels or ownerref?
```
jhrozek, line 78: Why are node-pod-labels for a platform scan needed? About the ownerref, I wouldn't update the pod for it, either it doesn't exist in which case the create does all the work or the pod exists and imo the function shouldn't then touch it.

pkg/controller/compliancescan/scan.go
```diff
@@ -213,14 +248,210 @@ func newScanPodForNode(scanInstance *compv1alpha1.ComplianceScan, node *corev1.N
...
+		"complianceScan": scanInstance.Name,
+	}
+
+	return &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: scanInstance.Namespace,
+			Labels:    podLabels,
+			Annotations: map[string]string{
+				"openshift.io/scc": "node-exporter",
```
jhrozek, line 267: Why is the `scc/node-exporter` needed?

cmd/api-resource-collector/log.go
mrogers950: let's do a follow up for this as there's probably something a little cleaner we can do. This one I used here mostly to save typing.

pkg/controller/compliancescan/compliancescan_controller.go
mrogers950: will do

pkg/controller/compliancescan/scan.go
mrogers950: removed this part

pkg/controller/compliancescan/scan.go
mrogers950: It should not. I've removed the hostroot and scc bits from this part.

cmd/api-resource-collector/scap.go
mrogers950: @JAORMX I've updated the parsing to match what we actually expect from CaC now. It starts by first getting the enabled rules from the selected profile.

deploy/role_binding.yaml
mrogers950: I intend to break this out into another role (for now was sharing with regular scap scan)

mrogers950:
@JAORMX @jhrozek I've pushed an update with the revised content parsing and removed hostroot/scc privs from the platform scan container. Still left todo from the last review:
- Make separate collector role
- Fix XML dependency

JAORMX:
> @JAORMX @jhrozek I've pushed an update with the revised content parsing and removed hostroot/scc privs from the platform scan container. Still left todo from the last review:
> 
>     * Make separate collector role
> 
>     * Fix XML dependency

That sounds good to me and could be addressed in a subsequent PR.

cmd/api-resource-collector/scap.go
```diff
@@ -0,0 +1,290 @@
...
+		}
+	}
+	return results, nil
+}
+
+func (c *scapContentDataStream) SaveResources(to string) error {
+	return saveResources(c.client, to, c.found)
+}
+
+func saveResources(client *kubernetes.Clientset, dir string, data map[string][]byte) error {
```
JAORMX, line 282: what's the client used for here?

JAORMX:
This passed in my local test, I think you just need to increase the timeout.

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -372,12 +403,17 @@ func (r *ReconcileComplianceScan) phaseDoneHandler(instance *compv1alpha1.Compli
 func getTargetNodes(r *ReconcileComplianceScan, instance *compv1alpha1.ComplianceScan) (corev1.NodeList, error) {
 	var nodes corev1.NodeList
 
-	listOpts := client.ListOptions{
-		LabelSelector: labels.SelectorFromSet(instance.Spec.NodeSelector),
-	}
+	switch instance.Spec.ScanType {
```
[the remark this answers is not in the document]
mrogers950: Yes, there are uses of getTargetNodes in the phase reconciling functions (phaseAggregatingHandler, etc)

tests/e2e/e2e_test.go
```diff
@@ -725,5 +725,92 @@ func TestE2E(t *testing.T) {
...
+				}
+
+				// At this point, both scans should be non-compliant given our current content
+				err = scanResultIsExpected(f, namespace, workerScanName, compv1alpha1.ResultNonCompliant)
+				if err != nil {
+					return err
+				}
+
+				// The profile should find one check for an htpasswd IDP, so we should be compliant.
+				err = scanResultIsExpected(f, namespace, platformScanName, compv1alpha1.ResultCompliant)
```
mrogers950, line 788: @JAORMX I had the test previosly set to assert non-compliance, but an htpasswd IDP should not be on the test cluster by default (that's the check that is selected with the `xccdf_org.ssgproject.content_profile_platform-moderate` profile), so being in compliance is the result we should expect. So there is still something funky with the scanner's check. I changed this to assert compliance and increased the timeout.

cmd/api-resource-collector/scap.go
```diff
@@ -0,0 +1,290 @@
...
+)
+
+// getPathsFromRuleWarning finds the API endpoint and save path from in. The expected structure is:
+//
+//  <warning category="general" lang="en-US"><code class="ocp-api-endpoint">/apis/config.openshift.io/v1/oauths/cluster
+//  </code><code class="ocp-dump-location"><sub idref="xccdf_org.ssgproject.content_value_ocp_data_root" use="legacy" />
+//  /idp.yml</code>file.</warning>
+//
+// We return both values or none at all.
+func getPathsFromWarningXML(in string) (string, string) {
```
JAORMX, line 135: Now that I've had a better chance at reading the XCCDF standard, I commented on the patch that adds the YAML checks to the content repo. Not very keen on using a warning here, let's see what the OpenSCAP team says.

tests/e2e/e2e_test.go
JAORMX: I take it this has been addressed now?

mrogers950:
I've revised the api-collector to use the API path as the save-to path. This should fail the tests until I push a new quay.io/mrogers950/ocp4-openscap-content:platform that contains the changes to https://github.com/ComplianceAsCode/content/pull/5326 (from comments https://github.com/ComplianceAsCode/content/pull/5326#discussion_r419898922 )

mrogers950:
I rebased and updated the testing content image. Manual testing looked good.

images/api-resource-collector/Dockerfile.ci
```diff
@@ -0,0 +1,18 @@
+# Step one: build fetcher
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
```
jhrozek, line 2: You can already use `FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder`, other components use 1.13 in the meantime. Doesn't need to be as part of this patch, but it would be nice to fix soon, I've already seen issues with some deps that use the new errors API..

pkg/controller/compliancescan/scan.go
```diff
@@ -219,14 +249,192 @@ func newScanPodForNode(scanInstance *compv1alpha1.ComplianceScan, node *corev1.N
...
+	}
+
+	return &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: scanInstance.Namespace,
+			Labels:    podLabels,
+		},
+		Spec: corev1.PodSpec{
+			ServiceAccountName: resultscollectorSA,
```
jhrozek, line 276: Wouldn't it be better to create a separate role just for the API collector rather than reuse two roles for the same thing?

tests/e2e/e2e_test.go
```diff
@@ -825,5 +825,92 @@ func TestE2E(t *testing.T) {
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      suiteName,
+						Namespace: namespace,
+					},
+					Spec: compv1alpha1.ComplianceSuiteSpec{
+						AutoApplyRemediations: false,
+						Scans: []compv1alpha1.ComplianceScanSpecWrapper{
+							{
+								ComplianceScanSpec: compv1alpha1.ComplianceScanSpec{
+									ContentImage: "quay.io/jhrozek/ocp4-openscap-content:ignition_remediation",
```
jhrozek, line 850: You can already use ` ContentImage: "quay.io/complianceascode/ocp4:latest"` to minimise usage of personal repositories in the codebase.

cmd/api-resource-collector/scap.go
```diff
@@ -279,12 +254,34 @@ func (c *scapContentDataStream) SaveResources(to string) error {
...
-func saveResources(dir string, data map[string][]byte) error {
-	for f, d := range data {
-		err := ioutil.WriteFile(path.Join(dir, f), d, 0600)
+func saveResources(rootDir string, data map[string][]byte) error {
+	for apiPath, fileContents := range data {
+		saveDir, saveFile, err := getSaveDirectoryAndFileName(rootDir, apiPath)
+		if err != nil {
+			return err
+		}
+		err = os.MkdirAll(saveDir, os.ModePerm)
```
jhrozek, line 263: why create the dirs with 777?

images/api-resource-collector/Dockerfile.ci
mrogers950: Actually I'm removing this image when I update CO to use sub-commands

pkg/controller/compliancescan/scan.go
mrogers950: Yep, that's a follow-up

tests/e2e/e2e_test.go
mrogers950: Follow-up

cmd/api-resource-collector/scap.go
mrogers950: Whoops, I left in that was my own laziness. Will follow-up

PR #221, opened 2020-04-20
Generic remediations round #1

JAORMX:
This does a first pass at implementing generic kubernetes remediations:

* Introduces an "Object" entry into the ComplianceRemediation object

* If "MachineConfigContents" is set into the ComplianceRemediation, the
  contents will get placed into the Object entry instead.

* MachineConfigs are now parsed as generic objects, special handling is
  only done in the upper layers of the operator (in the controllers)

* This also phases out the usage of ComplianceRemediation.Spec.Type. As
   we can now rely on the type of the unstructured object.

This is it, other objects are not addressed yet.


pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -121,6 +124,21 @@ func (r *ReconcileComplianceRemediation) Reconcile(request reconcile.Request) (r
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileComplianceRemediation) updateDeprecatedMcRemediation(instance *compv1alpha1.ComplianceRemediation, logger logr.Logger) error {
+	remCopy := instance.DeepCopy()
+	remCopy.Spec.Object = remCopy.Spec.MachineConfigContents.DeepCopy()
+	remCopy.Spec.MachineConfigContents = nil
+	return r.client.Update(context.TODO(), remCopy)
```
jhrozek, line 131: Can you just already continue to reconcileRemediation with the updated object?

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -218,7 +236,13 @@ func getApplicableMcList(r *ReconcileComplianceRemediation, instance *compv1alph
 			// TODO(jaosorior): Add status about remediation not being applicable
 			return appliedRemediations, nil
 		}
-		appliedRemediations = append(appliedRemediations, &instance.Spec.MachineConfigContents)
+
+		mc, err := parseMachineConfig(instance, instance.Spec.Object)
+		if err != nil {
+			logger.Error(err, "Cannot list the pools for the remediation")
```
jhrozek, line 242: Looks like wrong error message?

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -243,9 +267,10 @@ func getAppliedMcRemediations(r *ReconcileComplianceRemediation, rem *compv1alph
 
 	appliedRemediations := make([]*mcfgv1.MachineConfig, 0, len(scanSuiteRemediations.Items))
 	for i := range scanSuiteRemediations.Items {
-		if scanSuiteRemediations.Items[i].Spec.Type != compv1alpha1.McRemediation {
-			// We'll only merge MachineConfigs
-			// TODO: Add a log line with a very high log level
+		if scanSuiteRemediations.Items[i].Spec.Object == nil {
+			continue
+		}
```
jhrozek, line 272: Sounds like this condition could be rolled into the function (nil is not an MC..)

pkg/controller/complianceremediation/complianceremediation_controller.go
JAORMX: I couldn't. Cause it would error out anyway cause the object has been updated. So this avoids that.

pkg/controller/complianceremediation/complianceremediation_controller.go
JAORMX: oops

JAORMX:
/retest

I couldn't reproduce this locally...

pkg/controller/complianceremediation/complianceremediation_controller_test.go
```diff
@@ -37,30 +40,35 @@ func getMockedRemediation(name string, labels map[string]string, applied bool, s
...
+				Ignition: igntypes.Ignition{
+					Version: igntypes.MaxVersion.String(),
+				},
+				Storage: igntypes.Storage{
+					Files: files,
+				},
+			},
+		},
+	}
+	unstructuredobj, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(mc)
```
mrogers950, line 60: Should this test go from a raw definition to unstructured instead (ie, rawObjectToUnstructured())? It might not make much difference but just closer to what would happen in practice.

pkg/utils/parse_arf_result_test.go
```diff
@@ -1,19 +1,32 @@
...
 	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
 
 	compv1alpha1 "github.com/openshift/compliance-operator/pkg/apis/compliance/v1alpha1"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 )
 
+func parseMachineConfig(rem *compv1alpha1.ComplianceRemediation, obj *unstructured.Unstructured) (*mcfgv1.MachineConfig, error) {
```
mrogers950, line 19: dupe of parseMachineConfig() in complianceremediation_controller.go

mrogers950:
Looks good (the CI failure is certainly unrelated), just a few comments inline.

pkg/controller/complianceremediation/complianceremediation_controller_test.go
JAORMX: I actually didn't know how to define the MachineConfig as a raw object, hence why I converted it :smile: tips are welcome.

pkg/controller/complianceremediation/complianceremediation_controller_test.go
```diff
@@ -8,19 +8,23 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	compv1alpha1 "github.com/openshift/compliance-operator/pkg/apis/compliance/v1alpha1"
+	"github.com/openshift/compliance-operator/pkg/utils"
 )
 
 func isRemInList(mcList []*mcfgv1.MachineConfig, rem *compv1alpha1.ComplianceRemediation) bool {
+	remMc, _ := utils.ParseMachineConfig(rem, rem.Spec.Object)
```
jhrozek, line 25: Why is the error ignored here?

pkg/utils/parse_arf_result_test.go
```diff
@@ -152,7 +152,8 @@ var _ = Describe("XCCDF parser", func() {
 				)
 
 				BeforeEach(func() {
-					mcFiles = rem.Spec.MachineConfigContents.Spec.Config.Storage.Files
+					mcfg, _ := ParseMachineConfig(rem, rem.Spec.Object)
```
jhrozek, line 155: Also here the error could be checked

pkg/controller/complianceremediation/complianceremediation_controller_test.go
JAORMX: Cause the remediation in this test is something that's static and that we control. An error here would be an error to parse the MachineConfig, which would be highly unlikely in a unit test with a hardcoded MachineConfig.

pkg/utils/parse_arf_result_test.go
JAORMX: Same as above.

JAORMX:
> two questions left, otherwise looks good

thanks! responded.

jhrozek:
/approve
the patches look fine to me now

mrogers950:
@JAORMX I'm good with merging this.

PR #223, opened 2020-04-21
Enforce TLS1.2 and default ciphers between result components

mrogers950:


cmd/resultscollector/main.go
```diff
@@ -398,12 +400,15 @@ func getMutualHttpsTransport(c *scapresultsConfig) (*http.Transport, error) {
 	pool := x509.NewCertPool()
 	pool.AppendCertsFromPEM(ca)
 
-	// TODO: Configure cipher suites. Perhaps use library-go helper functions.
+	tlsConfig := &tls.Config{}
+	// Configures TLS 1.2
+	tlsConfig = libgocrypto.SecureTLSConfig(tlsConfig)
+	tlsConfig.CipherSuites = []uint16{tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256}
```
JAORMX, line 406: the default cipher suites seem quite OK also, Do you think it would be a good idea to just use those ones? The observation is that this one isn't available in TLS 1.3.

mrogers950:
@JAORMX @jhrozek Looks like we need to manually merge.

PR #232, opened 2020-04-28
Add Tailoring support

JAORMX:
This adds tailoring support for profiles by adding a `tailoringConfigMap` entry to the `ComplianceScan` object. This ConfigMap gets mounted on the scan pod and is used as the `--tailoring-file` flag to the `oscap` command.

The tailoring ConfigMap should:
* Have a key called `tailoring.xml` which is an opinionated/hardcoded file name to be used by the scan script.
* Contain a valid XML tailoring file.

For taking this feature into use, the user needs to set the profile ID to be used should be the one for the tailoring.


pkg/controller/compliancescan/scan.go
```diff
@@ -27,6 +32,12 @@ func (r *ReconcileComplianceScan) createScanPods(instance *compv1alpha1.Complian
 			return err
 		}
 
+		if instance.Spec.TailoringConfigMap != nil {
+			if err := r.addTailoringVolume(instance, pod); err != nil {
+				return err
+			}
+		}
+
```
jhrozek, line 40: Why isn't this hunk part of the function that creates the pod structure in the first place?

pkg/controller/compliancescan/scan.go
```diff
@@ -235,6 +246,58 @@ func (r *ReconcileComplianceScan) deleteScanPods(instance *compv1alpha1.Complian
 	return nil
 }
 
+func (r *ReconcileComplianceScan) addTailoringVolume(instance *compv1alpha1.ComplianceScan, pod *corev1.Pod) error {
+	mode := int32(0744)
```
jhrozek, line 250: Does the tailoring file need to be executable?

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -197,6 +197,17 @@ func (r *ReconcileComplianceScan) phaseLaunchingHandler(instance *compv1alpha1.C
 	}
 
 	if err = r.createScanPods(instance, nodes, logger); err != nil {
+		if !common.IsRetriable(err) {
```
jhrozek, line 200: I like surfacing the error, but wouldn't it be better to do it in the main `Reconcile` loop entry point?

pkg/controller/compliancescan/scan.go
JAORMX: just thought that having it in one spot would make validations and addition of the volume easier.

pkg/controller/compliancescan/scan.go
JAORMX: Good point, let me change that.

tests/e2e/e2e_test.go
```diff
@@ -45,6 +46,61 @@ func TestE2E(t *testing.T) {
...
+					},
+					Data: map[string]string{
+						"tailoring.xml": `<?xml version="1.0" encoding="UTF-8"?>
+<xccdf-1.2:Tailoring xmlns:xccdf-1.2="http://checklists.nist.gov/xccdf/1.2" id="xccdf_compliance.openshift.io_tailoring_test-tailoredprofile">
+	<xccdf-1.2:benchmark href="/content/ssg-ocp4-ds.xml"></xccdf-1.2:benchmark>
+	<xccdf-1.2:version time="2020-04-28T07:04:13Z">1</xccdf-1.2:version>
+	<xccdf-1.2:Profile id="xccdf_compliance.openshift.io_profile_test-tailoredprofile">
+		<xccdf-1.2:title>Test Tailored Profile</xccdf-1.2:title>
+		<xccdf-1.2:description>Test Tailored Profile</xccdf-1.2:description>
+		<xccdf-1.2:select idref="xccdf_org.ssgproject.content_rule_no_netrc_files" selected="true"></xccdf-1.2:select>
```
jhrozek, line 66: Is it possible to extend the test to check that the tailoring got processed?

JAORMX: The tailoring gets used because we specifically use the tailoring's profile ID in the CR. Which is `xccdf_compliance.openshift.io_profile_test-tailoredprofile`. That ID doesn't exist in the profile nor in the whole benchmark, just in the tailoring file.

jhrozek: Great, I missed that.

pkg/controller/compliancescan/scan.go
jhrozek: It's not a big deal.

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: Mind if I do that in a separate PR? That needs further refactoring as we need to handle when the result has a requeue.

jhrozek: sure

PR #234, opened 2020-04-29
Generic Remediations round #2

JAORMX:
This implements the missing bits to handle generic remediations in the
operator. If the remediation is not a MachineConfig, it'll just ensure
that the object exists in the cluster.

If the remediation is of an object of a kind that doesn't exist in the
cluster, the error gets reflected in the Remediation CR.

### DIFF
--- a/pkg/controller/complianceremediation/complianceremediation_controller.go
+++ b/pkg/controller/complianceremediation/complianceremediation_controller.go
@@ -7,6 +7,7 @@ import (
 	"github.com/go-logr/logr"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -108,12 +109,13 @@ func (r *ReconcileComplianceRemediation) Reconcile(request reconcile.Request) (r
 	}
 
 	// this would have been much nicer with go 1.13 using errors.Is()
-	if err != nil {
+	// Only return if the error is retriable. Else, we persist it in the status
+	if err != nil && common.IsRetriable(err) {
 		return common.ReturnWithRetriableError(reqLogger, err)
 	}
 
 	// Second, we'll reconcile the status of the Remediation itself
-	err = r.reconcileRemediationStatus(remediationInstance, reqLogger)
+	err = r.reconcileRemediationStatus(remediationInstance, reqLogger, err)
 	// this would have been much nicer with go 1.13 using errors.Is()
 	if err != nil {
 		return common.ReturnWithRetriableError(reqLogger, err)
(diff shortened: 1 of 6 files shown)


pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -134,7 +134,22 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
...
+// remediation if applicable
+func (r *ReconcileComplianceRemediation) reconcileGenericRemediation(instance *compv1alpha1.ComplianceRemediation, logger logr.Logger) error {
+	if instance.Spec.Apply {
+		obj := instance.Spec.Object
+		found := obj.DeepCopy()
+		err := r.client.Get(context.TODO(), types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, found)
+		if errors.IsNotFound(err) {
+			err = r.client.Create(context.TODO(), obj)
+		}
+		// TODO(jaosorior): If the object is found, should we update it?
```
jhrozek, line 150: Update how? I think until we have a mechanism to apply the generic remediations, I don't know what would we do here. The remediation status (Applied/Unapplied/...) is handled in a separate function.

JAORMX: We already have a mechanism to apply generic remediations. Generic in the sense that it has to be a kube object.

But what I mean is, let's say that you have a configmap you gotta check for, and the contents don't match the remediation. Right now it only checks if the object (in the example it would be the CM) exists in the cluster, if not, it creates it. But, should we do an `update` operation to overwrite the object? that's the question.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -134,7 +136,25 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
...
+	return r.reconcileGenericRemediation(instance, logger)
+}
+
+// Gets a generic remediation and ensures the object exists in the cluster if the
+// remediation if applicable
+func (r *ReconcileComplianceRemediation) reconcileGenericRemediation(instance *compv1alpha1.ComplianceRemediation, logger logr.Logger) error {
+	if instance.Spec.Apply {
+		obj := instance.Spec.Object
+		found := obj.DeepCopy()
+		err := r.client.Get(context.TODO(), types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, found)
```
mrogers950, line 148: Why make a deep copy of the object before getting?
Also, will this client work the same for a non-namespaced object?

JAORMX: > Why make a deep copy of the object before getting?

Because it copies the GVK which is needed in order to get the object. Can't use unstructured or runtime.Object for this one straight.

> Also, will this client work the same for a non-namespaced object?

Yep

pkg/controller/complianceremediation/complianceremediation_controller.go
jhrozek: That's a good question. Do you reckon we would have many Kinds of remediation objects? if not, we could have a switch that would branch off to comparing the types in a per-type manner. For CMs, we could just compare the key/value pairs. I'm not sure what would be the good fallback, maybe we could just `DeepCompare` a pair of `Spec` attributes or just do nothing.

Of course the downside is that the code would no longer be completely generic..

JAORMX: uhm... Maybe we should just leave this as-is for now. And start enhancing once we get more concrete examples (when the need comes). But yeah, so far, it seems that it'll be what we need to do, do comparison depending on the type.

PR #237, opened 2020-04-30
Upgrade to Operator sdk v0.17.0

JAORMX:
This upgrades to operator-sdk v0.17.0 which required some changes to the manager's main entry point.

The kube dependencies needed to be upgraded too, as well as some flags for our tests.


JAORMX:
/hold

This seems to have a bunch of timeouts

jhrozek:
@JAORMX does this just need a rebase? It would be nice to be able to use the latest sdk..

PR #240, opened 2020-05-06
Implement re-scanning

JAORMX:
This makes the annotation `compliance.openshift.io/rescan` execute a
rescan if it's encountered. Note that this only happens in the "DONE"
phase, and if it's found, it also deletes the previous pods and secrets
before going back to the "PENDING" phase. This is done in order to start
from a blank slate and avoid errors.

On the other hand, this introduces the `currentIndex` field to the
ComplianceScan's status. This represents the number of scans that have
been executed, and serves as a directory for the results in the PVC

`ChechResult` objects are updated once the scan is re-run, so fetching the results for the same scan should give you the latest updates.

One details is that the ConfigMaps that contained the results need to be deleted in order to run the scan again. So this happens only once the re-scan is detected and applied.

### DIFF
--- a/cmd/remediation-aggregator/main.go
+++ b/cmd/remediation-aggregator/main.go
@@ -120,7 +120,7 @@ func getScanConfigMaps(clientset *kubernetes.Clientset, scan, namespace string) 
 
 	// Look for configMap with this scan label
 	listOpts := metav1.ListOptions{
-		LabelSelector: fmt.Sprintf("compliance-scan=%s", scan),
+		LabelSelector: fmt.Sprintf("%s=%s", compv1alpha1.ComplianceScanIndicatorLabel, scan),
 	}
 
 	cMapList, err = clientset.CoreV1().ConfigMaps(namespace).List(listOpts)
--- a/cmd/remediation-aggregator/main.go
+++ b/cmd/remediation-aggregator/main.go
@@ -98,6 +98,8 @@ func createCrClient(config *rest.Config) (*complianceCrClient, error) {
 		&compv1alpha1.ComplianceCheckResult{})
 	scheme.AddKnownTypes(compv1alpha1.SchemeGroupVersion,
 		&metav1.CreateOptions{})
+	scheme.AddKnownTypes(compv1alpha1.SchemeGroupVersion,
+		&metav1.UpdateOptions{})
 
 	client, err := runtimeclient.New(config, runtimeclient.Options{
 		Scheme: scheme,
@@ -196,7 +198,7 @@ type compResultIface interface {
 	runtime.Object
 }
 
-func createOneResult(crClient *complianceCrClient, owner metav1.Object, labels map[string]string, res compResultIface) error {
+func createOrUpdateOneResult(crClient *complianceCrClient, owner metav1.Object, labels map[string]string, exists bool, res compResultIface) error {
 	kind := res.GetObjectKind()
 
 	if err := controllerutil.SetControllerReference(owner, res, crClient.scheme); err != nil {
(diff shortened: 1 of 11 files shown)


pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -364,6 +365,15 @@ func (r *ReconcileComplianceScan) phaseDoneHandler(instance *compv1alpha1.Compli
 			log.Error(err, "Cannot delete CA secret")
 			return reconcile.Result{}, err
 		}
+
+		if instance.NeedsRescan() {
+			instanceCopy := instance.DeepCopy()
+			instanceCopy.Status.Phase = compv1alpha1.PhasePending
+			instanceCopy.Status.CurrentIndex = instance.Status.CurrentIndex + 1
```
jhrozek, line 377: Do you know offhand how does golang handle integer overflow? I guess we won't go over INT_MAX anytime soon, but the C programmer in me doesn't like increasing an int without limits anyway :-)
(in C, the typical way to solve this might be if increased < previous { increased = 0 } or similar)

cmd/resultscollector/main.go
```diff
@@ -31,6 +31,7 @@ import (
 
 	backoff "github.com/cenkalti/backoff/v3"
 	"github.com/dsnet/compress/bzip2"
+	libgocrypto "github.com/openshift/library-go/pkg/crypto"
```
jhrozek, line 34: Does this implement the FIPS validated crypto on RH platforms?

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: Good question, gotta research it.

cmd/resultscollector/main.go
JAORMX: Supposedly. Note that I didn't introduce the usage of this library, merely moved the import.

JAORMX:
> The only conceptual question I have is about archiving or even diffing the results. Is fetching the ARF results from previous scans enough or should we implement some diff versus previous (and maybe only the previous) run?

So, the idea with this commit is to archive them all. I'm not sure if a diff of the ARFs is necessary. I would expect folks to fetch the ARF and store them in an artifact repo somewhere so they can eventually take them into use.

JAORMX:
One thing that we gotta start thinking about is ARF artifact rotation... as rescans happen, they will eventually (and very likely) fill up space in the PVC, we probably gotta rotate them to avoid this.

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: Addressed

cmd/resultserver/main.go
```diff
@@ -71,6 +75,20 @@ func getValidStringArg(cmd *cobra.Command, name string) string {
 	return val
 }
 
+func ensureDir(path string) error {
+	_, err := os.Stat(path)
```
jhrozek, line 79: I guess this is fine if you know only one resultserver for a path would be starting at the same time, but still isn't it better to try to create the directory and ignore EEXIST?

cmd/remediation-aggregator/main.go
```diff
@@ -210,9 +212,14 @@ func createOneResult(crClient *complianceCrClient, owner metav1.Object, labels m
 	fmt.Printf("Creating %s:%s\n", kind.GroupVersionKind().Kind, name)
 
 	err := backoff.Retry(func() error {
-		err := crClient.client.Create(context.TODO(), res)
+		var err error
```
jhrozek, line 215: similar to the question about creating the directory, wouldn't it be simpler to try creating the object and turn the create into update if create returns that the object already exists?

JAORMX: the Get is needed because we need the object metadata (such as the resourceVersion) else the update fails.

cmd/resultserver/main.go
JAORMX: that's a good suggestion. Not sure how EEXIST is exposed to golang though,

JAORMX: Done

JAORMX:
@jhrozek addressed.

PR #244, opened 2020-05-11
Generate events to be read by RHACM

JAORMX:
This generates events for the Compliancesuite once the result is
available, and if the suite is owned by a Policy object, it generates an
event for that too.

Co-Authored-By: Jakub Hrozek <jhrozek@redhat.com>

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -19,6 +19,7 @@ rules:
   - ""
   resources:
   - configmaps  # The log collecting sidecar uses configmaps to store results
+  - events
   verbs:
   - create      # The sidecar must create the configmap to store results in
   - get
--- a/pkg/apis/compliance/v1alpha1/compliancesuite_types.go
+++ b/pkg/apis/compliance/v1alpha1/compliancesuite_types.go
@@ -116,3 +116,8 @@ func (s *ComplianceSuite) LowestCommonResult() ComplianceScanStatusResult {
 
 	return lowestCommonResult
 }
+
+func (s *ComplianceSuite) IsResultAvailable() bool {
+	result := s.LowestCommonResult()
+	return result != "" && result != ResultNotAvailable
+}
--- a/pkg/controller/common/constants.go
+++ b/pkg/controller/common/constants.go
@@ -1,18 +1,33 @@
 package common
 
 import (
+	"os"
+
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 )
 
 var complianceOperatorNamespace = "openshift-compliance"
 
 func init() {
-	ns, err := k8sutil.GetOperatorNamespace()
-	if err == nil {
-		complianceOperatorNamespace = ns
+	if isRunModeLocal() {
+		ns, ok := os.LookupEnv("WATCH_NAMESPACE")
+		if ok {
+			complianceOperatorNamespace = ns
+		}
+	} else {
+		ns, err := k8sutil.GetOperatorNamespace()
+		if err == nil {
+			complianceOperatorNamespace = ns
+		}
 	}
 }
 
+func isRunModeLocal() bool {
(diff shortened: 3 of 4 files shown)


jhrozek:
This mostly works for me.
First change I had to do was in order for the resultserver to run when the operator was started on the foreground in a non-default namespace:
```
diff --git a/pkg/controller/common/constants.go b/pkg/controller/common/constants.go
index f1abcfb..1103e6c 100644
--- a/pkg/controller/common/constants.go
+++ b/pkg/controller/common/constants.go
@@ -2,17 +2,29 @@ package common
 
 import (
        "github.com/operator-framework/operator-sdk/pkg/k8sutil"
+       "os"
 )
 
 var complianceOperatorNamespace = "openshift-compliance"
 
 func init() {
-       ns, err := k8sutil.GetOperatorNamespace()
-       if err == nil {
-               complianceOperatorNamespace = ns
+       if isRunModeLocal() {
+               ns, ok := os.LookupEnv("WATCH_NAMESPACE")
+               if ok {
+                       complianceOperatorNamespace = ns
+               }
+       } else {
+               ns, err := k8sutil.GetOperatorNamespace()
+               if err == nil {
+                       complianceOperatorNamespace = ns
+               }
        }
 }
 
+func isRunModeLocal() bool {
+       return os.Getenv(k8sutil.ForceRunModeEnv) == string(k8sutil.LocalRunMode)
+}
+
 func GetComplianceOperatorNamespace() string {
        return complianceOperatorNamespace
 }
```

Second, the reason the events were suppressed with the condition that returns if the phase is the same is that before calling the event emit, we update the suite phase to match the scan :-) Maybe something like this is the simplest solution:
```
diff --git a/pkg/controller/compliancesuite/compliancesuite_controller.go b/pkg/controller/compliancesuite/compliancesuite_controller.go
index f171724..09445f3 100644
--- a/pkg/controller/compliancesuite/compliancesuite_controller.go
+++ b/pkg/controller/compliancesuite/compliancesuite_controller.go
@@ -152,9 +152,6 @@ func (r *ReconcileComplianceSuite) reconcileScanStatus(suite *compv1alpha1.Compl
                                logger.Error(err, "Could not update scan status")
                                return err
                        }
-                       if r.recorder != nil && suite.IsResultAvailable() {
-                               r.generateEventsForSuite(suite, &scanStatusWrap, scan, logger)
-                       }
                        return nil
                }
        }
@@ -185,16 +182,18 @@ func (r *ReconcileComplianceSuite) updateScanStatus(suite *compv1alpha1.Complian
        suite.Status.AggregatedResult = suite.LowestCommonResult()
        logger.Info("Updating scan status", "ComplianceScan.Name", scanStatusWrap.Name, "ComplianceScan.Phase", scanStatusWrap.Phase)
 
-       return r.client.Status().Update(context.TODO(), suite)
+       err := r.client.Status().Update(context.TODO(), suite)
+       if err != nil {
+               return err
+       }
+
+       if r.recorder != nil && suite.IsResultAvailable() {
+               r.generateEventsForSuite(suite, scanStatusWrap, scan, logger)
+       }
+       return nil
 }
 
 func (r *ReconcileComplianceSuite) generateEventsForSuite(suite *compv1alpha1.ComplianceSuite, scanStatusWrap *compv1alpha1.ComplianceScanStatusWrapper, scan *compv1alpha1.ComplianceScan, logger logr.Logger) {
-       // if yes, update it, if the status differs
-       //if scanStatusWrap.Phase == scan.Status.Phase {
-       //      logger.Info("Not generating event for scan, the phase is the same", "ComplianceScan.Name", scanStatusWrap.Name, "ComplianceScan.Phase", scanStatusWrap.Phase)
-       //      return
-       //}
-
        logger.Info("Generating event for scan", "ComplianceScan.Name", scanStatusWrap.Name, "ComplianceScan.Phase", scanStatusWrap.Phase)
 
        // Event for Suite
```

jhrozek:
btw it even seems the violation goes away when I remediate :-)

PR #246, opened 2020-05-13
Introduce scheduled scans

JAORMX:
This introduces the `schedule` parameter to the ComplianceSuite object.
this parameter effectively creates a cronjob that will re-trigger scans
with the cadence that's configured by the parameter.

Note that this will only happen once the scans have initial results.

If the schedule is empty (which is the default) no rerunner cronjob will
be created.

Note also that the rerunner command is part of the main operator binary.

### DIFF
--- a/cmd/manager/common.go
+++ b/cmd/manager/common.go
@@ -7,12 +7,11 @@ import (
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
-	compv1alpha1 "github.com/openshift/compliance-operator/pkg/apis/compliance/v1alpha1"
+	compapis "github.com/openshift/compliance-operator/pkg/apis"
 )
 
 type complianceCrClient struct {
@@ -25,23 +24,7 @@ func createCrClient(config *rest.Config) (*complianceCrClient, error) {
 
 	corev1.AddToScheme(scheme)
 	mcfgv1.AddToScheme(scheme)
-
-	scheme.AddKnownTypes(compv1alpha1.SchemeGroupVersion,
-		&compv1alpha1.ComplianceRemediation{})
-	scheme.AddKnownTypes(compv1alpha1.SchemeGroupVersion,
-		&compv1alpha1.ComplianceScan{})
-	scheme.AddKnownTypes(compv1alpha1.SchemeGroupVersion,
-		&compv1alpha1.ComplianceScanList{})
-	scheme.AddKnownTypes(compv1alpha1.SchemeGroupVersion,
-		&compv1alpha1.ComplianceSuite{})
-	scheme.AddKnownTypes(compv1alpha1.SchemeGroupVersion,
-		&compv1alpha1.ComplianceCheckResult{})
-	scheme.AddKnownTypes(compv1alpha1.SchemeGroupVersion,
-		&metav1.CreateOptions{})
-	scheme.AddKnownTypes(compv1alpha1.SchemeGroupVersion,
-		&metav1.UpdateOptions{})
-	scheme.AddKnownTypes(compv1alpha1.SchemeGroupVersion,
-		&metav1.ListOptions{})
+	compapis.AddToScheme(scheme)
 
 	client, err := runtimeclient.New(config, runtimeclient.Options{
 		Scheme: scheme,
--- a/cmd/manager/common.go
+++ b/cmd/manager/common.go
@@ -1,0 +1,66 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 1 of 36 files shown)


pkg/apis/compliance/v1alpha1/compliancesuite_types.go
```diff
@@ -4,6 +4,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// ComplianceSuiteIndicatorLabel indicates that an object (normally
```
jhrozek, line 7: Typo? Should the comment say `SuiteLabel` ?

pkg/controller/compliancesuite/suitererunner.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		msg := "ComplianceSuite's schedule is wrongly formatted"
+		r.recorder.Event(
+			suite,
+			corev1.EventTypeWarning,
+			"ScheduleFormatError",
+			msg,
+		)
+		logger.Info("WARNING: " + msg)
+		// TODO(jaosorior): Should we surface this to the CRD?
```
jhrozek, line 38: I would say yes. The admin has expressed intent to schedule the scans and the controller cannot converge to the state because the input was wrong. I would say this should be expressed in the status.

pkg/apis/compliance/v1alpha1/compliancesuite_types.go
JAORMX: It should.

pkg/controller/compliancesuite/suitererunner.go
JAORMX: The reason I didn't do this is that we currently don't have a field in the status where such warnings or errors would appear. Right now it's just an event. The other thing is that this is a non-fatal issue, a scan will still be executed, this just won't have scheduled scans afterwards... the Aggregated phase and results are probably not the places to put this.. so... any suggestions on what to add to reflect this?

pkg/apis/compliance/v1alpha1/compliancesuite_types.go
JAORMX: Fixed.

pkg/controller/compliancesuite/suitererunner.go
jhrozek: Oh, the suite status we have now is just an aggregate of scan statuses..but still, couldn't we (ab)use the aggregated phase and result and surface the error there?
I was thinking about this: if a schedule is specified, but cannot be parsed, then don't schedule new scans and kill the existing ones, then surface the error to the aggregated error and status fields.

JAORMX: OK, that could be a way to go.

JAORMX: Mind if I do this in a subsequent PR?

jhrozek: sure, let me lgtm this patchset so that it merges when CI finishes

JAORMX:
/retest

lol, crap, the CI cluster failed:

```
The connection to the server api.ci-op-y36czp7k-d825e.origin-ci-int-aws.dev.rhcloud.com:6443 was refused - did you specify the right host or port?
```

cmd/manager/suitererunner.go
```diff
@@ -0,0 +1,89 @@
...
+		if scanCopy.Annotations == nil {
+			scanCopy.Annotations = make(map[string]string)
+		}
+		scanCopy.Annotations[compv1alpha1.ComplianceScanRescanAnnotation] = ""
+
+		fmt.Printf("Re-running ComplianceScan '%s'\n", scanCopy.Name)
+		err := conf.client.client.Update(context.TODO(), scanCopy)
+		if err != nil {
+			fmt.Printf("Error while updating scan '%s', err: %s\n", scanCopy.Name, err)
+			os.Exit(1)
```
mrogers950, line 86: Resource version conflicts might happen fairly often here. We might want a second Update attempt on error instead of aborting the list.. unless we can guarantee the re-runner will be launched again in this situation.

deploy/crds/compliance.openshift.io_compliancescans_crd.yaml
```diff
@@ -70,6 +70,9 @@ spec:
                 a specific rule. Note that when leaving this empty, the scan will
                 check for all the rules for a specific profile.
               type: string
+            scanType:
```
mrogers950, line 73: I thought this went in as part of the platform scan patch, from running `make generate` ??

cmd/manager/suitererunner.go
JAORMX: It will be launched again in this situation. The restartPolicy is "restartOnFailure".

deploy/crds/compliance.openshift.io_compliancescans_crd.yaml
JAORMX: that's just how it turned up. Did a make generate and it changed the place of this :/

cmd/manager/suitererunner.go
JAORMX: I could add a retry/backoff here to address this in a subsequent PR if you think that would be beneficial.

tests/e2e/helpers.go
```diff
@@ -219,6 +219,52 @@ func waitForScanStatus(t *testing.T, f *framework.Framework, namespace, name str
...
 
+// waitForReScanStatus will poll until the compliancescan that we're lookingfor reaches a certain status for a re-scan, or until
+// a timeout is reached.
+func waitForReScanStatus(t *testing.T, f *framework.Framework, namespace, name string, targetStatus compv1alpha1.ComplianceScanStatusPhase) error {
+	foundScan := &compv1alpha1.ComplianceScan{}
+	// unset initial index
+	var scanIndex int64 = -1
+	var lastErr error
+	// retry and ignore errors until timeout
+	timeouterr := wait.Poll(retryInterval, timeout, func() (bool, error) {
```
mrogers950, line 230: nit: camelCase timeouterr

cmd/manager/common.go
```diff
@@ -0,0 +1,49 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
```
mrogers950, line 7: nit: import ordering

JAORMX: this was generated by VSCode's import ordering linter :/

PR #249, opened 2020-05-14
Resilience: retry updating scans in rerunner

JAORMX:
Even though the current failure policy is "restartOnFailure", it's
beneficial to retry scan updates if there's an issue given that we
wouldn't need to re-iterate the scans again.

So this introduces a retry/backoff to each update and if there's an
issue, it re-fetches the scan, or simply retries.

### DIFF
--- a/cmd/manager/suitererunner.go
+++ b/cmd/manager/suitererunner.go
@@ -5,12 +5,17 @@ import (
 	"fmt"
 	"os"
 
+	backoff "github.com/cenkalti/backoff/v3"
 	compv1alpha1 "github.com/openshift/compliance-operator/pkg/apis/compliance/v1alpha1"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


cmd/manager/suitererunner.go
```diff
@@ -73,16 +78,40 @@ func RerunSuite(cmd *cobra.Command, args []string) {
...
-		scanCopy := scan.DeepCopy()
-		if scanCopy.Annotations == nil {
-			scanCopy.Annotations = make(map[string]string)
-		}
-		scanCopy.Annotations[compv1alpha1.ComplianceScanRescanAnnotation] = ""
+		currentScan := &scan
+		key := types.NamespacedName{Name: scan.GetName(), Namespace: scan.GetNamespace()}
+		err := backoff.Retry(func() error {
+			var scanCopy *compv1alpha1.ComplianceScan
+			if currentScan == nil {
```
mrogers950, line 85: I think we will need to fetch again regardless of the error from the previous attempt. In the non-expire/conflict case the version of the resource we previously got from the list would be old still.

PR #253, opened 2020-05-19
Use explicit roles/clusterroles for api-resource-collector

JAORMX:
The OLM doesn't allow us to use pre-existing roles/clusterroles, nor
use aggregated roles either. So we need to define the roles explicitly.

See: https://github.com/operator-framework/operator-lifecycle-manager/issues/732


jhrozek:
Since this is blocking the release, I'm going to lgtm the PR so that we can do the release, Ozz showed me the result in a tmate session. We can iterate on the patch later.

/lgtm

PR #256, opened 2020-05-20
Add friendly rule name as annotation to check result

JAORMX:
This is an attempt at correlating check results with profile rules. Both
objects will contain the same annotations that indicate the same name.

The reason they're set as annotations and not labels is because the
length that the IDs may have may be more than what labels can handle.

Also, remove useless nil check from remediation-aggregator
Add friendly rule name as annotation to check result

### DIFF
--- a/cmd/remediation-aggregator/main.go
+++ b/cmd/remediation-aggregator/main.go
@@ -198,7 +198,7 @@ type compResultIface interface {
 	runtime.Object
 }
 
-func createOrUpdateOneResult(crClient *complianceCrClient, owner metav1.Object, labels map[string]string, exists bool, res compResultIface) error {
+func createOrUpdateOneResult(crClient *complianceCrClient, owner metav1.Object, labels map[string]string, annotations map[string]string, exists bool, res compResultIface) error {
 	kind := res.GetObjectKind()
 
 	if err := controllerutil.SetControllerReference(owner, res, crClient.scheme); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/remediation-aggregator/main.go
```diff
@@ -207,6 +207,9 @@ func createOrUpdateOneResult(crClient *complianceCrClient, owner metav1.Object,
 	}
 
 	res.SetLabels(labels)
+	if annotations != nil {
```
jhrozek, line 210: I don't understand this check. Is it mean to prevent from setting nil annotations on updates?

JAORMX: This function is shared. It creates both complianceCheckResults and Remediations. Remediations don't have annotations.

jhrozek: Yes, I guess what I'm trying to ask is why is not OK to just call `SetLabels(nil)`

JAORMX: I guess it would work. I just didn't wanna touch the annotations if there really isn't any. As the `SetAnnotations` function will create an empty map if it's `nil`: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/meta/meta.go#L585

jhrozek: OK, that makes sense. I ended up looking elsewhere (`https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/meta.go#L166`) where it was just a simple assignment.

PR #257, opened 2020-05-20
e2e: Extend the platform test

mrogers950:
Previously we just tested the default state of the cluster for a compliant
platform result. This follows it up by installing an HTPasswd IDP and expecting
a non-compliant result.


JAORMX:
/retest

unrelated CI issue

PR #258, opened 2020-05-22
Decouple resources from CRs

JAORMX:
this decouples the resources the operator creates with the resources it tracks. Now the operator can listen to whatever namespace(s) are specified by the `WATCH_NAMESPACE` environment variable, and it'll be able to react to those changes.

The operator's namespace is enforced as part of the `WATCH_NAMESPACE` set in order to get the cached client to track the operator's resources as well.

resource cleanup is now handled via finalizers for both the `ComplianceSuite` and `ComplianceScan` resources. This way, deletion is blocked until the resources as cleaned up, and then the finalizer is removed.

### DIFF
--- a/pkg/controller/compliancescan/compliancescan_controller.go
+++ b/pkg/controller/compliancescan/compliancescan_controller.go
@@ -272,7 +272,7 @@ func (r *ReconcileComplianceScan) phaseRunningHandler(instance *compv1alpha1.Com
 
 		if running {
 			// The platform scan pod is still running, go back to queue.
-			return reconcile.Result{}, err
+			return reconcile.Result{Requeue: true, RequeueAfter: requeueAfterDefault}, err
 		}
 	default: // ScanTypeNode
 		if nodes, err = getTargetNodes(r, instance); err != nil {
--- a/pkg/controller/compliancescan/compliancescan_controller.go
+++ b/pkg/controller/compliancescan/compliancescan_controller.go
@@ -117,6 +117,22 @@ func (r *ReconcileComplianceScan) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
+	// examine DeletionTimestamp to determine if object is under deletion
+	if instance.ObjectMeta.DeletionTimestamp.IsZero() {
+		// The object is not being deleted, so if it does not have our finalizer,
+		// then lets add the finalizer and update the object. This is equivalent
+		// registering our finalizer.
+		if !common.ContainsFinalizer(instance.ObjectMeta.Finalizers, compv1alpha1.ScanFinalizer) {
+			instance.ObjectMeta.Finalizers = append(instance.ObjectMeta.Finalizers, compv1alpha1.ScanFinalizer)
+			if err := r.client.Update(context.TODO(), instance); err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+	} else {
+		// The object is being deleted
+		return r.scanDeleteHandler(instance, reqLogger)
+	}
+
 	// At this point, we make a copy of the instance, so we can modify it in the functions below.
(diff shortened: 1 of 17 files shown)


JAORMX:
/hold

This might not be as urgent after all. We might get around this by automating the installation while enrolling the managed cluster to ACM. This is hacky but works for demo purposes.

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -286,7 +286,7 @@ func (r *ReconcileComplianceScan) phaseRunningHandler(instance *compv1alpha1.Com
 			}
 			if running {
 				// at least one pod is still running, just go back to the queue
-				return reconcile.Result{}, err
+				return reconcile.Result{Requeue: true, RequeueAfter: requeueAfterDefault}, err
```
mrogers950, line 305: The commit msg mentions that the controller no longer listens to pod changes, but c.Watch() is still using Pod as a secondary

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -438,6 +453,45 @@ func (r *ReconcileComplianceScan) phaseDoneHandler(instance *compv1alpha1.Compli
...
 
+func (r *ReconcileComplianceScan) scanDeleteHandler(instance *compv1alpha1.ComplianceScan, logger logr.Logger) (reconcile.Result, error) {
+	if common.ContainsFinalizer(instance.ObjectMeta.Finalizers, compv1alpha1.ScanFinalizer) {
+		scanToBeDeleted := instance.DeepCopy()
+
+		// Force remove rescan annotation since we're deleting the scan
+		if scanToBeDeleted.NeedsRescan() {
+			delete(scanToBeDeleted.Annotations, compv1alpha1.ComplianceScanRescanAnnotation)
+		}
+		// Force debug to be false so we actually remove dependent objects
```
mrogers950, line 464: Why does debug stop object removal?

JAORMX: because we need it for CI, to be able to fetch the pods' logs. This would be useful for debugging and testing in a deployment for the same reasoning. This is not desirable if you're actually deleting an object; then you REALLY want to delete everything.

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: I'll delete that. It tries to listen for pods (as long as their owner references are properly set), but none will be tracked as they now aren't owned by the compliancescan

pkg/controller/compliancescan/compliancescan_controller.go
mrogers950: Makes sense, adding some more of this context to the comment would be helpful

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -103,6 +103,22 @@ func (r *ReconcileComplianceSuite) Reconcile(request reconcile.Request) (reconci
 		return reconcile.Result{}, err
 	}
 
+	// examine DeletionTimestamp to determine if object is under deletion
+	if suite.ObjectMeta.DeletionTimestamp.IsZero() {
```
mrogers950, line 107: Would be good to function-ize since it's shared with compliancescan_controller

mrogers950: Although that might be overkill since it's working on a different object.. up to you

JAORMX: I think I'll leave this as-is

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: Done.

PR #266, opened 2020-06-01
aggregator: merge to main binary

JAORMX:
This merges the aggregator to be yet another subcommand of the main binary.
Effectively making this just part of the main operator image.

Also, the aggregator used to use a combination of the kube clientset interface
and the controller runtime client. This commit converges to using only
the controller runtime's client since that's what the unified binary
will use.


JAORMX:
/test all

CI having issues

PR #271, opened 2020-06-03
Introduce functionality imported from profile operator

JAORMX:
The compliance-profile-operator[1] served as a PoC to be the UI for the
profiles and creating tailoring files. We used this to demonstrate RHACM
integration. This commit imports the functionality that came from that
operator and merges it into this one.

[1] https://github.com/JAORMX/compliance-profile-operator


deploy/crds/compliance.openshift.io_v1alpha1_rule_cr.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: compliance.openshift.io/v1alpha1
```
jhrozek, line 1: Can we remove this CR? It's not like it's an example CR that people are supposed to create.

cmd/manager/profileparser.go
```diff
@@ -0,0 +1,310 @@
...
+	Long:  `The profileparser reads a data stream file and generates profile objects from it.`,
+	Run:   runProfileParser,
+}
+
+func init() {
+	defineRerunnerFlags(profileparserCmd)
+}
+
+// XMLDocument is a wrapper that keeps the interface XML-parser-agnostic
+type XMLDocument struct {
```
jhrozek, line 51: It looks like we already have this type in the ARF parser, can we reuse it instead of duplicating?

PR #279, opened 2020-06-05
Only mark scan as COMPLIANT if there is at least one passing result

jhrozek:
I was unable to run the tests locally for some reason, so I'm submitting the WIP PR to see what CI has to say about it

PR #286, opened 2020-06-10
Reconcile differing remediations

jhrozek:
Instead of always erroring out if any remediations or rule results
differ, be a bit more lenient and create the checks anyway, but clearly
mark that they are inconsistent. This has the advantage that the
aggregator always runs to completion instead of failing and falling into
a CrashLoopBackoff.

The new algorithm works like this:
    - when processing a batch of checks from a configMap, compare them
      with checks from previous batches. Each check is stored in either
      of two lists: consistent and inconsistent
        - the consistent checks are created as before
        - the inconsistent checks are created with a special state
        "Inconsistent" and labeled so that it is easy to find them
    - for the inconsistent checks, try to find the most common state
      (where at least 60% of sources agree on a state)
        - if there is a common state, we add it to annotations and then
          list the differing states separately per source
        - if we couldn't find a common state, list all states per source
          in annotations

TODO:
    - tests, probably use spew for e2e tests to create the compliant
      state on one of the nodes manually
    - mark the whole scan and suite as inconsistent when there are any
      inconsistent checks at all
        - create an Event in this case
    - if something else than the check result state differs, create the
      result with ERROR
    - amend creating remediations for inconsistent results: Only create
      remediations for checks that are either consistent or, if they are
      inconsistent, all the check results are one of pass/info/fail.
    - the old diff code is still used in tests, remove that

### DIFF
--- a/cmd/manager/aggregator.go
+++ b/cmd/manager/aggregator.go
@@ -122,18 +122,23 @@ func readCompressedData(compressed string) (*bzip2.Reader, error) {
 	return bzip2.NewReader(compressedReader, &bzip2.ReaderConfig{})
 }
 
-func parseResultRemediations(scheme *runtime.Scheme, scanName, namespace string, content *utils.XMLDocument, cm *v1.ConfigMap) ([]*utils.ParseResult, error) {
+// parseResultRemediations parses scan results from a configMap with the help of DS provided in the
+// content parameter.
+// Returns a triple of (array-of-ParseResults, source, error) where source identifies the entity whose
+// scan produced this configMap -- typically a nodeName for node scans. For platform scans, the source
+// is empty. The source is used later when reconciling inconsistent results
+func parseResultRemediations(scheme *runtime.Scheme, scanName, namespace string, content *utils.XMLDocument, cm *v1.ConfigMap) ([]*utils.ParseResult, string, error) {
 	var scanReader io.Reader
 
 	_, ok := cm.Annotations[configMapRemediationsProcessed]
 	if ok {
 		fmt.Printf("ConfigMap %s already processed\n", cm.Name)
-		return nil, nil
+		return nil, "", nil
 	}
 
 	cmScanResult, ok := cm.Data["results"]
 	if !ok {
-		return nil, fmt.Errorf("no results in configmap %s", cm.Name)
+		return nil, "", fmt.Errorf("no results in configmap %s", cm.Name)
 	}
 
 	_, ok = cm.Annotations[configMapCompressed]
@@ -141,7 +146,7 @@ func parseResultRemediations(scheme *runtime.Scheme, scanName, namespace string,
 		fmt.Printf("Results are compressed\n")
 		scanResult, err := readCompressedData(cmScanResult)
 		if err != nil {
-			return nil, err
+			return nil, "", err
 		}
 		defer scanResult.Close()
 		scanReader = scanResult
@@ -149,7 +154,11 @@ func parseResultRemediations(scheme *runtime.Scheme, scanName, namespace string,
 		scanReader = strings.NewReader(cmScanResult)
 	}
 
-	return utils.ParseResultsFromContentAndXccdf(scheme, scanName, namespace, content, scanReader)
+	// This would return an empty string for a platform check that is handled later explicitly
+	nodeName := cm.Annotations["openscap-scan-result/node"]
+
+	table, err := utils.ParseResultsFromContentAndXccdf(scheme, scanName, namespace, content, scanReader)
+	return table, nodeName, err
 }
 
 func getScanResult(cm *v1.ConfigMap) (compv1alpha1.ComplianceScanStatusResult, string) {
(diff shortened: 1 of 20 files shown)


cmd/manager/aggregator.go
```diff
@@ -112,34 +112,38 @@ func readCompressedData(compressed string) (*bzip2.Reader, error) {
 	return bzip2.NewReader(compressedReader, &bzip2.ReaderConfig{})
 }
 
-func parseResultRemediations(scheme *runtime.Scheme, scanName, namespace string, content *utils.XMLDocument, cm *v1.ConfigMap) ([]*utils.ParseResult, error) {
+func parseResultRemediations(scheme *runtime.Scheme, scanName, namespace string, content *utils.XMLDocument, cm *v1.ConfigMap) ([]*utils.ParseResult, string, error) {
```
JAORMX, line 115: Mind adding a comment about the output of this function? it's not too long but it would make it easier for the reader.

pkg/utils/remediation_diff.go
```diff
@@ -1,12 +1,237 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	consistentList := make([]*ParseResultContextItem, len(prCtx.consistent))
+	idx := 0
+
+	for _, item := range prCtx.consistent {
+		consistentList[idx] = item
+		idx++
```
JAORMX, line 163: why add them with the index if you could just append them?

```
consistentList = append(consistentList, item)
```

jhrozek: I don't honestly know why I wrote the code the way I did. Fixed.

jhrozek:
I think the current PR is complete and it works for me locally but still we should merge #279 first as this PR depends on 279 (and 279 is part of this PR at the moment anyway)

jhrozek:
`Error: Unable to find matching route for Route Table (rtb-0ece4133468f7074d) and destination CIDR block (0.0.0.0/0)."`
/retest

deploy/operator.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: compliance-operator
       containers:
         - name: compliance-operator
-          image: "quay.io/compliance-operator/compliance-operator:latest"
+          image: "image-registry.openshift-image-registry.svc:5000/openshift-compliance/compliance-operator:latest"
```
JAORMX, line 18: This is why this test is failing with timeouts

deploy/operator.yaml
```diff
@@ -35,7 +35,7 @@ spec:
               # Hardcoding this temporarily until its propagated to CI
               value: "quay.io/compliance-operator/openscap-ocp:1.3.3"
             - name: OPERATOR_IMAGE
-              value: "quay.io/compliance-operator/compliance-operator:latest"
+              value: "image-registry.openshift-image-registry.svc:5000/openshift-compliance/compliance-operator:latest"
```
JAORMX, line 38: and this.

jhrozek:
/hold
because there is am extra debugging patch

jhrozek:
/hold cancel
this should hopefully work now, let me point out the two places that were changed

pkg/utils/remediation_diff.go
```diff
@@ -1,44 +1,287 @@
...
+		annotations[compv1alpha1.ComplianceCheckResultMostCommonAnnotation] = string(mostCommonState)
+	}
+
+	return annotations, createRemediation
+}
+
+func mostCommonState(inconsistent []*ParseResultContextItem) (compv1alpha1.ComplianceCheckStatus, bool) {
+	statusCounter := make(map[compv1alpha1.ComplianceCheckStatus]int)
+	for _, check := range inconsistent {
+		statusCounter[check.CheckResult.Status] = statusCounter[check.CheckResult.Status] + len(check.sources)
```
jhrozek, line 265: Here we were always increasing by one which doesn't work if we first get several passing results and then one failing.

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -747,6 +752,22 @@ func gatherResults(r *ReconcileComplianceScan, instance *compv1alpha1.Compliance
 		}
 	}
 
+	// If there are any inconsistent results, always just return
+	// the state as inconsistent unless there was an error earlier
+	var checkList compv1alpha1.ComplianceCheckResultList
+	checkListOpts := client.MatchingLabels{
+		compv1alpha1.ComplianceCheckInconsistentLabel: "",
+		compv1alpha1.ScanLabel:                        instance.Name,
```
jhrozek, line 760: This was missing, thus any inconsistent scan in the namespaces marked all subsequent scans as inconsistent.

jhrozek:
First, it's odd that the test failed, I was tailing them in the console and clearly saw the "PASS" result..I wonder if prow restarts the jobs if something is merged to master in the meantime? Anyway, the last error seems to be quite unrelated:
```
        --- FAIL: TestE2E/Parallel_tests/TestScheduledSuite (1391.14s)
...
        helpers.go:63: 2020-06-29T16:57:15Z: Waiting until the raw result checker workload is done: osdk-e2e-19bfc00b-d819-4d71-af7b-d58c6b6c7f40/rotation-checker
...
          helpers.go:63: 2020-06-29T17:12:16Z: Pod not done. retrying: %!s(<nil>)
          helpers.go:63: 2020-06-29T17:12:16Z: Pod not done. retrying: %!s(<nil>)
          helpers.go:182: timed out waiting for the condition
```

jhrozek:
Looks like the checker pod is racing for the PVC with the results-server:
```
  Warning  FailedAttachVolume  97s        attachdetach-controller  Multi-Attach error for volume "pvc-182244b9-c2a3-40e9-aa4d-e3d1e8d34b4e" Volume is already used by pod(s) test-scheduled-suite-workers-scan-rs-7977bcb598-9kml2
```

jhrozek:
/retest
I'm rescheduling the test because I was toying around in the CI environment due to the error above and might have tainted the test run (it would have passed while it should fail), so I think it's fair to retrigger and let prow merge the PR only when it's really passing on its own.

PR #299, opened 2020-06-15
Update machine-config-operator to latest code

JAORMX:
This also changes the test code to use the new format in MC; with the
ignition configuration being a raw extension as opposed to a concrete
type.


pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -336,15 +336,15 @@ func getAppliedMcRemediations(r *ReconcileComplianceRemediation, rem *compv1alph
...
+	mergedMc, err := mcfgcommon.MergeMachineConfigs(configs, "")
 
-	if mergedMc == nil {
+	// FIXME(jaosorior): Handle errors
+	if err != nil {
 		return nil
 	}
 
-	// NOTE(jaosorior): If no version was set (for some reason) lets just add a default
-	if mergedMc.Spec.Config.Ignition.Version == "" {
```
jhrozek, line 346: Are we now relying on the MCs from the content to have the `Ignition.Version` set?

pkg/controller/complianceremediation/complianceremediation_controller_test.go
```diff
@@ -41,20 +42,24 @@ func getMockedRemediation(name string, labels map[string]string, applied bool, s
 		},
 	}
 
+	ign := igntypes.Config{
+		Ignition: igntypes.Ignition{
+			Version: igntypes.MaxVersion.String(),
```
jhrozek, line 47: Is this necessary since in the non-test code we don't set the version either? `make test-unit` seems to pass with the version removed.

JAORMX: Let me remove it.

JAORMX: Done

pkg/controller/complianceremediation/complianceremediation_controller.go
JAORMX: yes... we probably need to write unit tests in CaC to ensure this.

PR #301, opened 2020-06-16
Standard flags for all subcommands and common way of getting kubeconfig

JAORMX:
This enables standard flags for all subcommands (logging configurations
and kubeconfig related flags) which will enable us to configure similar
capabilities for all of them.

With this into use, this also enables a common way of getting the
cluster/client configuration. It can now dynamically check if it's in
cluster environment or if it needs to use a kubeconfig, thus enabling us
to use all of the subcommands locally if needed.

### DIFF
--- a/cmd/manager/aggregator.go
+++ b/cmd/manager/aggregator.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"flag"
 	"fmt"
 	"io"
 	"os"
@@ -27,14 +28,15 @@ import (
 
 	backoff "github.com/cenkalti/backoff/v3"
 	"github.com/dsnet/compress/bzip2"
+	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	"github.com/spf13/cobra"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	compv1alpha1 "github.com/openshift/compliance-operator/pkg/apis/compliance/v1alpha1"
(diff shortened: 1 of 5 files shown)


cmd/manager/aggregator.go
```diff
@@ -312,9 +321,9 @@ func aggregator(cmd *cobra.Command, args []string) {
 
 	aggregatorConf := parseAggregatorConfig(cmd)
 
-	config, err := rest.InClusterConfig()
+	config, err := config.GetConfig()
```
jhrozek, line 324: Nit: Can you rename either the variable name or add an import alias? Right now the variable name shadows the import name.

cmd/manager/api-resource-collector.go
```diff
@@ -54,60 +54,42 @@ type fetcherConfig struct {
...
-		if err != nil {
-			FATAL("Error building config: %v", err)
-		}
-		return conf
-	}
-
-	DBG("Running in-cluster")
-	conf, err := rest.InClusterConfig()
+func getConfig() *rest.Config {
+	config, err := config.GetConfig()
```
jhrozek, line 83: Same comment about config shadowing.

cmd/manager/resultcollector.go
```diff
@@ -409,9 +418,9 @@ func exitCodeIsError(exitcode string) bool {
 func resultCollectorMain(cmd *cobra.Command, args []string) {
 	scapresultsconf := parseConfig(cmd)
 
-	config, err := rest.InClusterConfig()
+	config, err := config.GetConfig()
```
jhrozek, line 421: Same thing about config shadowing.

cmd/manager/suitererunner.go
```diff
@@ -37,16 +39,23 @@ type rerunnerconfig struct {
...
+	flags.AddGoFlagSet(flag.CommandLine)
 }
 
 func getRerunnerConfig(cmd *cobra.Command) *rerunnerconfig {
 	var conf rerunnerconfig
 	conf.Name = getValidStringArg(cmd, "name")
 	conf.Namespace = getValidStringArg(cmd, "namespace")
 
-	config, err := rest.InClusterConfig()
+	config, err := config.GetConfig()
```
jhrozek, line 56: One more config shadowing.

JAORMX: Done

cmd/manager/resultcollector.go
JAORMX: Done

cmd/manager/api-resource-collector.go
JAORMX: Done

cmd/manager/aggregator.go
JAORMX: Done.

JAORMX:
> See the nits inline. But this is cool work, I wonder if this would allow us to write tests for the individual commands in isolation.

right! It could even allow us to run the subcommands outside of the operator.

PR #304, opened 2020-06-17
Label Profiles and Rules; and test

JAORMX:
In order to easily differentiate between what ProfileBundle owns what
profiles and rules, a label has been created.

This is tested in an e2e test that tests both that the profiles and
rules were correctly parsed (the data stream being correct) and that the
objects were labeled and we can filter them.

### DIFF
--- a/cmd/manager/profileparser.go
+++ b/cmd/manager/profileparser.go
@@ -225,6 +225,12 @@ func runProfileParser(cmd *cobra.Command, args []string) {
 	err = parseProfilesAndDo(contentDom, pcfg, func(p *cmpv1alpha1.Profile) error {
 		pCopy := p.DeepCopy()
 		profileName := pCopy.Name
+
+		if pCopy.Labels == nil {
+			pCopy.Labels = make(map[string]string)
+		}
+		pCopy.Labels[cmpv1alpha1.ProfileBundleOwnerLabel] = pb.Name
+
 		// overwrite name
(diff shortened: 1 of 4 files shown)


jhrozek:
This seems like a legitimate bug:
```
helpers.go:157: ProfileBundle.compliance.openshift.io "test-e2-e/-test-profile-parsing-works" is invalid: metadata.name: Invalid value: "test-e2-e/-test-profile-parsing-works": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

JAORMX:
> This seems like a legitimate bug:
> 
> ```
> helpers.go:157: ProfileBundle.compliance.openshift.io "test-e2-e/-test-profile-parsing-works" is invalid: metadata.name: Invalid value: "test-e2-e/-test-profile-parsing-works": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
> ```

right, I hadn't taken into account that tests have prefixes. So now it should be fixed.

PR #306, opened 2020-06-17
Only mark scan as COMPLIANT if there is at least one passing result

JAORMX:
This protects against cases where we can't for some reason detect the
platform we are running on.

In order to base the overall scan result on the individual rule results,
the patch also moves evaluating the scan result from the controller to
the aggregator. The aggregator now adds an annotation with the result
and an error instead. In case the scanner returned compliant but not a
single rule passed, the overall result is NotApplicable. In addition, an
event and a Suite error message are generated.


JAORMX:
This is basically https://github.com/openshift/compliance-operator/pull/279 but without the extra logging statements and rebased on top of master.

PR #315, opened 2020-06-22
Regenerate CRDs as v1 instead of v1beta1

JAORMX:
operator-sdk v0.18.X now generates CRDs using the v1 version instead of
v1beta1. This takes it into account.


JAORMX:
/lgtm cancel

Seems something is missing from this.

PR #319, opened 2020-06-22
Reflect PVC for the scan in the status

JAORMX:
This adds a "resultsStorage" output to the complianceScan (and suite)
which reflects where the raw results are stored.

This should make it easier for folks to find that.


pkg/apis/compliance/v1alpha1/compliancescan_types.go
```diff
@@ -150,6 +150,27 @@ type ComplianceScanStatus struct {
 	// Specifies the current index of the scan. Given multiple scans, this marks the
 	// amount that have been executed.
 	CurrentIndex int64 `json:"currentIndex,omitempty"`
+	// Specifies the object that's storing the raw resultsf the scan.
```
jhrozek, line 153: Fix this typo before merging please, otherwise looks good.

pkg/controller/compliancescan/rawresults.go
```diff
@@ -0,0 +1,80 @@
...
+// handles the necessary things to store and expose the raw results from the scan. This implies
+// that the PVC gets created, and, if necessary, the scan instance will get updated too.
+// Returns whether the reconcile loop should continue or not, and an error if encountered.
+func (r *ReconcileComplianceScan) handleRawResultsForScan(instance *compv1alpha1.ComplianceScan, logger logr.Logger) (bool, error) {
+	// Create PVC
+	pvc := getPVCForScan(instance)
+	logger.Info("Creating PVC for scan", "PersistentVolumeClaim.Name", pvc.Name, "PersistentVolumeClaim.Namespace", pvc.Namespace)
+	if err := r.client.Create(context.TODO(), pvc); err != nil && !errors.IsAlreadyExists(err) {
+		return false, err
+	}
```
jhrozek, line 24: btw do you think creating a PVC is something that should issue an event?

JAORMX: it already does. But not on the compliancescan. I could do that on a subsequent iteration.

PR #320, opened 2020-06-23
Add parallelism to e2e tests

JAORMX:
This makes the e2e tests run in parallel.

This has several implications

* We no longer run the tests in the openshift-compliance namespace.
  Instead, we rely on the namespace created by the operator-sdk test
  framework. This is because specifying the namespace made the framework
  disable parallelism

* We also no longer push the images to the `openshift-compliance`
  namespace. This is because we can no longer predict the namespace
  where the tests will run. Instead we push the images to the
  `openshift` namespace, where any pod can pull images from.

To be able to use the cluster role bindings on the appropriate
namespace, we edit the "namespaced manifest" created by the framework
before the test executes, since this is the only place where we know the
namespace the test will run in.


JAORMX:
Seems we do need https://github.com/openshift/compliance-operator/pull/321

PR #321, opened 2020-06-23
Fix race condition when dealing with missing tailoring config map

JAORMX:
Previously, we set the scan's result to error if the configmap wasn't
found, then requeued in and updated.

This was problematic since we would get a requeue of the scan without
error, and we would also have the errored scan in the reconcile loop.
This would cause a race condition where, at points, the scan would end
up in the appropriate result in some cases, and error out in others.

This removes the offending requeuing and instead relies on the reconcile
loop requeuing after the status update. Requeuing is only done if we had
previously encountered the issue and an update is not necessary.

On the other hand, this also cleans up the error status if we ended up
finding the tailoring configmap after previously not being able to.

### DIFF
--- a/pkg/controller/compliancescan/scan.go
+++ b/pkg/controller/compliancescan/scan.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"strings"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/controller/compliancescan/scan.go
```diff
@@ -525,9 +527,24 @@ func (r *ReconcileComplianceScan) reconcileReplicatedTailoringConfigMap(scan *co
 	origCM := &corev1.ConfigMap{}
 	origKey := types.NamespacedName{Name: origName, Namespace: origNs}
 	err := r.client.Get(context.TODO(), origKey, origCM)
+	// Tailoring ConfigMap not found
 	if err != nil && errors.IsNotFound(err) {
+		// We previously had dealt with this issue, just requeue
+		if strings.HasSuffix(scan.Status.ErrorMessage, tailoringNotFoundPrefix) {
```
jhrozek, line 533: Are you sure this should be a `HasSuffix` not a `HasPrefix` ?

JAORMX: you're right. Thanks for catching this!

JAORMX:
/hold

Haven't seen that race condition lately.

JAORMX:
/hold cancel

Nope, race condition still there.

PR #322, opened 2020-06-23
Make raw result storage size configurable

JAORMX:
The ComplianceScan stores the raw reslults in a PVC. As of now, the size
was hardcoded to 1Gi. This introduces a parameter in the scan to be able
to set it. The aforementioned parameter is called
`rawResultStorageSize`, and it defaults to 1Gi.

### DIFF
--- a/pkg/controller/compliancescan/compliancescan_controller_test.go
+++ b/pkg/controller/compliancescan/compliancescan_controller_test.go
@@ -42,7 +42,9 @@ var _ = Describe("Testing compliancescan controller phases", func() {
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "test",
 			},
-			Spec: compv1alpha1.ComplianceScanSpec{},
+			Spec: compv1alpha1.ComplianceScanSpec{
+				RawResultStorageSize: compv1alpha1.DefaultRawStorageSize,
+			},
 		}
 		objs = append(objs, compliancescaninstance)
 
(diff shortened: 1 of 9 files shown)


pkg/controller/compliancescan/compliancescan_controller_test.go
```diff
@@ -78,6 +80,52 @@ var _ = Describe("Testing compliancescan controller phases", func() {
...
+					Namespace: compliancescaninstance.Namespace,
+				}
+				reconciler.client.Get(context.TODO(), key, scan)
+				Expect(scan.Spec.RawResultStorageSize).To(Equal(compv1alpha1.DefaultRawStorageSize))
+			})
+		})
+
+		Context("With invalid RawResultStorageSize", func() {
+			It("report an error and move to phase DONE", func() {
+				compliancescaninstance.Spec.RawResultStorageSize = "invalid"
```
mrogers950, line 114: Would it be worth testing a scenario where you gave a valid size but one that the cluster can't accommodate due to other limits? i.e., 1000000Gi

JAORMX: That's a good point! Let me test it out locally and add a test case for that.

JAORMX: Should probably output an error too.

JAORMX: Done. Added an e2e test for that.

JAORMX:
/retest

installer error

PR #327, opened 2020-06-24
Re-introduce custom cleanup handler for e2e tests

JAORMX:
The cleanup done by operator-sdk was still cleaning resources even if it
shouldn't. Let's go back to using our own custom cleanup function.


JAORMX:
/hold

This doesn't seem to help.

PR #328, opened 2020-06-24
Debug flag no longer cleans up resources

JAORMX:
While at first glance it seemed like a good idea. It turns out that it's
useful to keep resources around to be able to debug and diagnose
deployments.

So, this changes the debug flag to no longer clean up pods and
resources. Instead it just enables debug logging of workloads and
OpenSCAP.

### DIFF
--- a/pkg/controller/compliancescan/compliancescan_controller.go
+++ b/pkg/controller/compliancescan/compliancescan_controller.go
@@ -35,6 +35,13 @@ var (
 )
 
 const (
+	// flag that indicates that deletion should be done
+	doDelete = true
+	// flag that indicates that no deletion should take place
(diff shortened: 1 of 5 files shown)


pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -32,6 +32,10 @@ var oneReplica int32 = 1
 var (
 	trueVal     = true
 	hostPathDir = corev1.HostPathDirectory
+	// flag that indicates that deletion should be done
+	doDelete = true
+	// flag that indicates that no deletion should take place
+	dontDelete = false
```
jhrozek, line 38: Can these be a const instead?

PR #329, opened 2020-06-24
Remove divergence in content in testcontent image

JAORMX:
The versioning issue was recently fixed, and so we no longer need this
hack. I intend to keep the testcontent image though, as it ensures that
content is available in CI without having to do multiple calls to quay
(just the initial image build)


jhrozek (approved):
/lgtm

PR #330, opened 2020-06-24
Add a ScanSettingBinding controller

jhrozek:
Introduces two new CRs: ScanSettingBinding and ScanSetting.
ScanSettingBinding is the one the controller watches as primary. It
contains references to Profiles or TailoredProfiles as well as a
reference to ScanSetting. Based on information in these CRs, the
controller generates a ComplianceSuite without the administrator having
to fill in any of the low-level SCAP data like content image or file.

The controller also watches the ScanSetting as a secondary resource and
maps updates to the Setting CRs to Reconcile events of any
ScanSettingBinding that uses this setting in order to allow updates.

Still TODO:
    - e2e tests
    - RBAC
    - not all errors are reconcilable
    - extend tests:
        - test updates
        - test updates to ScanSetting only
        - negative tests with mixup of platforms
    - when generating profiles, actually annotate the profiles with
      platform type and name
    - check bundles status for validity

Open questions:
    - if something goes wrong, is issuing an Event enough or should we
      add a status with an error message? So far I'm thinking an Event
      should be enough
        - OTOH, a Status could hold a link to the created Suite
    - currently Setting have Spec, maybe it should just have top-level
      attributes since it doesn't have a Status and would never have


pkg/apis/compliance/v1alpha1/profile_types.go
```diff
@@ -4,6 +4,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// PlatformTypeLabel specifies what kind of platform (node,platform)
+// this Profile or a TailoredProfile targets
+const PlatformTypeLabel = "compliance.openshift.io/platform-type"
```
JAORMX, line 9: nit: `platform-type: platform` sounds quite redundant; what about `scan-type`?

pkg/apis/compliance/v1alpha1/profile_types.go
```diff
@@ -4,6 +4,13 @@ import (
...
+// PlatformLabel specifies the name of the platform this Profile
+// or TailoredProfile is targetting. Example: ocp4, rhcos4, ...
+const PlatformLabel = "compliance.openshift.io/platform"
```
JAORMX, line 12: nit: What if we re-use OpenSCAP terminology here and use `product`? although, platform here is not bad... maybe `platform-id` instead?

pkg/apis/compliance/v1alpha1/scansetting_types.go
```diff
@@ -0,0 +1,44 @@
...
+type ScanSettingSpec struct {
+	// Defines whether or not the remediations should be applied automatically
+	AutoApplyRemediations bool `json:"autoApplyRemediations,omitempty"`
+	// Defines a schedule for the scans to run. This is in cronjob format.
+	// Note the scan will still be triggered immediately, and the scheduled
+	// scans will start running only after the initial results are ready.
+	Schedule string `json:"schedule,omitempty"`
+	// Disables cleaning up resources in the DONE phase, this might be useful for debugging.
+	Debug bool `json:"debug,omitempty"`
+	// The list of nodes to apply node-speicific checks to
```
JAORMX, line 17: s/nodes/roles/

pkg/controller/scansettingbinding/scansettingbinding_controller.go
```diff
@@ -0,0 +1,421 @@
...
+	// Watch for changes to primary resource ScanSettingBinding
+	err = c.Watch(&source.Kind{Type: &compliancev1alpha1.ScanSettingBinding{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to secordary resource ScanSetting. Since Setting does not link directly to a Binding,
+	// but the other way around, we use a mapper to enqueue requests for Binding(s) used by a Setting
+	err = c.Watch(&source.Kind{Type: &compliancev1alpha1.ScanSetting{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: &scanSettingMapper{mgr.GetClient()},
```
JAORMX, line 53: I didn't know you could do this! Nice!

pkg/controller/scansettingbinding/scansettingbinding_controller.go
```diff
@@ -0,0 +1,421 @@
...
+
+	v1alphaTp := compliancev1alpha1.TailoredProfile{}
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(tp.Object, &v1alphaTp)
+	if err != nil {
+		return err
+	}
+
+	scan.Profile = v1alphaTp.Status.ID
+	if v1alphaTp.Status.OutputRef.Name != "" {
+		// FIXME: OutputRef also has a namespace, but tailorringCofnigMapRef not?
```
JAORMX, line 295: Right, that's because the TailoringconfigMapRef that's used in the ComplianceScan is required to be in the same namespace as the scan. It was modeled similar to a pod, where the reference to the configmap for a pod is also without namespace since the namespace is assumed.

However, when outputting a reference, I wanted to be explicit and mention the namespace too since that would allow for doing extra checks on the object, and surface errors (did we output the configmap on the operator namespace or the watch namespace?).

Feel free to change it (add remove a namespace parameter) as you see fit.

pkg/apis/compliance/v1alpha1/profile_types.go
jhrozek: Oh, naming :-) Since the label is about Profile and not scans, what about `profile-type` or `profile-kind` or even more generic `target-type` ?

pkg/apis/compliance/v1alpha1/profile_types.go
jhrozek: I like product, because platform is already quite overloaded across CO.

pkg/apis/compliance/v1alpha1/profile_types.go
jhrozek: Given we would rename `PlatformLabel` below to `compliance.openshift.io/product`, what about `product-type` ?

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: I think so far it is OK, I just put the FIXME in the code to remember not to break any of the flows we have with RHACM.

pkg/apis/compliance/v1alpha1/scansetting_types.go
```diff
@@ -0,0 +1,44 @@
...
+	AutoApplyRemediations bool `json:"autoApplyRemediations,omitempty"`
+	// Defines a schedule for the scans to run. This is in cronjob format.
+	// Note the scan will still be triggered immediately, and the scheduled
+	// scans will start running only after the initial results are ready.
+	Schedule string `json:"schedule,omitempty"`
+	// Disables cleaning up resources in the DONE phase, this might be useful for debugging.
+	Debug bool `json:"debug,omitempty"`
+	// The list of roles to apply node-specific checks to
+	Roles []string `json:"roles,omitempty"`
+}
```
JAORMX, line 19: What about adding this to the ComplianceScan and replacing the values we have there? That way the settings will always be up-to-date

jhrozek: Sorry, I haven't seen this comment before I pushed the new version. I'll think about this tomorrow. In the meantime, the current version is no longer WIP and has a test (although I guess more e2e tests should be added...)

jhrozek:
No longer a WIP.

pkg/apis/compliance/v1alpha1/scansetting_types.go
JAORMX: At this point, with all the new parameters that have been added to the scan, I think it's necessary to embed this structure there and update the parameters

tests/e2e/helpers.go
```diff
@@ -1565,3 +1576,24 @@ func runPod(t *testing.T, f *framework.Framework, namespace string, podToRun *co
 func createAndRemoveEtcSecurettyOnNode(t *testing.T, f *framework.Framework, namespace, name, nodeName string) (*corev1.Pod, error) {
 	return runPod(t, f, namespace, createAndRemoveEtcSecurettyPod(namespace, name, nodeName))
 }
+
+func createReadyProfileBundle(t *testing.T, f *framework.Framework, name, namespace, contentImage, contentFile string) error {
```
JAORMX, line 1580: instead of this, what about already creating some profile bundles by default?

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -141,7 +141,7 @@ func (r *ReconcileComplianceSuite) Reconcile(request reconcile.Request) (reconci
 	return res, nil
 }
 
-func (r *ReconcileComplianceSuite) suiteDeleteHandler(suite *compv1alpha1.ComplianceSuite, logger logr.Logger) error {
+func (r *ReconcileComplianceSuite) suiteDeleteHandler(suite *compv1alpha1.ComplianceSuite) error {
```
JAORMX, line 144: instead of removing these parameters, I rather issue logs.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
```diff
@@ -0,0 +1,496 @@
...
+			return nil, fmt.Errorf("expected that tailored profile has 1 owner, got %d", len(o.GetOwnerReferences()))
+		}
+		owner := o.GetOwnerReferences()[0]
+
+		profileObj := unstructured.Unstructured{}
+		profileObj.SetAPIVersion(owner.APIVersion)
+		profileObj.SetKind(owner.Kind)
+
+		err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: instance.Namespace, Name: owner.Name}, &profileObj)
+		if err != nil {
```
JAORMX, line 460: If the Profile isn't found, we probably should fail and notify the user, shouldn't we?

pkg/controller/scansettingbinding/scansettingbinding_controller.go
```diff
@@ -0,0 +1,496 @@
...
+
+	// finally retrieve the bundle that the profile points to
+	if profReference.profile == nil {
+		return nil, fmt.Errorf("couldn't resolve profile")
+	}
+	if len(profReference.profile.GetOwnerReferences()) != 1 {
+		return nil, fmt.Errorf("expected that profile has 1 owner, got %d", len(profReference.profile.GetOwnerReferences()))
+	}
+
+	bundleReference := profReference.profile.GetOwnerReferences()[0]
```
JAORMX, line 478: shouldn't the kind already be properly set in the owner references? I think here you could just verify that it's the appropriate kind, and use a ProfileBundle object straight, as opposed to using an unstructured object.

On the other hand, you probably want to iterate the owner references here, just in case there is more than one.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
```diff
@@ -0,0 +1,496 @@
...
+
+	return scan, platform, nil
+}
+
+type profileReference struct {
+	name string
+
+	tailoredProfile *unstructured.Unstructured
+	profile         *unstructured.Unstructured
+	profileBundle   *unstructured.Unstructured
```
JAORMX, line 257: is there a reason why you're using unstructured objects here as opposed to the types themselves?

tests/e2e/helpers.go
jhrozek: You mean the test suite itself would create the bundles?

pkg/controller/compliancesuite/compliancesuite_controller.go
jhrozek: OK, this change doesn't belong to the patch anyway

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: By fail you mean a non-retriable error? It's a clear mistake I haven't added an Event, but looking at the codepath again, I wonder if we should issue one and requeue with a delay to allow the admin to create the link. I'm also OK with just failing with a non-retriable error, though.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: This is also related to the next question about using Unstructured objects. On one hand we use quite generic references in the CRDs to allow for different versions and even kinds to be used, on the other hand we need to "cast" them to typed objects at some point. So the idea was to have a function that just grabs the links regardless of what they are and then another separate one that uses the typed structures with the right version to construct the scan out of the references. If later we support more versions or different kids, we just branch the code that constructs the scan.

About the multiple owner references, what would the processing of other owners do except fail if there are multiple?

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: Hopefully this is answered above.

pkg/apis/compliance/v1alpha1/scansetting_types.go
jhrozek: I guess you're right, code duplication is bad.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
JAORMX: uhm... maybe retrying is the way to go. But yeah, an event is needed.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
JAORMX: I think other owners can be ignored. We really just care about a ProfileBundle being the owner here.

tests/e2e/helpers.go
JAORMX: I mean that, when the operator starts, we could create some bundles by default. Let me give it a try,

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: So, would a check for Kind being a ProfileBundle be OK? Was the explanation why I used the unstructured types sufficient or would you do the type-casting differently?

JAORMX: Just checking the kind is fine. And the explanation was sufficient.

pkg/apis/compliance/v1alpha1/scansetting_types.go
jhrozek: This is done is a separate commit as the existing controllers also had to be changed.

pkg/controller/compliancesuite/compliancesuite_controller.go
jhrozek: Removed

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: Done

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: Done, also for the other ownership (TP->P)

pkg/controller/scansettingbinding/scansettingbinding_controller.go
```diff
@@ -0,0 +1,515 @@
...
+	}
+
+	err = r.client.Create(context.TODO(), &suite)
+	if err == nil {
+		r.Eventf(
+			instance, corev1.EventTypeWarning, "SuiteCreated",
+			"ComplianceSuite %s/%s created", suite.Namespace, suite.Name,
+		)
+	} else if errors.IsAlreadyExists(err) {
+		err = r.client.Update(context.TODO(), &suite)
```
JAORMX, line 159: you can't do updates from an object that you instantiated as above (I've run into this several times :confused:  ). Updates have to be for objects fetched from the server. This is because the server will fill out the resource version, uid, etc. These are used to handle collisions in objects (optimistic concurrency control).

the logic would need to be something like this:

```
found := compliancev1alpha1.ComplianceSuite{}
err = r.client.Get(context.TODO(), suiteKey, found)
if err != nil && errors.IsNotFound(err) {
    // No object exists let's try to create it
    err = r.client.Create(context.TODO(), &suite)
    ...
// error handling
} else ... {
}

//  The object was found, so let's figure out if it needs an update
if suiteNeedsUpdate(suite) {
    // Copy the versioning and metadata references that we need
    suiteToBeUpdated := found.DeepCopy()
    // Copy the scans from the suite that you were updating and processing already
    // This way we make sure the update gathers all the info we need.
    suiteToBeUpdated.Spec.Scans := suite.Spec.Scans
    // Copy any other parameters that we need to copy
    // e.g.
    suiteToBeUpdated.Spec.ComplianceSuiteSettings := suite.Spec.ComplianceSuiteSettings
    suiteToBeUpdated.Spec.ComplianceScanSettings := suite.Spec.ComplianceScanSettings
    ....
}
```

JAORMX: There is a test for the compliance suite which updates the schedule. You could use that to test this out.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
```diff
@@ -0,0 +1,515 @@
...
+var _ reconcile.Reconciler = &ReconcileScanSettingBinding{}
+
+// ReconcileScanSettingBinding reconciles a ScanSettingBinding object
+type ReconcileScanSettingBinding struct {
+	client   client.Client
+	scheme   *runtime.Scheme
+	recorder record.EventRecorder
+}
+
+// FIXME: generalize for other controllers?
```
JAORMX, line 82: That would be a nice refactoring for another PR.

At this point, it would be nice to have a common Reconcile object that allows us to set and use common things such as the event recorder and logging with a simple function.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
JAORMX: The event is still missing from the case where the profile isn't resolved.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
```diff
@@ -0,0 +1,515 @@
...
+	o.SetAPIVersion(apiGroup)
+	o.SetKind(kind)
+
+	err := r.client.Get(context.TODO(), key, &o)
+	if errors.IsNotFound(err) {
+		return nil, common.NewRetriableCtrlErrorWithCustomHandler(func() (reconcile.Result, error) {
+			// This might be a temporary issue in the order the objects are being created
+			r.Eventf(
+				instance, corev1.EventTypeWarning, "NamedReferenceLookupError",
+				"NamedObjectReference %s not found", key,
```
JAORMX, line 468: We probably want to add the kind to the event too, to help people figure out which object was the issue.

tests/e2e/e2e_test.go
```diff
@@ -1152,10 +1198,103 @@ func TestE2E(t *testing.T) {
...
+				}
+
+				scanSettingBindingName := "generated-suite"
+				scanSettingBinding := compv1alpha1.ScanSettingBinding{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      scanSettingBindingName,
+						Namespace: namespace,
+					},
+					Profiles:   []compv1alpha1.NamedObjectReference{
+						// TODO: test also OCP profile when it works completely
```
JAORMX, line 1250: the OCP profile is not working?

tests/e2e/e2e_test.go
```diff
@@ -1152,10 +1198,103 @@ func TestE2E(t *testing.T) {
...
 			},
 		},
+		testExecution{
+			Name:       "TestScanSettingBinding",
+			IsParallel: true,
+			TestFn: func(t *testing.T, f *framework.Framework, ctx *framework.Context, mcTctx *mcTestCtx, namespace string) error {
+				objName := getObjNameFromTest(t)
+
+				rhcosPbName := objName + "-rhcos4"
+				err := createReadyProfileBundle(t, f, rhcosPbName, namespace, contentImagePath, rhcosContentFile)
```
JAORMX, line 1211: we already have some default profilebundles, would it be OK to use those? or do you think there's value in creating this temporary one?

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: It is a bit lower in the callstack now. `resolveProfile` -> `resolveTypedParent` -> `getUnstructured`. In `getUnstructured` we issue an Event in case the Get call returns NotFound. Do you think we should issue an even on all errors from Get?

pkg/apis/compliance/v1alpha1/scansettingbinding_types.go
```diff
@@ -0,0 +1,36 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ScanSettingBinding is the Schema for the scansettingbindings API
+// +kubebuilder:resource:path=scansettingbindings,scope=Namespaced
+type ScanSettingBinding struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Profiles   []NamedObjectReference `json:"profiles,omitempty"`
+	Constraint *NamedObjectReference  `json:"constraint,omitempty"`
```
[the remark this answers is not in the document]
jhrozek: Good idea, done

pkg/controller/scansettingbinding/scansettingbinding_controller.go
JAORMX: @jhrozek got it, I got a little lost there. It's good then.

tests/e2e/e2e_test.go
jhrozek: Wasn't there a case where you couldn't use the whole OCP profile because of some issue in the content? I forgot the details, but we were only selecting a single rule because of this...

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: done

tests/e2e/e2e_test.go
JAORMX: It was an issue with the expectations of the test code. Not the profile itself.

tests/e2e/e2e_test.go
jhrozek: I used them, btw I ran into an issue that unless we delete the bundles in the e2e test teardown, the test namespace stays stuck in terminating, so I added two defer statements to the helpers.go to address that, is that OK? There is still a helper function that waits until the bundles are ready on test setup.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: So far I added a new events.go module with a safer even that can or doesn't have to be initialized (unit tests can't initialize events it seems). If you prefer to also have a common reconcile object, I think it is out of scope of this PR.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
jhrozek: Updates now work, tested manually so far. I will add a test that makes sure that a Suite can be updated, but the suite controller so far doesn't update scans, so we need to fix that, too

deploy/crds/compliance.openshift.io_v1alpha1_scansetting_cr.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: compliance.openshift.io/v1alpha1
+kind: ScanSetting
+metadata:
+  name: example-scansetting
+spec:
+  # Add fields here
+  size: 3
```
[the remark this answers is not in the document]
jhrozek: Done

deploy/crds/compliance.openshift.io_v1alpha1_scansettingbinding_cr.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: compliance.openshift.io/v1alpha1
+kind: ScanSettingBinding
+metadata:
+  name: example-scansettingbinding
+spec:
+  # Add fields here
+  size: 3
```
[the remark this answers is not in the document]
jhrozek: Done

pkg/controller/scansettingbinding/scansettingbinding_controller.go
JAORMX: yep, that's why I said that it would be for another PR.

pkg/controller/scansettingbinding/scansettingbinding_controller.go
JAORMX: I guess you still need to remove this block, cause the same logic (with the corrections) is now in line 168.

PR #331, opened 2020-06-25
Migrate scap_test to ginkgo and re-introduce gingko.noColor flag

JAORMX:
* Migrating that test ensures that all our unit tests are in sync using the
  same framework

* Re-introducing that flag ensures that test failures are readable in CI.

### DIFF
--- a/cmd/manager/scap_test.go
+++ b/cmd/manager/scap_test.go
@@ -49,8 +49,8 @@ var _ = Describe("Testing SCAP parsing and storage", func() {
 		})
 	})
 
-	Context("Storing files in their appropriate directories", func() {
-		It("Stores the file in the right place with the root being '/tmp'", func() {
+	Context("Parses the save path appropriately", func() {
+		It("Parses correctly with the root being '/tmp'", func() {
 			root := "/tmp"
 			path := "/apis/foo"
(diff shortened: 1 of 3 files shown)


cmd/manager/scap_test.go
```diff
@@ -2,105 +2,88 @@ package main
...
+			expectedAPIPath := "/apis/config.openshift.io/v1/oauths/cluster"
+
+			warning := `<warning category="general" lang="en-US"><code class="ocp-api-endpoint">/apis/config.openshift.io/v1/oauths/cluster</code><code class="ocp-dump-location"><sub idref="xccdf_org.ssgproject.content_value_ocp_data_root" use="legacy" />/idp.yml</code>file.</warning>`
+			api := getPathFromWarningXML(warning)
+			Expect(api).To(Equal(expectedAPIPath))
+		})
+	})
+
+	Context("Storing files in their appropriate directories", func() {
+		It("Stores the file in the right place with the root being '/tmp'", func() {
```
mrogers950, line 53: The function doesn't do any file storing, just handling the save path, so a more accurate description would be something like "Parses the save path properly.."

JAORMX: Done

PR #332, opened 2020-06-26
Enable raw result rotation

JAORMX:
This adds a flag called `rawResultStorageRotation` which, allows users to
configure the rotation for the result directories in the PVC.

It defaults to '3' and will skip the `lost+found` directory if it's
present.

It works by checking the creation date for the directories and making
the rotation decision based on that. This way we don't have to think
about the indexing names and rotation of indexes.

A rotation policy of 0 disables the rotation entirely.


JAORMX:
/hold

oops! found a bug. Fixing.

JAORMX:
/hold cancel

fixed.

PR #337, opened 2020-06-29
resultscollector: Only base64 results for ConfigMaps

### LINKED ISSUE #295: Please change directions to say  oc exec pods/pv-extract ls /workers-scan-results/0 instead of  oc exec pods/pv-extract ls /workers-scan-results/
Please change directions to say
```
oc exec pods/pv-extract ls /workers-scan-results/0 
```
instead of  
```
oc exec pods/pv-extract ls /workers-scan-results/
```
I thought the results were missing for a moment...
Thanks!

JAORMX:
This sends the results to the resultserver as bzip2 files instead of base64
with bzip2 in it. This results in slightly ness network usage due to the
reduced size of the output.

Only results as they're stored in ConfigMaps are stored with base64 wrapping.

On the other hand, this changes the logging from the resultcollector to use the
common log object.

Closes #295

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,9 +396,9 @@ $ oc get nodes
 Once the nodes reboot, you might want to run another Suite to ensure that
 the remediation that you applied previously was no longer found.
 
-## Extracting results
+## Extracting raw results
 
-The scans provide two kinds of results: the full report in the ARF format
+The scans provide two kinds of raw results: the full report in the ARF format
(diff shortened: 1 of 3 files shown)


JAORMX:
https://github.com/openshift/compliance-operator/pull/337/commits/89e80cfaa26f3d01b8b56c168f0a5d64a55e1977 addressed #295

Issue #293, opened 2020-06-13
ComplainceSuite with taints/tolerations


akilada:
Is it possible to add support when there are taints configured for a node?

The scan pods couldn't get deployed to worker nodes when the nodes had taints.

```
apiVersion: compliance.openshift.io/v1alpha1
kind: ComplianceSuite
metadata:
  name: worker-coreos
spec:
  autoApplyRemediations: false
  schedule: "0 6 * * *"
  scans:
    - name: worker-node-scan
      profile: xccdf_org.ssgproject.content_profile_moderate
      content: ssg-rhcos4-ds.xml
      contentImage: quay.io/complianceascode/ocp4:latest
      nodeSelector:
        node-role.kubernetes.io/worker: ""
```

Taints,

```
  taints:
    - key: clusterx
      value: 'true'
      effect: NoSchedule
```

JAORMX:
It's not something we can set at the moment, but that's definitely something we should support. Thanks for bringing it up!

PR #338, opened 2020-06-29
Enable setting tolerations for scans

JAORMX:
This adds a `scanTolerations` field to the complianceScan. This, as
expected, sets tolerationsf or the scan workloads. Thus enabling users
to set this up for nodes with custom taints.

Closes: https://github.com/openshift/compliance-operator/issues/293


JAORMX:
/hold

Missing e2e tests.

PR #344, opened 2020-06-29
Delete the ResultsServer deployment prior to mounting the results checker pod

jhrozek:
I've been seeing errors like:
        --- FAIL: TestE2E/Parallel_tests/TestScheduledSuite (1391.14s)
...
        helpers.go:63: 2020-06-29T16:57:15Z: Waiting until the raw result checker workload is done: osdk-e2e-19bfc00b-d819-4d71-af7b-d58c6b6c7f40/rotation-checker
...
          helpers.go:63: 2020-06-29T17:12:16Z: Pod not done. retrying: %!s(<nil>)
          helpers.go:63: 2020-06-29T17:12:16Z: Pod not done. retrying: %!s(<nil>)
          helpers.go:182: timed out waiting for the condition

In the e2e logs. Looking into the environment, there seems to be a problem
mounting the PVC:
    Warning  FailedAttachVolume  97s        attachdetach-controller  Multi-Attach error for volume "pvc-182244b9-c2a3-40e9-aa4d-e3d1e8d34b4e" Volume is already used by pod(s) test-scheduled-suite-workers-scan-rs-7977bcb598-9kml2

Terminating the Deployment should work around the issue. An alternative
might be to schedule the pod on the same node as the RS pod.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	goctx "context"
 	"fmt"
+	appsv1 "k8s.io/api/apps/v1"
 	"math/rand"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


JAORMX:
huh... the checker is mounting the pvc in read-only mode. It should have been fine :/ that's weird.

jhrozek:
It also only started happening today, might not be an error on our end.

PR #346, opened 2020-06-29
Remove unused parameter

jhrozek:
Does what it says on the box.


JAORMX:
would probably be good to actually use the parameter and log though

PR #347, opened 2020-06-30
Migrate aggregator logs to using zap

JAORMX:
As the other cmdline components, this moves the aggregator to use the
zap logs.


JAORMX:
Pull-request updated, HEAD is now 72f3cff47109cfcab5f2411aab359d02a704a32a

PR #352, opened 2020-07-01
e2e: Create objects with default cleanup options

JAORMX:
This ensures that we cleanup after e2e tests so we can run them easily
subsequently.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -39,13 +39,13 @@ func TestE2E(t *testing.T) {
 						ContentFile:  rhcosContentFile,
 					},
 				}
-				if err := f.Client.Create(goctx.TODO(), testPB, nil); err != nil {
+				if err := f.Client.Create(goctx.TODO(), testPB, getCleanupOpts(ctx)); err != nil {
 					return err
 				}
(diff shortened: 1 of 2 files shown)


jhrozek:
This looks good, but there is a bunch of calls that use a custom cleanup option structure that is in fact the same as the default cleanup options, e.g.:
```
				err := f.Client.Create(goctx.TODO(), testSuite, &framework.CleanupOptions{TestContext: ctx, Timeout: cleanupTimeout, RetryInterval: cleanupRetryInterval})
```
Do you think it makes sense to convert them in one go, too?

JAORMX:
> This looks good, but there is a bunch of calls that use a custom cleanup option structure that is in fact the same as the default cleanup options, e.g.:
> 
> ```
> 				err := f.Client.Create(goctx.TODO(), testSuite, &framework.CleanupOptions{TestContext: ctx, Timeout: cleanupTimeout, RetryInterval: cleanupRetryInterval})
> ```
> 
> Do you think it makes sense to convert them in one go, too?

it does. I'll do that.

jhrozek:
btw the cleanup itself seems to work well, I ran a couple of tests and never ran into the issue of the NS being stuck in terminating state again.

JAORMX:
@jhrozek added the missing cleanups

PR #353, opened 2020-07-01
Parse out profile name and type

jhrozek:
Parses the CPE string into components and parses out the product type, 
either node or platform and the product name. For the product type, the
"part" piece of the CPE URI which can either be "a" for "application" or
"o" for operating system.


jhrozek:
```
   * could not run steps: step src failed: could not wait for build: the build src failed after 58m51s with reason PullBuilderImageFailed: Failed pulling builder image. 
```
/retest

PR #355, opened 2020-07-01
Handle ProfileBundle content image updates

JAORMX:
This changes the profile bundle controller to detect if a new image has
been set, if so, it'll attempt to update a Deployment which actually
creates the profileparser.

This way we could address updates in the long run, by asking folks to
point to a specific hash and subsequent updates would merely update the
image's digest for the bundles.


JAORMX:
@jhrozek @mrogers950 PTAL

PR #356, opened 2020-07-03
Encapsulate RawResultStorage options into its own section

JAORMX:
This introduces a section to the scan called `RawResultStorage`, this
sets storage related options through the RawResultStorageSettings
object.

As we add more storage-related options to these objects, this makes sure
that we don't use that many redundant names (e.g. RawResultStorage as a
prefix for all of them) and instead have them all in one logical place.


tests/e2e/e2e_test.go
```diff
@@ -920,13 +926,15 @@ func TestE2E(t *testing.T) {
...
-									Rule:                     "xccdf_org.ssgproject.content_rule_no_netrc_files",
-									NodeSelector:             selectWorkers,
+									ContentImage: contentImagePath,
+									Profile:      "xccdf_org.ssgproject.content_profile_moderate",
+									Content:      rhcosContentFile,
+									Rule:         "xccdf_org.ssgproject.content_rule_no_netrc_files",
+									NodeSelector: selectWorkers,
 									ComplianceScanSettings: compv1alpha1.ComplianceScanSettings{
-										RawResultStorageRotation: 1,
+										RawResultStorage: compv1alpha1.RawResltStorageSettings{
```
jhrozek, line 935: This typo prevents the tests from compiling.

JAORMX: ugh... I had fixed it localy and forgot to push it.

PR #365, opened 2020-07-15
Update the README for recent changes to the operator

jhrozek:
This updates the README.md with descriptions of the objects we added 
recently like Profile, TailoredProfile, ScanSetting and ScanSettingBinding.

Also includes another minor update to document the PVCs instead of PVs
which is more user-friendly.


mrogers950:
Great work!
/lgtm

PR #373, opened 2020-07-24
profileparser: Update data parsed out from a ProfileBundle

jhrozek:
In order to provide the most up-to-date view of data parsed from a 
ProfileBundle, the PB controller already re-runs a pod each time a new 
image is detected.

In addition, this patch updates objects parsed from the bundle such as 
Profiles, Rules and Variables. After the existing objects have been 
updated, all objects that are not parsed from the current bundle are 
removed. The detection is implemented by annotating the objects with the 
current image digest, so the cleanup removes all objects that are not 
annotated with that digest.

The code that parses the whole bundle has been moved from the manager to 
the profileparser package so that it is easier to test. Most of the tests
are unit tests, the e2e tests are a subset of the unit tests.

Jira: CMP-665


jhrozek:
This is a large PR; I'm going to highlight some things I dislike or would like to get some help with ...

pkg/profileparser/profileparser.go
```diff
@@ -56,6 +60,179 @@ func GetPrefixedName(pbName, objName string) string {
 	return pbName + "-" + objName
 }
 
+func ParseBundle(contentDom *xmldom.Document, pb *cmpv1alpha1.ProfileBundle, pcfg *ParserConfig) error {
+	err := ParseProfilesAndDo(contentDom, pb, func(p *cmpv1alpha1.Profile) error {
+		err := parseAction(p, "Profile", pb, pcfg, func(found, updated interface{}) error {
```
jhrozek, line 65: Honestly I dislike that the block that calls `parseAction` is a callback for the per-type parsers which itself has an update callback. This is all done so that the per-type payload can be updated. I tried working around this by adding UpdatePayload methods to the three types and passing an interface instead, but that seems a bit clunky as well...

pkg/profileparser/profileparser.go
```diff
@@ -56,6 +60,179 @@ func GetPrefixedName(pbName, objName string) string {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func deleteIfNotCurrentDigest(client runtimeclient.Client, imageDigest string, item *unstructured.Unstructured) error {
+	itemDigest := item.GetAnnotations()[cmpv1alpha1.ProfileImageDigestAnnotation]
```
jhrozek, line 227: I put the digest into annotations because it can be very long, longer than the limit for labels. This means that we have to iterate over all objects of a type and read their annotations one by one. I thought this was OK, because profile update is a relatively rare operation. The alternative might be to put a shortened version of the digest into the label. e.g. git or quay.io use just first N characters of a digest, but at the same time I think they must have some heuristics to make sure the shortened version is unique. In the end, this is safer, but slower...

pkg/apis/compliance/v1alpha1/variable_test.go
```diff
@@ -37,25 +39,27 @@ var _ = Describe("Testing variables API", func() {
...
+							"vocals",
+							"john",
+						},
+						{
+							"bass",
+							"paul",
+						},
+						{
+							"drums",
+							"ringo",
```
mrogers950, line 57: Good ol' Ringo

pkg/profileparser/profileparser.go
mrogers950: This part is fine with me.

tests/e2e/e2e_test.go
```diff
@@ -23,21 +23,31 @@ import (
 func TestE2E(t *testing.T) {
 	executeTests(t,
 		testExecution{
-			Name:       "TestProfileParsingWorks",
+			Name:       "TestProfileModification",
 			IsParallel: true,
 			TestFn: func(t *testing.T, f *framework.Framework, ctx *framework.Context, mcTctx *mcTestCtx, namespace string) error {
+				const (
+					baselineImage = "quay.io/jhrozek/ocp4-openscap-content@sha256:a1709f5150b17a9560a5732fe48a89f07bffc72c0832aa8c49ee5504510ae687"
+					modifiedImage = "quay.io/jhrozek/ocp4-openscap-content@sha256:7999243c0b005792bd58c6f5e1776ca88cf20adac1519c00ef08b18e77188db7"
```
mrogers950, line 31: Just wondering if the tag parsing will get the digest properly from a regular tag like `:latest` etc.

mrogers950:
Looks pretty good to me, just one comment about the digests.

pkg/profileparser/profileparser.go
mrogers950: Agree that this is not frequent enough to worry about optimizing. Might be good to leave a comment about it.

pkg/profileparser/profileparser.go
jhrozek: Thanks, marking as resolved, then

pkg/apis/compliance/v1alpha1/variable_test.go
jhrozek: ba dum tss...

tests/e2e/e2e_test.go
jhrozek: No, sorry. First, I wasn't sure if this was safe, because the profileparser runs on one node, but the scans run on different. And because the content image is pulled when it's being used by the scans, using the digest is the only way to make sure everything points to the same image (the content provider should in theory be nice and point one versioned tag to one digest and never change it but...). So by using the tag, the profileparser might parse something, then the digest changes because someone pushes a new version of the tag and then the scan uses a different content.. And even on the code side, I wasn't sure how to get the sha hash short of having a cron job that pulls the image periodically and checks the sha itself.

That said, maybe relying on the sha being there is a bit of overkill. Maybe we could instead just use the whole `imageName` regardless of what's there as an annotation? That way we could at least support changing the imageName from e.g. `v0.1` to `v0.2` or such..this still doesn't get away with the race between defining the profilebundle by tag and using the actual content, but maybe it's easier for the admin. I actually start leaning that way. Let me know what you prefer, changing the PR is not a big deal.

pkg/profileparser/profileparser.go
jhrozek: Done.

jhrozek:
The latest version just adds a comment about the annotations being potentially slow.

tests/e2e/e2e_test.go
mrogers950: This is OK for now. if we want to support a regular tag we might be able to mirror images to the internal registry and compare/detect sha changes through the API.

PR #374, opened 2020-07-29
Update MC remediations

jhrozek:
If contents of remediation that was previously aplied is changed, the
remediation is now switched to a new state called "Outdated". The
previous contents are stored in a separate attribute and the admin is
supposed to choose between the outdated or the new content, IOW the
admin would accept the new remediation contents by removing the outdated
fields.

The outdated remediations are also labeled and a suite event is issued when
any such remediations exist.


cmd/manager/aggregator.go
```diff
@@ -382,14 +388,64 @@ func createResults(crClient *complianceCrClient, scan *compv1alpha1.ComplianceSc
...
+				// the current state at their own pace
+				foundRemediation.Spec.Current.DeepCopyInto(&pr.Remediation.Spec.Outdated)
+				// The application state must be preserved or else we'd un-apply the remediation
+				// once it's updated
+				pr.Remediation.Spec.Apply = foundRemediation.Spec.Apply
+			}
+
 			// Copy resource version and other metadata needed for update
 			foundRemediation.ObjectMeta.DeepCopyInto(&pr.Remediation.ObjectMeta)
+			// Copy the status; in case we are updating an already applied remediation
```
mrogers950, line 413: Might be some missing code/TODO related to this comment?

JAORMX:
@jhrozek could you add more details on why the current change is being done this way? Would be nice to have more context on why the current/outdated fields were added. It seems to me like this could have been done without those fields, but maybe I'm missing context.

jhrozek:
On Wed, Jul 29, 2020 at 09:48:06PM -0700, Juan Osorio Robles wrote:
> @jhrozek could you add more details on why the current change is being done this way? Would be nice to have more context on why the current/outdated fields were added. It seems to me like this could have been done without those fields, but maybe I'm missing context. 

Because otherwise there is no way for the admin to review the updated
remediations and the updated remediations would be applied as soon as
the scan re-runs.

Or in more detail: the remediations are bound to a scan/suite, so the
only way remediation updates make sense is when the suite is running on
a schedule, or is restarted with an annotation. For a one-off scan, that
you delete and run another, the updates don't make sense.

So with a scheduled scan, you have the aggregator that takes the
parsed-out checks and remediations Go objects and updates the check and
remediations CRs. The complianceremediation controller listens to any
changes to remediation CRs, so if you just update the remediation (and
don't store the obsolete status), that would trigger re-rendering the
composite MC with the updated remediations at the time the remediation
CRs are updated. I thought the admins wouldn't like their cluster
rebooting after re-running a scan with no chance of them reviewing the
updates.

The only other way I can think of is, if any Outdated remediations
exist, never re-render the composite MachineConfig object and let the
admin either accept the new remediation or not. This would have been
simpler, but having the previous version of the remediation payload
allows you to 1) see exactly what changed and 2) decide you want to
actually keep the old version for the moment and 3) re-render composite
MC by changing the other remediations.

If you think any of that is not needed and we could get away by asking
the admin to either accept the new remediation or not accept it as a
whole, we can ditch the old contents. But I would really like to keep
the obsolete state and force the admin to make a choice of accepting the
new remediations and reboot the cluster at their own pace.

JAORMX:
@jhrozek thanks for the explanation. The idea makes sense to me. Lgtm from my side.

jhrozek:
I added a test that works for me locally, let's see what CI has to say...

cmd/manager/aggregator.go
jhrozek: Copy-paste error.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -300,12 +317,24 @@ func getAppliedMcRemediations(r *ReconcileComplianceRemediation, rem *compv1alph
...
 	for i := range scanSuiteRemediations.Items {
-		if !utils.IsMachineConfig(scanSuiteRemediations.Items[i].Spec.Object) {
+		if !utils.IsMachineConfig(scanSuiteRemediations.Items[i].Spec.Current.Object) {
 			continue
 		}
+
+		var mcObj *unstructured.Unstructured
+
+		// We'll only merge the one that is being reconciled with those that are already
+		// applied. For remediations that are obsolete, we're merge
```
mrogers950, line 327: Incomplete comment

jhrozek: Fixed locally, I'll push the changes all at once to not trigger CI over and over again.

jhrozek:
I think the previous error was not related. This is the high-level log:
```
             helpers.go:70: 2020-07-30T21:46:10Z: Waiting until suite test-scheduled-suite reaches target status 'DONE'. Current status: AGGREGATING
            helpers.go:203: expecting COMPLIANT got ERROR 
```
which correlates to:
```
{"level":"info","ts":1596145572.1514523,"logger":"scanctrl","msg":"Reconciling ComplianceScan","Request.Namespace":"osdk-e2e-4bbcaa80-96bd-4bff-8dd1-b862c355d3be","Request.Name":"test-scheduled-suite-workers-scan"}
{"level":"info","ts":1596145572.1517777,"logger":"scanctrl","msg":"Phase: Aggregating","Request.Namespace":"osdk-e2e-4bbcaa80-96bd-4bff-8dd1-b862c355d3be","Request.Name":"test-scheduled-suite-workers-scan"}
{"level":"info","ts":1596145572.1552417,"logger":"scanctrl","msg":"Creating an aggregator pod for scan","Request.Namespace":"osdk-e2e-4bbcaa80-96bd-4bff-8dd1-b862c355d3be","Request.Name":"test-scheduled-suite-workers-scan"}
{"level":"info","ts":1596145572.1553884,"logger":"scanctrl","msg":"Content image","Request.Namespace":"osdk-e2e-4bbcaa80-96bd-4bff-8dd1-b862c355d3be","Request.Name":"test-scheduled-suite-workers-scan","image":"registry.svc.ci.openshift.org/ci-op-mm0nkswx/stable:testcontent"}
{"level":"info","ts":1596145572.1801846,"logger":"scanctrl","msg":"Checking aggregator pod for scan","Request.Namespace":"osdk-e2e-4bbcaa80-96bd-4bff-8dd1-b862c355d3be","Request.Name":"test-scheduled-suite-workers-scan","ComplianceScan.Name":"test-scheduled-suite-workers-scan"}
{"level":"info","ts":1596145572.1805148,"logger":"scanctrl","msg":"Pod has finished","Request.Namespace":"osdk-e2e-4bbcaa80-96bd-4bff-8dd1-b862c355d3be","Request.Name":"test-scheduled-suite-workers-scan"}
{"level":"info","ts":1596145572.1806245,"logger":"scanctrl","msg":"Moving on to the Done phase"}
{"level":"info","ts":1596145572.180798,"logger":"scanctrl","msg":"Node scan error","node.Name":"ip-10-0-224-175.us-west-1.compute.internal","errMsg":"the ConfigMap 'openscap-pod-656e90145a1104869203f4d41de4968acc5cc4f3' annotation was missing the result key"}
```
Let's see if it happens again...

jhrozek:
One know-to-be-flaky test failed, trying again

jhrozek:
Huh, again this issue?!?
```
{"level":"info","ts":1596145572.180798,"logger":"scanctrl","msg":"Node scan error","node.Name":"ip-10-0-224-175.us-west-1.compute.internal","errMsg":"the ConfigMap 'openscap-pod-656e90145a1104869203f4d41de4968acc5cc4f3' annotation was missing the result key"}
```
trying locally..

jhrozek:
I really don't know what's going on. The log complaining that:
```
{"level":"info","ts":1596184628.955853,"logger":"scanctrl","msg":"Node scan error","node.Name":"ip-10-0-189-238.ec2.internal","errMsg":"the ConfigMap 'test-filtered-scan-ip-10-0-189-238.ec2.internal-pod' annotation was missing the result key"}
```
but that config map looks like this:
```
   "metadata": {
                "annotations": {
                    "compliance-remediations/processed": "",
                    "compliance.openshift.io/scan-error-msg": "",
                    "compliance.openshift.io/scan-result": "COMPLIANT",
                    "openscap-scan-result/node": "ip-10-0-189-238.ec2.internal"
                },
                "creationTimestamp": "2020-07-31T08:35:52Z",
                "labels": {
                    "compliance-scan": "test-filtered-scan"
                },
                "managedFields": [
                    {
                        "apiVersion": "v1",
                        "fieldsType": "FieldsV1",
                        "fieldsV1": {
                            "f:data": {
                                ".": {},
                                "f:exit-code": {},
                                "f:results": {}
                            },
                            "f:metadata": {
                                "f:annotations": {
                                    ".": {},
                                    "f:compliance-remediations/processed": {},
                                    "f:compliance.openshift.io/scan-error-msg": {},
                                    "f:compliance.openshift.io/scan-result": {},
                                    "f:openscap-scan-result/node": {}
                                },
                                "f:labels": {
                                    ".": {},
                                    "f:compliance-scan": {}
                                }
                            }
                        },
                        "manager": "compliance-operator",
                        "operation": "Update",
                        "time": "2020-07-31T08:37:05Z"
                    }
                ],
                "name": "test-filtered-scan-ip-10-0-189-238.ec2.internal-pod",
```
the key the code is looking for is compliance.openshift.io/scan-result ...

jhrozek:
and I guess it's not a race between updating the map and checking it because the CM says it was updated `2020-07-31T08:37:05Z"` but the log message is from ` 08:37:08 AM UTC`, so 3 seconds later...

PR #378, opened 2020-08-03
Upgrade to operator-sdk v0.18.2

JAORMX:


vendor/github.com/coreos/ignition/config/v2_2/types/schema.go
```diff
@@ -45,8 +45,8 @@ type Disk struct {
 }
 
 type File struct {
-	Node          `json:",inline"`
-	FileEmbedded1 `json:",inline"`
+	Node
```
jhrozek, line 48: I think we still need to keep these

Makefile
```diff
@@ -48,7 +48,7 @@ export OPERATOR_NAMESPACE?=openshift-compliance
 
 # Operator-sdk variables
 # ======================
-SDK_VERSION?=v0.18.1
+SDK_VERSION?=v0.18.2
```
jhrozek, line 51: This is OK with me as long as we stay on the 0.18 branch for now. It might be better to skip upgrading for a while unless there is a reason to upgrade and then go directly to 1.0 and figure out the bundle/manifest format along the way.

vendor/github.com/coreos/ignition/config/v2_2/types/schema.go
JAORMX: oops! you're right

PR #384, opened 2020-08-04
Detect if nodeSelector doesn't match

JAORMX:
For node scan types, if the node selector doesn't match, the result
won't be valid. So let's detect this and so something about it:

* We issue an event that indicates that the nodeSelector didn't match nodes.
* We move the state directly to "Done" and the result to "NotApplicable"

piggy-backed to this, was some refactoring to the handling of scan types.

This keeps the functionality that allows case-insensitive usage of scan types
but does validation so only allowed scan types can be used. This will result in
an event and an appropriate status on the CR.

Jira: OCPBUGSM-1430

### DIFF
--- a/pkg/controller/compliancescan/compliancescan_controller.go
+++ b/pkg/controller/compliancescan/compliancescan_controller.go
@@ -165,6 +165,26 @@ func (r *ReconcileComplianceScan) phasePendingHandler(instance *compv1alpha1.Com
 		return reconcile.Result{}, err
 	}
 
+	// Set default scan type if missing
+	if instance.Spec.ScanType == "" {
+		instanceCopy := instance.DeepCopy()
+		instanceCopy.Spec.ScanType = compv1alpha1.ScanTypeNode
+		err := r.client.Update(context.TODO(), instanceCopy)
+		return reconcile.Result{}, err
+	}
+
+	// validate scan type
+	if _, err := instance.GetScanTypeIfValid(); err != nil {
+		r.recorder.Event(instance, corev1.EventTypeWarning, "InvalidScanType",
+			"The scan type was invalid")
+		instanceCopy := instance.DeepCopy()
+		instanceCopy.Status.Result = compv1alpha1.ResultError
+		instanceCopy.Status.ErrorMessage = fmt.Sprintf("Scan type '%s' is not valid", instance.Spec.ScanType)
+		instanceCopy.Status.Phase = compv1alpha1.PhaseDone
+		updateErr := r.client.Status().Update(context.TODO(), instanceCopy)
+		return reconcile.Result{}, updateErr
+	}
+
 	// Set default storage if missing
 	if instance.Spec.RawResultStorage.Size == "" {
 		instanceCopy := instance.DeepCopy()
@@ -183,7 +203,7 @@ func (r *ReconcileComplianceScan) phasePendingHandler(instance *compv1alpha1.Com
 		return reconcile.Result{}, err
 	}
 
-	if isNodeScan(instance) {
+	if instance.GetScanType() == compv1alpha1.ScanTypeNode {
 		var nodes corev1.NodeList
 		var err error
 		if nodes, err = getTargetNodes(r, instance); err != nil {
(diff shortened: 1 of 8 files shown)


pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -813,5 +830,9 @@ func getInitContainerImage(scanSpec *compv1alpha1.ComplianceScanSpec, logger log
 }
 
 func isPlatformScan(instance *compv1alpha1.ComplianceScan) bool {
-	return strings.ToLower(string(instance.Spec.ScanType)) == "platform"
+	return instance.Spec.ScanType == compv1alpha1.ScanTypePlatform
```
jhrozek, line 833: While the code reads nicer after the change, removing the tolower(string).. construct would regress the fix done in c6ebe30eabac8c36aee67eb4510cdb8b9cb6b0b3

tests/e2e/e2e_test.go
```diff
@@ -422,6 +422,41 @@ func TestE2E(t *testing.T) {
 				return scanResultIsExpected(t, f, namespace, "test-filtered-scan", compv1alpha1.ResultCompliant)
 			},
 		},
+		testExecution{
+			Name:       "TestScanWithNodeSelectorNoMatches",
+			IsParallel: true,
+			TestFn: func(t *testing.T, f *framework.Framework, ctx *framework.Context, mcTctx *mcTestCtx, namespace string) error {
+				scanName := getObjNameFromTest(t)
+				selectWorkers := map[string]string{
```
jhrozek, line 430: Please rename the variable from `selectWorkers` to something like `selectNone` or similar.

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -573,8 +573,8 @@ func (r *ReconcileComplianceScan) generateResultEventForScan(scan *compv1alpha1.
 
 	if scan.Status.Result == compv1alpha1.ResultNotApplicable {
 		r.recorder.Eventf(
-			scan, corev1.EventTypeNormal, "ScanNotApplicable",
-			"The scan result is not applicable, please check if you're using the correct platform")
+			scan, corev1.EventTypeWarning, "ScanNotApplicable",
+			"The scan result is not applicable, please check if you're using the correct platform or if the nodeSelector matches nodes.")
```
jhrozek, line 577: Should we suggest nodes or pool?

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: why do we need to allow lowercase?

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: I would add `nodes` here given that it's canonically what should happen... the fact that the nodeSelector should ideally matcha pool should still be indicated as part of the documentation. But it's not strictly necessary, it just won't generate remediations.

pkg/controller/compliancescan/compliancescan_controller.go
JAORMX: answerd via slack

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -313,7 +313,8 @@ func (r *ReconcileComplianceScan) phaseRunningHandler(instance *compv1alpha1.Com
 
 	logger.Info("Phase: Running")
 
-	if instance.GetScanType() == compv1alpha1.ScanTypePlatform {
+	switch instance.GetScanType() {
+	case compv1alpha1.ScanTypePlatform:
```
jhrozek, line 317: If we're already switching to switch-case, what do you think about adding a default and log the wrong scan type? In general I agree that the code shouldn't do "something" unless it knows what the scan is so either if-elseif-else or switch-case is preferable over if-else.

JAORMX: nothing unknown should come at this stage, else it'll panic. Validations are done in the pending phase which would prevent this.

jhrozek: Thanks, explained on slack

PR #387, opened 2020-08-05
WIP: Enforce resource limits for all pods

JAORMX:
This enforces resource limits for all pods related to the operator.

**For the operator itself:**

We were using three replicas and not settings the resources. This was
unnecessary, as it's not a critical workload. So instead let's just
ensure we have one replica running, and then reduce the resource limits
to have 150m of cpu and 200M of memory. These numbers were taken from
what etcd uses.

**aggregator:**
* cpu: 100m
* memory: 100M

**resultserver:**
* cpu: 100m
* memory: 100M

**logcollector:**
* cpu: 100m
* memory: 100M

**openscap:**
* cpu: 150m
* memory: 200M

**api resource collector:**
* cpu: 100m
* memory: 100M

**suite rerunner:**
* cpu: 10m
* memory: 20M

**profileparser:**
* cpu: 100m
* memory: 100M

Note that OpenSCAP gets more resources since it's known to use a lot of
resources.

### DIFF
--- a/cmd/manager/profileparser.go
+++ b/cmd/manager/profileparser.go
@@ -141,13 +141,14 @@ func runProfileParser(cmd *cobra.Command, args []string) {
 		updateProfileBundleStatus(pcfg, pb, fmt.Errorf("Couldn't read content file: %s", err))
 		os.Exit(1)
 	}
-	// #nosec
-	defer contentFile.Close()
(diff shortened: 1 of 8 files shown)


jhrozek:
@JAORMX  is this ready or pending some performance enhancements? Could you mark this as WIP or hold if the PR is not ready?

PR #388, opened 2020-08-05
Add a troubleshooting document

jhrozek:
Adds a markdown document that adds some general troubleshooting tips
as well as describes a flow of a Suite and a Remediations, which should
hopefully make it easier for users to understand how the operator works.

### DIFF
--- a/doc/troubleshooting.md
+++ b/doc/troubleshooting.md
@@ -147,11 +147,11 @@ In this phase, several `ConfigMaps` that contain either environment for the
 scanner pods or directly the script that the scanner pods will be evaluating.
 List the CMs with:
 ```shell
-oc get cm -lcomplianceoperator.openshift.io/scan=$scan_name
+oc get cm -lcompliance.openshift.io/scan-name=$scan_name,complianceoperator.openshift.io/scan-script=
 ```
 e.g.:
 ```shell
-oc get cm -lcomplianceoperator.openshift.io/scan=rhcos4-e8-worker
+oc get cm -lcompliance.openshift.io/scan-name=rhcos4-e8-worker,complianceoperator.openshift.io/scan-script=
 ```
 These `ConfigMaps` will be used by the scanner pods. If you ever needed to
 modify the scanner behaviour, change the scanner debug level or print the
@@ -181,9 +181,9 @@ a `Platform` scan instance and one scanner pod per matching node for a
 `Node` scan instance. The per-node pods are labeled with the node name,
 each pod is always labeled with the `ComplianceScan` name, for example:
 ```
-oc get pods -lcomplianceScan=rhcos4-e8-worker --show-labels
+oc get pods -lcompliance.openshift.io/scan-name=rhcos4-e8-worker --show-labels
 NAME                                                              READY   STATUS      RESTARTS   AGE   LABELS
-rhcos4-e8-worker-ip-10-0-169-90.eu-north-1.compute.internal-pod   0/2     Completed   0          39m   complianceScan=rhcos4-e8-worker,targetNode=ip-10-0-169-90.eu-north-1.compute.internal
+rhcos4-e8-worker-ip-10-0-169-90.eu-north-1.compute.internal-pod   0/2     Completed   0          39m   compliance.openshift.io/scan-name=rhcos4-e8-worker,targetNode=ip-10-0-169-90.eu-north-1.compute.internal
 ```
 
 At this point, the scan proceeds to the Running phase.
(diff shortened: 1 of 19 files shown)


doc/troubleshooting.md
```diff
@@ -0,0 +1,362 @@
...
+provide information in a bug report.
+
+Please refer to the README for general information about the API objects.
+
+## General tips
+
+   * The Compliance Operator emits Kubernetes events when something
+     important happens. You can either view all events in the cluster (`oc get events
+     -nopenshift-compliance`) or events for an object, e.g. for a scan
+     (`oc get compliancescan/$scan`)
```
JAORMX, line 13: wouldn't you want to use `oc describe <object>/<name>`? since that outputs the events as well for that object/

doc/troubleshooting.md
```diff
@@ -0,0 +1,362 @@
...
+
+   * The Compliance Operator consists of several controllers, roughly
+     one per API object. It could be handy to filter only those controller that pertain to
+     the API object having issues, e.g. if a `ComplianceRemediation` can't be applied,
+     the first place to look might be messages from the `remediationctrl` controller. You can filter the messages
+     from a single controller e.g. using `jq`:
+     `oc logs compliance-operator-775d7bddbd-gj58f | jq -c 'select(.logger == "controller_profilebundle")' `
```
JAORMX, line 20: It would be `profilebundlectrl` nowadays

doc/troubleshooting.md
```diff
@@ -0,0 +1,362 @@
...
+     that match the `machineconfiguration.openshift.io/role` label - see the
+     MachineConfig Operator documentation for more details.
+
+The remediation loop ends once the rendered MachineConfig is updated, if needed,
+and the reconciled remediation object status is updated.
+
+In our case, applying the remediation would trigger a reboot. After that we can
+annotate the scan to re-run it:
+```
+oc annotate compliancescans/rhcos4-e8-worker compliance.openshift.io/rescan
```
pdhamdhe, line 354: The command would be :

> $ oc annotate compliancescans/worker-scan compliance.openshift.io/rescan=
> compliancescan.compliance.openshift.io/worker-scan annotated

  to annotate the scan at line 290 & 354 otherwise, it reports the following error :

> $ oc annotate compliancescans/worker-scan compliance.openshift.io/rescan
> error: at least one annotation update is required

doc/troubleshooting.md
jhrozek: of course, fixed

doc/troubleshooting.md
jhrozek: fixed

doc/troubleshooting.md
jhrozek: good catch, fixed

jhrozek:
By the way, while writing this document I found out that we are quite inconsistent when it comes to label keys. Sometimes if a CR belongs to a scan, we label it with `complianceoperator.openshift.io/scan` (CheckResults.), sometimes just scan-name (PVC) and sometimes `compliance-scan.
I think it would be nice to make the labels consistent. Thoughts?

JAORMX:
> By the way, while writing this document I found out that we are quite inconsistent when it comes to label keys. Sometimes if a CR belongs to a scan, we label it with `complianceoperator.openshift.io/scan` (CheckResults.), sometimes just scan-name (PVC) and sometimes `compliance-scan.
> I think it would be nice to make the labels consistent. Thoughts?

That would be nice!

jhrozek:
On Mon, Aug 10, 2020 at 09:59:00PM -0700, Juan Osorio Robles wrote:
> > By the way, while writing this document I found out that we are quite inconsistent when it comes to label keys. Sometimes if a CR belongs to a scan, we label it with `complianceoperator.openshift.io/scan` (CheckResults.), sometimes just scan-name (PVC) and sometimes `compliance-scan.
> > I think it would be nice to make the labels consistent. Thoughts?
> 
> That would be nice!

Done in a follow-up patch.

jhrozek:
@JAORMX if you're OK with this PR, can you add the lgtm label?

jhrozek:
The `Inconsistent` e2e test is failing because it is expecting a cluster with 3 worker nodes, then if one of them would be inconsistent and 2 consistent, we could still found the "most common state". But in the e2e test we are receiving a cluster with only 2 worker nodes, so one of them reports one result and another node reports other result, so there's no way to get a "quorum". I guess we need to take this into account and support both cases..

PR #389, opened 2020-08-07
Create one reference ScanSettings object by default

JAORMX:
to enhance the usability of the operator, this creates a ScanSettings
object called `default`, which contains reference values that could be
taken into use.

The default values are:
* raw storage allocation of 1Gi
* rotation policy of 3
* debugging turned off
* tolerations that allow scanning master nodes
* auto applying remediations is off
* the scans will run every day at 1am

This way, the only thing that an admin would need to do is create a
scanSettingBinding object to start running scans.

### DIFF
--- a/cmd/manager/operator.go
+++ b/cmd/manager/operator.go
@@ -57,6 +57,9 @@ var (
 		"rhcos4",
 		"ocp4",
 	}
+	defaultScanSettingsName = "default"
+	// Run scan every day at 1am
+	defaultScanSettingsSchedule = "0 1 * * *"
 )
 
 func printVersion() {
@@ -106,13 +109,14 @@ func RunOperator(cmd *cobra.Command, args []string) {
 		Namespace:          namespace,
 		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
 	}
+	namespaceList := strings.Split(namespace, ",")
 	// Add support for MultiNamespace set in WATCH_NAMESPACE (e.g ns1,ns2)
 	// Note that this is not intended to be used for excluding namespaces, this is better done via a Predicate
 	// Also note that you may face performance issues when using this with a high number of namespaces.
 	// More Info: https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/cache#MultiNamespacedCacheBuilder
 	if strings.Contains(namespace, ",") {
 		options.Namespace = ""
-		options.NewCache = cache.MultiNamespacedCacheBuilder(strings.Split(namespace, ","))
+		options.NewCache = cache.MultiNamespacedCacheBuilder(namespaceList)
 	}
 
 	// Get a config to talk to the apiserver
@@ -159,8 +163,12 @@ func RunOperator(cmd *cobra.Command, args []string) {
 	// Add the Metrics Service
 	addMetrics(ctx, cfg)
 
-	if err := ensureDefaultProfileBundles(ctx, mgr.GetClient(), namespace); err != nil {
+	if err := ensureDefaultProfileBundles(ctx, mgr.GetClient(), namespaceList); err != nil {
 		log.Error(err, "Error creating default ProfileBundles. Continuing anyway.")
+	}
+
+	if err := ensureDefaultScanSettings(ctx, mgr.GetClient(), namespaceList); err != nil {
+		log.Error(err, "Error creating default ScanSettings. Continuing anyway.")
 	}
 
 	log.Info("Starting the Cmd.")
@@ -216,12 +224,11 @@ func addMetrics(ctx context.Context, cfg *rest.Config) {
 	}
 }
 
-func ensureDefaultProfileBundles(ctx context.Context, crclient client.Client, namespaces string) error {
+func ensureDefaultProfileBundles(ctx context.Context, crclient client.Client, namespaceList []string) error {
 	pbimg := utils.GetComponentImage(utils.DEFAULT_PROFILE_BUNDLES)
 	var lastErr error
-	nsList := strings.Split(namespaces, ",")
 	for _, prod := range defaultProducts {
-		for _, ns := range nsList {
+		for _, ns := range namespaceList {
 			pb := &compv1alpha1.ProfileBundle{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      prod,
@@ -236,6 +243,30 @@ func ensureDefaultProfileBundles(ctx context.Context, crclient client.Client, na
 			if !k8serrors.IsAlreadyExists(err) {
 				lastErr = err
 			}
+		}
+	}
+	return lastErr
+}
+
+func ensureDefaultScanSettings(ctx context.Context, crclient client.Client, namespaceList []string) error {
+	var lastErr error
+	for _, ns := range namespaceList {
+		pb := &compv1alpha1.ScanSetting{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      defaultScanSettingsName,
+				Namespace: ns,
+			},
+			ComplianceSuiteSettings: compv1alpha1.ComplianceSuiteSettings{
+				Schedule: defaultScanSettingsSchedule,
+			},
+			Roles: []string{
+				"worker",
+				"master",
+			},
+		}
+		err := crclient.Create(ctx, pb)
+		if !k8serrors.IsAlreadyExists(err) {
+			lastErr = err
 		}
 	}
 	return lastErr


cmd/manager/operator.go
```diff
@@ -241,6 +249,45 @@ func ensureDefaultProfileBundles(ctx context.Context, crclient client.Client, na
...
+	var lastErr error
+	for _, ns := range namespaceList {
+		pb := &compv1alpha1.ScanSetting{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      defaultScanSettingsName,
+				Namespace: ns,
+			},
+			ComplianceScanSettings: compv1alpha1.ComplianceScanSettings{
+				Debug: false,
+				RawResultStorage: compv1alpha1.RawResultStorageSettings{
```
jhrozek, line 262: I like the general idea of creating the default object, but do we need to explicitly mention the defaults? It appears that the defaults are applied by the respective controllers when the objects are already created. And should we ever change the defaults, the example object wouldn't change unless we also update it, unlike the defaults applied when creating the CRs.

JAORMX: We probably don't need to specify defaults for the default object. Let's do this, I'll only leave the defaults that come from a constant.

jhrozek:
I have two remarks, not hard nacks, but I'm interested in what you think
1) There seem to be a bug somewhere, I'm sure it is not related to your patch, but I think we should file an issue and try to get to the root of it. The reason some of the scans fail is:
```
{"level":"info","ts":1597125319.7650719,"logger":"scanctrl","msg":"Node scan error","node.Name":"ip-10-0-166-55.us-west-2.compute.internal","errMsg":"the ConfigMap 'openscap-pod-fd08fa28d8a49f1f9c75c264db59ab60ebc5e38a' annotation was missing the result key"}
```
but, the configmap DOES have the result key:
```
"kind": "ConfigMap",
            "metadata": {
                "annotations": {
                    "compliance-remediations/processed": "",
                    "compliance.openshift.io/scan-error-msg": "",
                    "compliance.openshift.io/scan-result": "COMPLIANT",
                    "openscap-scan-result/node": "ip-10-0-166-55.us-west-2.compute.internal"
                },
                "creationTimestamp": "2020-08-11T05:53:11Z",
                "labels": {
                    "compliance-scan": "test-scan-w-missing-tailoring-cm"
                },
                "managedFields": [
                    {
                        "apiVersion": "v1",
                        "fieldsType": "FieldsV1",
                        "fieldsV1": {
                            "f:data": {
                                ".": {},
                                "f:exit-code": {},
                                "f:results": {}
                            },
                            "f:metadata": {
                                "f:annotations": {
                                    ".": {},
                                    "f:compliance-remediations/processed": {},
                                    "f:compliance.openshift.io/scan-error-msg": {},
                                    "f:compliance.openshift.io/scan-result": {},
                                    "f:openscap-scan-result/node": {}
                                },
                                "f:labels": {
                                    ".": {},
                                    "f:compliance-scan": {}
                                }
                            }
                        },
                        "manager": "compliance-operator",
                        "operation": "Update",
                        "time": "2020-08-11T05:55:17Z"
                    }
                ],
                "name": "openscap-pod-fd08fa28d8a49f1f9c75c264db59ab60ebc5e38a",
```
and the aggregator pod says that it had annotated the objects:
```
{"level":"info","ts":1597125316.3485558,"logger":"cmd","msg":"Annotating ConfigMaps"}
```
Looking at the timestamps, the annotating configMaps message was emitted at 1597125316.3485558 (=Tue 11 Aug 2020 05:55:16 AM UTC) and the check for the annotations happened at 1597125319.7650719 (=Tue 11 Aug 2020 05:55:19 AM UTC), so three seconds later.
Could it be that the informer is returning obsolete data? I'm really not sure how to handle this, but at the same time it's not the first time I've seen this issue either so I wouldn't like to just ignore it..

2) It still seems that the `RawResultStorage` defaults are not necessary

JAORMX:
> I have two remarks, not hard nacks, but I'm interested in what you think
> 
>     1. There seem to be a bug somewhere, I'm sure it is not related to your patch, but I think we should file an issue and try to get to the root of it. The reason some of the scans fail is:
> 
> 
> ```
> {"level":"info","ts":1597125319.7650719,"logger":"scanctrl","msg":"Node scan error","node.Name":"ip-10-0-166-55.us-west-2.compute.internal","errMsg":"the ConfigMap 'openscap-pod-fd08fa28d8a49f1f9c75c264db59ab60ebc5e38a' annotation was missing the result key"}
> ```
> 
> but, the configmap DOES have the result key:
> 
> ```
> "kind": "ConfigMap",
>             "metadata": {
>                 "annotations": {
>                     "compliance-remediations/processed": "",
>                     "compliance.openshift.io/scan-error-msg": "",
>                     "compliance.openshift.io/scan-result": "COMPLIANT",
>                     "openscap-scan-result/node": "ip-10-0-166-55.us-west-2.compute.internal"
>                 },
>                 "creationTimestamp": "2020-08-11T05:53:11Z",
>                 "labels": {
>                     "compliance-scan": "test-scan-w-missing-tailoring-cm"
>                 },
>                 "managedFields": [
>                     {
>                         "apiVersion": "v1",
>                         "fieldsType": "FieldsV1",
>                         "fieldsV1": {
>                             "f:data": {
>                                 ".": {},
>                                 "f:exit-code": {},
>                                 "f:results": {}
>                             },
>                             "f:metadata": {
>                                 "f:annotations": {
>                                     ".": {},
>                                     "f:compliance-remediations/processed": {},
>                                     "f:compliance.openshift.io/scan-error-msg": {},
>                                     "f:compliance.openshift.io/scan-result": {},
>                                     "f:openscap-scan-result/node": {}
>                                 },
>                                 "f:labels": {
>                                     ".": {},
>                                     "f:compliance-scan": {}
>                                 }
>                             }
>                         },
>                         "manager": "compliance-operator",
>                         "operation": "Update",
>                         "time": "2020-08-11T05:55:17Z"
>                     }
>                 ],
>                 "name": "openscap-pod-fd08fa28d8a49f1f9c75c264db59ab60ebc5e38a",
> ```
> 
> and the aggregator pod says that it had annotated the objects:
> 
> ```
> {"level":"info","ts":1597125316.3485558,"logger":"cmd","msg":"Annotating ConfigMaps"}
> ```
> 
> Looking at the timestamps, the annotating configMaps message was emitted at 1597125316.3485558 (=Tue 11 Aug 2020 05:55:16 AM UTC) and the check for the annotations happened at 1597125319.7650719 (=Tue 11 Aug 2020 05:55:19 AM UTC), so three seconds later.
> Could it be that the informer is returning obsolete data? I'm really not sure how to handle this, but at the same time it's not the first time I've seen this issue either so I wouldn't like to just ignore it..

Right, it could be that it's obsolete data. Mostly we're using the controller-runtime's client, which uses a cache client to do the Get and List operations. These will hold objects for a specific amount of time and thus reduce the hits we do to the api server... but, it has the issue that some objects might be old. I guess this is not much of an issue for our main controllers, but the aggregator is having issues with this. So... we either move away from using the controller-runtime client and always hit the apiserver, or we address this limitation in the code.


> 
>     1. It still seems that the `RawResultStorage` defaults are not necessary

jhrozek:
On Tue, Aug 11, 2020 at 05:14:25AM -0700, Juan Osorio Robles wrote:
> > Looking at the timestamps, the annotating configMaps message was emitted at 1597125316.3485558 (=Tue 11 Aug 2020 05:55:16 AM UTC) and the check for the annotations happened at 1597125319.7650719 (=Tue 11 Aug 2020 05:55:19 AM UTC), so three seconds later.
> > Could it be that the informer is returning obsolete data? I'm really not sure how to handle this, but at the same time it's not the first time I've seen this issue either so I wouldn't like to just ignore it..
> 
> Right, it could be that it's obsolete data. Mostly we're using the
> controller-runtime's client, which uses a cache client to do the Get and
> List operations. These will hold objects for a specific amount of time and
> thus reduce the hits we do to the api server... but, it has the issue that
> some objects might be old. I guess this is not much of an issue for our
> main controllers, but the aggregator is having issues with this. So... we
> either move away from using the controller-runtime client and always hit
> the apiserver, or we address this limitation in the code.

What bothers me is that we haven't really seen an issue with the cache
so far except this one. And I know too little about the informers and when does the
cache gets invalidated or updated to tell if it's really the root cause.

But, looping again to see if that helps might be an easy patch. I'll
take a look at submitting some test PR that backs off and retries a
number of times to see if it helps.

> 
> 
> > 
> >     1. It still seems that the `RawResultStorage` defaults are not necessary

What do you think about this?

JAORMX:
> On Tue, Aug 11, 2020 at 05:14:25AM -0700, Juan Osorio Robles wrote: > Looking at the timestamps, the annotating configMaps message was emitted at 1597125316.3485558 (=Tue 11 Aug 2020 05:55:16 AM UTC) and the check for the annotations happened at 1597125319.7650719 (=Tue 11 Aug 2020 05:55:19 AM UTC), so three seconds later. > Could it be that the informer is returning obsolete data? I'm really not sure how to handle this, but at the same time it's not the first time I've seen this issue either so I wouldn't like to just ignore it.. Right, it could be that it's obsolete data. Mostly we're using the controller-runtime's client, which uses a cache client to do the Get and List operations. These will hold objects for a specific amount of time and thus reduce the hits we do to the api server... but, it has the issue that some objects might be old. I guess this is not much of an issue for our main controllers, but the aggregator is having issues with this. So... we either move away from using the controller-runtime client and always hit the apiserver, or we address this limitation in the code.
> What bothers me is that we haven't really seen an issue with the cache so far except this one. And I know too little about the informers and when does the cache gets invalidated or updated to tell if it's really the root cause. But, looping again to see if that helps might be an easy patch. I'll take a look at submitting some test PR that backs off and retries a number of times to see if it helps.

That would be good, thanks!

> > > 1. It still seems that the `RawResultStorage` defaults are not necessary
> What do you think about this?

Done

PR #390, opened 2020-08-07
Deprecate Type field from ComplianceRemediation object

JAORMX:
We shouldn't ship with deprecated values. This is not used anywhere.

### DIFF
--- a/pkg/apis/compliance/v1alpha1/complianceremediation_types.go
+++ b/pkg/apis/compliance/v1alpha1/complianceremediation_types.go
@@ -32,8 +32,6 @@ const (
 )
 
 type ComplianceRemediationSpecMeta struct {
-	// Remediation type specifies the artifact the remediation is based on. For now, only MachineConfig is supported
-	Type RemediationType `json:"type,omitempty"`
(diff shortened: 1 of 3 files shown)


jhrozek:
Why don't we also deprecate handling of the deprecated values in the remediation controller and the conversion into the new format? Do you think that it is still needed?

JAORMX:
> Why don't we also deprecate handling of the deprecated values in the remediation controller and the conversion into the new format? Do you think that it is still needed?

I can do that in a subsequent PR.

You mean the "machineConfig" key in the remediations, right?

jhrozek:
On Mon, Aug 10, 2020 at 06:04:23AM -0700, Juan Osorio Robles wrote:
> > Why don't we also deprecate handling of the deprecated values in the remediation controller and the conversion into the new format? Do you think that it is still needed?
> 
> I can do that in a subsequent PR.

OK, that's fine.

> 
> You mean the "machineConfig" key in the remediations, right?

I think it's called `machineConfigContents`, what I was mostly after is
eliminating the `updateDeprecatedMcRemediation()` function and related
logic in the remediations controller.

PR #392, opened 2020-08-11
profileparser: Close XML file when run has ended

JAORMX:
The file was left lingering open for the rest of the duration of the
run... so let's close it explicitly instead of relying on the defer
which will never run.

### DIFF
--- a/cmd/manager/profileparser.go
+++ b/cmd/manager/profileparser.go
@@ -141,13 +141,14 @@ func runProfileParser(cmd *cobra.Command, args []string) {
 		updateProfileBundleStatus(pcfg, pb, fmt.Errorf("Couldn't read content file: %s", err))
 		os.Exit(1)
 	}
-	// #nosec
-	defer contentFile.Close()
 	bufContentFile := bufio.NewReader(contentFile)
 	contentDom, err := xmldom.Parse(bufContentFile)
 	if err != nil {
 		log.Error(err, "Couldn't read the content XML")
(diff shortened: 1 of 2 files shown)


jhrozek:
Can you explain in more details why would the defer statement not run?

JAORMX:
> Can you explain in more details why would the defer statement not run?

that's because we don't end the pod, we wait forever for the exit signal. This is done so the container is kept running and eventually, when a new image is used, the kube deployment will replace the running container automatically and an update will happen.

jhrozek:
> > Can you explain in more details why would the defer statement not run?
> 
> that's because we don't end the pod, we wait forever for the exit signal. This is done so the container is kept running and eventually, when a new image is used, the kube deployment will replace the running container automatically and an update will happen.

But do we get a fd leak in this case? The new container is a new process, right, so the file gets closed automatically. I mean, there is no issue with the patch, I'm just trying to understand what issue is the patch trying to solve.

JAORMX:
> > > Can you explain in more details why would the defer statement not run?
> > 
> > 
> > that's because we don't end the pod, we wait forever for the exit signal. This is done so the container is kept running and eventually, when a new image is used, the kube deployment will replace the running container automatically and an update will happen.
> 
> But do we get a fd leak in this case? The new container is a new process, right, so the file gets closed automatically. I mean, there is no issue with the patch, I'm just trying to understand what issue is the patch trying to solve.

Right, we don't leak cause it's a new process. This merely cleans up this file descriptor.

PR #394, opened 2020-08-12
Detect if scan is unschedulable

JAORMX:
If a pod (or more) is not schedulable in the scan due to a taint or
another scheduling issue, we detect this and issue partial results;
however, an error is persisted as the ultimate result of the scan.

This is done by creating a configMap (as the log-collector would do)
that contains a status that the node didn't get a scan due to it being
unschedulable, and adding the pod's condition as an error message.

This also changes how the aggregator is scheduled, as it now only gets
skipped if there's an unknown error.

### DIFF
--- a/pkg/controller/compliancescan/compliancescan_controller.go
+++ b/pkg/controller/compliancescan/compliancescan_controller.go
@@ -2,6 +2,7 @@ package compliancescan
 
 import (
 	"context"
+	goerrors "errors"
 	"fmt"
 	"math"
 	"time"
@@ -346,6 +347,7 @@ func (r *ReconcileComplianceScan) phaseRunningHandler(instance *compv1alpha1.Com
 
 		// On each eligible node..
 		for _, node := range nodes.Items {
+			var unschedulableErr *podUnschedulableError
 			running, err := isPodRunningInNode(r, instance, &node, logger)
 			if errors.IsNotFound(err) {
 				// Let's go back to the previous state and make sure all the nodes are covered.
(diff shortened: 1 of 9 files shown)


JAORMX:
@jhrozek PTAL

jhrozek:
I'm going to take a look this afternoon/evening.

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -357,6 +359,25 @@ func (r *ReconcileComplianceScan) phaseRunningHandler(instance *compv1alpha1.Com
...
 				return reconcile.Result{}, nil
+			} else if goerrors.As(err, &unschedulableErr) {
+				// Create custom error message for this pod that couldn't be scheduled
+				cmName := getConfigMapForNodeName(instance.Name, node.Name)
+				cm := utils.GetResultConfigMap(instance, cmName, "error-msg", node.Name,
+					[]byte(err.Error()), false, common.PodUnschedulableExitCode)
+				cmKey := types.NamespacedName{Name: cm.Name, Namespace: cm.Namespace}
+				foundcm := corev1.ConfigMap{}
+				cmGetErr := r.client.Get(context.TODO(), cmKey, &foundcm)
+				if errors.IsNotFound(cmGetErr); cmGetErr != nil {
```
jhrozek, line 370: This condition and the following one sounds strange, wouldn't the first condition hit for any `cmGetErr != nil` meaning that the second condition is never hit? Don't you just mean to say `if errors.IsNotFound(cmGetErr)` ?

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -675,18 +696,30 @@ func isPlatformScanPodRunning(r *ReconcileComplianceScan, scanInstance *compv1al
 }
 
 func isPodRunning(r *ReconcileComplianceScan, podName, namespace string, logger logr.Logger) (bool, error) {
+	podlogger := logger.WithValues("Pod.Name", podName)
```
jhrozek, line 699: Out of curiosity, what does this do? Does it add a key-value pair to the logger that is passed to the function by default?

JAORMX: yep

JAORMX:
@jhrozek addressed the comments.

jhrozek:
I'm afraid there's a conflict now. You'll need to apply this diff:
```
diff --git a/pkg/utils/resultconfigmap.go b/pkg/utils/resultconfigmap.go
index 3fd1278..571a4ed 100644
--- a/pkg/utils/resultconfigmap.go
+++ b/pkg/utils/resultconfigmap.go
@@ -40,7 +40,7 @@ func GetResultConfigMap(owner metav1.Object, configMapName, filename, nodeName s
                        Namespace:   common.GetComplianceOperatorNamespace(),
                        Annotations: annotations,
                        Labels: map[string]string{
-                               compv1alpha1.ComplianceScanIndicatorLabel: owner.GetName(),
+                               compv1alpha1.ComplianceScanLabel: owner.GetName(),
                        },
                },
                Data: map[string]string{

```
otherwise lgtm, the condition appears to be fixed now.

jhrozek:
This version of the PR looks good. I'll ack it once the release PR merges.

jhrozek:
FWIW, the latest test failures seem to be unrelated to our code:
```
 2020/08/19 05:52:27 Create release image default-route-openshift-image-registry.apps.build01.ci.devcluster.openshift.com/ci-op-yykz4csq/release:latest
{"component":"entrypoint","file":"prow/entrypoint/run.go:165","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2020-08-19T09:39:59Z"} 
```

jhrozek (approved):
/lgtm

PR #396, opened 2020-08-12
Clean up suite-rerunner pods if suite is deleted

JAORMX:
The CronJob object doesn't fully clean up all objects; so we need to do
that ourselves.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -6,7 +6,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - pods        # Pods are used to run the compliance container
   - persistentvolumeclaims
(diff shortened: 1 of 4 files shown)


JAORMX:
> This looks OK conceptually, but I would recommend to use two labels for the cronojb pods: one with the suite like you did and one that would signal that the pod is actually a cronjob pod, e.g. for configmaps we use "complianceoperator.openshift.io/scan-script=". I don't know if cronjob pods already have some useful labels, feel free to use something existing.
> The reason I think this would be useful is that otherwise it might be easy to delete more than you want epecially if we add more similarly labeled pods in the future.

Done.

PR #398, opened 2020-08-12
Don't touch other profiles when updating one profile

jhrozek:
The code would end up checking all profiles in a namespace and
subsequently deleting content from profiles which don't have the same
has as the one updated, this patch restricts the update to the profile
being updated by using a label.

Jira: OCPBUGSM-13679

### DIFF
--- a/pkg/profileparser/profileparser.go
+++ b/pkg/profileparser/profileparser.go
@@ -83,7 +83,7 @@ func ParseBundle(contentDom *xmldom.Document, pb *cmpv1alpha1.ProfileBundle, pcf
 		return err
 	}
 
-	if err := deleteObsoleteItems(pcfg.Client, "Profile", pb.Namespace, pb.GetImageDigest()); err != nil {
+	if err := deleteObsoleteItems(pcfg.Client, "Profile", pb.Name, pb.Namespace, pb.GetImageDigest()); err != nil {
 		return err
 	}
 
@@ -114,7 +114,7 @@ func ParseBundle(contentDom *xmldom.Document, pb *cmpv1alpha1.ProfileBundle, pcf
 		return err
 	}
 
-	if err := deleteObsoleteItems(pcfg.Client, "Rule", pb.Namespace, pb.GetImageDigest()); err != nil {
+	if err := deleteObsoleteItems(pcfg.Client, "Rule", pb.Name, pb.Namespace, pb.GetImageDigest()); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/profileparser/profileparser.go
```diff
@@ -199,17 +199,20 @@ func createOrUpdate(cli runtimeclient.Client, kind string, key types.NamespacedN
...
 	})
 	inNs := runtimeclient.InNamespace(namespace)
+	withPbOwnerLabel := runtimeclient.MatchingLabels{
+		cmpv1alpha1.ProfileBundleOwnerLabel: pbName,
+	}
 
-	log.Info("Checking for unused object", "kind", kind, "namespace", namespace)
-	if err := cli.List(context.TODO(), &list, inNs); err != nil {
+	log.Info("Checking for unused object", "kind", kind, "owner", pbName, "namespace", namespace)
+	if err := cli.List(context.TODO(), &list, inNs, withPbOwnerLabel); err != nil {
```
JAORMX, line 215: Why not refactor this to use `DeleteAllOf` as done here https://github.com/openshift/compliance-operator/pull/396 ?

JAORMX: Note that you'd need to add an extra permission via RBAC to do this.

jhrozek:
On Wed, Aug 12, 2020 at 05:04:39AM -0700, Juan Osorio Robles wrote:
> @JAORMX commented on this pull request.
> 
> 
> 
> >  
> -	log.Info("Checking for unused object", "kind", kind, "namespace", namespace)
> -	if err := cli.List(context.TODO(), &list, inNs); err != nil {
> +	log.Info("Checking for unused object", "kind", kind, "owner", pbName, "namespace", namespace)
> +	if err := cli.List(context.TODO(), &list, inNs, withPbOwnerLabel); err != nil {
> 
> Note that you'd need to add an extra permission via RBAC to do this.

Is it the deletecollection permission? Overall seems like a good idea.

JAORMX:
> On Wed, Aug 12, 2020 at 05:04:39AM -0700, Juan Osorio Robles wrote: @JAORMX commented on this pull request. > - log.Info("Checking for unused object", "kind", kind, "namespace", namespace) - if err := cli.List(context.TODO(), &list, inNs); err != nil { + log.Info("Checking for unused object", "kind", kind, "owner", pbName, "namespace", namespace) + if err := cli.List(context.TODO(), &list, inNs, withPbOwnerLabel); err != nil { Note that you'd need to add an extra permission via RBAC to do this.
> Is it the deletecollection permission? Overall seems like a good idea.

exactly

pkg/profileparser/profileparser.go
jhrozek: Ah, OK, I remembered why I didn't use DeleteAllOf..the hashes are stored as annotations because a typical sha256 hash is longer than what labels would allow. And typically the fields you use as DeleteOptions or ListOptions must be indexed, annotations are not. It appears that it's possible to index annotations and then use MatchingFields (e.g. https://github.com/kubermatic/utils/blob/39ab8a8d6830548e12f069ad75216c2849065907/pkg/multiowner/multi_owner.go) but I'm not sure if it's a good pattern.

But that brings a more important quesition: do you think it's OK to use annotations and loop? My thought when I wrote the code was that updating a profile is a rare operation and not really performance sensitive. But if you don't like that we can either index the annotations or somehow put the sha hash into the labels (e.g. quay or git use shortened hash versions, but I wasn't able to find the details. I haven't looked too deep though tbh)

PR #401, opened 2020-08-17
readme: Document that just deleting the namespace might not process finalizers

jhrozek:
Jira: OCPBUGSM-4630

### DIFF
--- a/README.md
+++ b/README.md
@@ -683,6 +683,21 @@ make run
 ```
 This is mostly useful for local development.
 
+### Note on namespace removal
+Many custom resources deployed with the compliance operators use finalizers
+to handle dependencies between objects. If the whole operator namespace gets
+deleted (e.g. with `oc delete ns openshift-compliance`), the order of deleting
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -683,6 +683,21 @@ make run
 ```
 This is mostly useful for local development.
 
+### Note on namespace removal
+Many custom resources deployed with the compliance operators use finalizers
+to handle dependencies between objects. If the whole operator namespace gets
+deleted (e.g. with `oc delete ns/openshift-operator`), the order of deleting
```
JAORMX, line 689: was `ns/openshift-operator` intentional?

jhrozek:
The tests failed because of the issue in CMP-690

README.md
jhrozek: No, that was a typo thanks..

(Hello, my name is Jakub, I'm a Senior Software Engineer and I have trouble patching a README :-) ...)

PR #403, opened 2020-08-17
Permit creating remediations from a ComplianceScan

jhrozek:
I'm opening this PR as a WIP because my cluster is cleaned up for the day
and I'd like to run the e2e tests to see if this is a viable approach..

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@ operator to be deployed in other namespaces as well.
 The primary interface towards the Compliance Operator is the
 `ComplianceSuite` object, representing a set of scans. The `ComplianceSuite`
 can be defined either manually or with the help of `ScanSetting` and
-`ScanSettingBinding` objects.
+`ScanSettingBinding` objects. Note that while it is possible to use the
+lower-level `ComplianceScan` directly as well, it is not recommended.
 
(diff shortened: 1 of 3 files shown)


jhrozek:
of course there was a bug, let's see if this version fares any better while I work on a test

JAORMX:
> The test is rather minimal, it just ensures that the remediations are created and all of them have the expected state. The test doesn't actually apply the remediations: let me know if that test would be preferable and I can change the patch. I just didn't want to introduce another non-parallel test for a corner case.

No, this is fine. It tests this specific functionality and that's exactly what we wanted.

jhrozek:
@JAORMX the PR is ready for review

JAORMX (commented):
/lgtm

jhrozek:
This actually needs a rebase now:
```
# github.com/openshift/compliance-operator/tests/e2e [github.com/openshift/compliance-operator/tests/e2e.test]
tests/e2e/e2e_test.go:203:40: undefined: "github.com/openshift/compliance-operator/pkg/apis/compliance/v1alpha1".ScanLabel
FAIL	github.com/openshift/compliance-operator/tests/e2e [build failed]
```

PR #404, opened 2020-08-18
Make access modes and storage class configurable for PVs

JAORMX:
In cases where there isn't a default storage class configured or simply
the admin wants to use something other than the default, we are now able
to configure the storage class through the raw result storage settings
in the scan/suite.

This requires us to be able to set the access modes, which could vary
between storage classes. By making this configurable, this gives the
advantage of allowing admins to have a long-running process that
forwards the raw results if the storage class supports a ReadWriteMany
access mode. Note that it would still be easier to just clone the
volume...


jhrozek (approved):
/lgtm

jhrozek:
This version of the PR looks good. I'll ack it once the release PR merges.
btw looking at the test failures it looks like the cluster couldn't even be installed..so nothing for us to check here..

jhrozek:
Oops, sorry, I meant to comment to another PR, the point about why the test fail still stands though..

jhrozek:
FWIW, the latest test failures seem to be unrelated to our code:
```
 2020/08/19 05:52:27 Create release image default-route-openshift-image-registry.apps.build01.ci.devcluster.openshift.com/ci-op-yykz4csq/release:latest
{"component":"entrypoint","file":"prow/entrypoint/run.go:165","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2020-08-19T09:39:59Z"} 
```

PR #409, opened 2020-08-24
Operator deployment: Change replicas to only deploy one

JAORMX:
There is no reason why we would need 3 replicas of the operator running
at the same time. One is enough. The Deployment will ensure it's running
in case a crash happens.


jhrozek (approved):
I checked a couple of OCP deployments and they all had the replicas set to 1 as well
/lgtm

PR #410, opened 2020-08-24
Also push to latest when doing a release

JAORMX:
when doing a release, the push to latest didn't happen automatically.
This adds that.


JAORMX:
Pull-request updated, HEAD is now 682c0aec43baabc82d90c84b5eac908503bef10a

PR #412, opened 2020-08-24
Workaround: Only use profileparser as an init container

JAORMX:
The profileparser pod takes a lot of resources and doesn't let them
go... This shows up as excessive resource usage in the monitoring
dashboard (up to 400MiB).

By changing that profileparser to be an init container, as opposed to a
daemon, the parser process will end and the resources will be freed. The
daemon bit will now be a "pauser" container, which merely waits for a
kill signal. This way, this container takes a very minimal amount of
resources and it keeps running as is needed by the kubernetes' Deployment
that manages it.

### DIFF
--- a/cmd/manager/pause.go
+++ b/cmd/manager/pause.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/spf13/cobra"
+)
+
+var pauseCmd = &cobra.Command{
+	Use:   "pause",
+	Short: "A command that simply pauses",
+	Long:  `Pauses and waits for the SIGTERM signal`,
+	Run:   pause,
+}
+
+func init() {
+	rootCmd.AddCommand(pauseCmd)
+}
+
+func pause(cmd *cobra.Command, args []string) {
+	exitSignal := make(chan os.Signal, 1)
+	signal.Notify(exitSignal, syscall.SIGINT, syscall.SIGTERM)
+
+	<-exitSignal
+}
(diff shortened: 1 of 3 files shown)


cmd/manager/pause.go
```diff
@@ -0,0 +1,27 @@
...
+	Short: "A command that simply pauses",
+	Long:  `Pauses and waits for the SIGTERM signal`,
+	Run:   pause,
+}
+
+func init() {
+	rootCmd.AddCommand(pauseCmd)
+}
+
+func pause(cmd *cobra.Command, args []string) {
```
jhrozek, line 22: Nit: Could we add a log message (or just a fmt.Println) that tells the user that the actual work is done in an init container? Otherwise `oc logs` doesn't show anything.

jhrozek (commented):
Except for the nit inline, I wonder if there is any timeout for init containers? I couldn't find anything, but we should keep in mind that the profileparser can run for quite a bit..
Memory usage of the pod after the patch seems to be hovering around 50MB

JAORMX:
> Except for the nit inline, I wonder if there is any timeout for init containers? I couldn't find anything, but we should keep in mind that the profileparser can run for quite a bit..
> Memory usage of the pod after the patch seems to be hovering around 50MB

yeah... that's better than 400MB though :smile: ... but I'm guessing that's leaks from other files. Need to investigate those.

cmd/manager/pause.go
JAORMX: will do

jhrozek (approved):
/lgtm

PR #413, opened 2020-08-24
Also push to latest when doing a release

JAORMX:
This makes sure that when we do releases, the latest tag is also
updated.


jhrozek (approved):
/lgtm

PR #414, opened 2020-08-25
Fix memory leak in profileparser

JAORMX:
The reference parser is initialized as a global variable and allocated
in the profileparser's package init function.

While this works nicely for the profileparser subcommand, it does this
for every workload since we compile the subcommands together. This
causes a memory leak since the parser stays allocated and in memory.

This is fixed by adding the initialization as part of the
`newStandardParser()` function, and calling that explicitly where it's
needed.


jhrozek (approved):
/lgtm

PR #416, opened 2020-08-27
Adjust release process to use manifest bundles instead of package bundles

JAORMX:
Since OCP 4.6, package bundles are no longer supported in favor of using manifest bundles.

This requires some changes in the way we do releases, such as calling the bundle commands from `operator-sdk` and creating an index image in order to make the operator installable from a CatalogSource.


JAORMX:
Pull-request updated, HEAD is now ecd0d2bacfd6f3eecdea0d4af70d2f0612c7bcc1

jhrozek (approved):
/lgtm

PR #419, opened 2020-08-28
utils: Don't use ctime but mtime instead

jhrozek:
ctime updates even when the contents of the files don't update, e.g.
with permissions update or adding links. It appers that when the
resultserver is scaled up or down, the ctime is always updated,
resulting in the sort of the directories being quite random.

Let's use mtime instead.

Jira: OCPBUGSM-13482

### DIFF
--- a/cmd/manager/resultserver_test.go
+++ b/cmd/manager/resultserver_test.go
@@ -68,6 +68,18 @@ var _ = Describe("Resultserver testing", func() {
 			dir3, err = ioutil.TempDir(rootDir, "rotate-3")
 			Expect(err).To(BeNil())
 
+			// chmod the dirs in reverse order to make sure we are really relying on
+			// modification time and not change time (see OCPBUGSM-13482)
+			time.Sleep(5 * time.Millisecond)
+			err = os.Chmod(dir3, os.ModePerm)
+			Expect(err).To(BeNil())
+			time.Sleep(5 * time.Millisecond)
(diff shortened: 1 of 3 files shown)


jhrozek:
Maybe this should be marked as WIP because I still can't answer two questions:
 1) What changes the ctime actually? Does AWS or k8s use some weird magic when the PV is mounted?
 2) How about a test? I can "touch" the files to change their metadata in some random order, would that be OK?

JAORMX:
mtime does feel more appropriate than ctime. this LGTM.

JAORMX:
> Maybe this should be marked as WIP because I still can't answer two questions:
> 
>     1. What changes the ctime actually? Does AWS or k8s use some weird magic when the PV is mounted?

That's a good question :/ I don't really know.

> 
>     2. How about a test? I can "touch" the files to change their metadata in some random order, would that be OK?

That would be fine. Here we merely read a file in between https://github.com/openshift/compliance-operator/blob/master/cmd/manager/resultserver_test.go#L71 but touching a file would be better.

jhrozek:
FWIW, last test failure:
```
   * could not run steps: step [output:stable:testcontent] failed: could not resolve base image: imagestreamtags.image.openshift.io "pipeline:testcontent" not found
  * could not run steps: step [output:stable:compliance-operator] failed: could not resolve base image: imagestreamtags.image.openshift.io "pipeline:compliance-operator" not found 
```

PR #421, opened 2020-08-31
Only compare Machineconfig spec, not the metadata

JAORMX:
the mcDiffers was comparing the whole machineConfig. This is problematic
on updates, as the metadata (e.g. resourceVersion) will be differen this
scenario. What we really want to evaluate is the contents of the
MachineConfig, which is in the spec.


jhrozek (approved):
Oops, what a silly bug.
/lgtm

PR #422, opened 2020-08-31
Update base images of our container files

JAORMX:
They were pointing to older releases.


jhrozek (approved):
/lgtm
this matches what other OCP components are doing

PR #424, opened 2020-09-01
Normalize the usage of labels to identify components in the operator

JAORMX:
The operator creates several workloads in order to do what it needs;
To ease debugging, this commit normalizes the usage of the "workload"
label to identify the different pods, so they could be dynamically
fetched using standard kube utilities.


jhrozek (approved):
/lgtm

PR #425, opened 2020-09-01
Rename openscap-ocp container to just `scanner`

JAORMX:
This decouples the implementation details from the pod/container names
we're using. Also, it's easier to remember for someone troubleshooting.


jhrozek (approved):
The only remaining mentions of openscap-ocp in the repo are now the Dockerfile for the actual container running openscap
/lgtm

PR #426, opened 2020-09-01
MachineConfig: Use generation to check if we have an up-to-date remediation

JAORMX:
Our previous method of checking if machineconfig remediations were up to
date had a flaw: It was order dependant. This means that whenver the
merging of machine configs from remediation ended up in a different
order (e.g. fore files or units) we would have a differing machine
config, and so MCO would create a new rendering.

This would end up in an endless loop and thus the cluster would keep
restarting.

This changes the way we check for remediations being up-to-date, by
using the remediation's generation. If there is a change in the
remediation definition, then we would see a generation change. Else, we
assume that all is good and up to date.

### DIFF
--- a/pkg/controller/complianceremediation/complianceremediation_controller.go
+++ b/pkg/controller/complianceremediation/complianceremediation_controller.go
@@ -3,7 +3,6 @@ package complianceremediation
 import (
 	"context"
 	"fmt"
-	"reflect"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


jhrozek:
Great, this seems to have worked for me. The nodes went through multiple reboots, but eventually converged and the generation of the composite MC stayed the same once the pool was unpaused. The code also looks good to me.

JAORMX:
Pull-request updated, HEAD is now ad226fd1b29e6f2a5dbb1bae793e10a41f43df26

jhrozek (approved):
This also fixes the UnapplyTest
/lgtm

PR #432, opened 2020-09-07
readme: Document how to install packages

jhrozek:
We used to document how to install from source, but not how to install
the packages we release.


JAORMX (approved):
/lgtm

PR #434, opened 2020-09-08
Take OPM binary from specific release instead of the go module

JAORMX:
This way we ensure that the binary is installed, without importing all
its dependencies. Thus, this reduces the binary size and the amount of
dependencies we carry.


jhrozek (approved):
tested by running make opm and make index-image, both work
/lgtm

PR #436, opened 2020-09-09
Bug 1862453: Also delete jobs when deleting compliancesuites

JAORMX:
It was the case that we were deleting the pods and the cronjobs... but
the jobs were being left out. This fixes that.
    
Also, this reconciles the logic for cleanup to be only in one place.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -113,6 +113,12 @@ rules:
   - delete 
   - update
 - apiGroups:
+  - batch
+  resources:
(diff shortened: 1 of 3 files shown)


JAORMX:
Pull-request updated, HEAD is now 50347e4d0e3905c03e1ef15ef130a554c5a18b15

jhrozek (approved):
tested with a schedules scans in a suite, the job is gone after deleting the suite
/lgtm

PR #438, opened 2020-09-10
Bug 1877807: Support disconnected installs

JAORMX:
The current default behavior is to allow the `oscap` command to download
external resources. This is an issue in disconnected installs since
they'll either have no access to the internet, or have so, but through a
proxy.

This enables the usage of the operator in such environments by allowing
the admin to set two new scan settings:

* **noExternalResources**: This will entirely remove the flag that
  downloads external resources, and stuff like cve checks simply won't
  run.

* **httpsProxy**: This sets the `http_proxy` environment variable, so
  the `oscap` command will use the proxy. Note that the name
  `httpsProxy` was chosen to encourage folks to use https for their
  proxies.

### DIFF
--- a/deploy/crds/compliance.openshift.io_compliancescans_crd.yaml
+++ b/deploy/crds/compliance.openshift.io_compliancescans_crd.yaml
@@ -52,6 +52,17 @@ spec:
               debug:
                 description: Enable debug logging of workloads and OpenSCAP
                 type: boolean
+              httpsProxy:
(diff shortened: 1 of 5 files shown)


JAORMX:
Pull-request updated, HEAD is now 8d34ab1b30a4fd08edc9a71485a9809c06680034

jhrozek (approved):
The configMaps have the expected evironment variables created
/lgtm

PR #441, opened 2020-09-11
Migrate operator-sdk and kube libraries to use a v0.19.0

JAORMX:
OCP 4.6 is now using kube v0.19.0; in an attempt to synchronize with
that release, this migrates our modules to use that.

This also upgrades the MCO and OpenShift dependencies. This was needed
in order for us to be able to take advantage of the new ignition version.

Note that this doesn't migrate the operator-sdk binary we're using, as that
has extra steps we need to do before upgrading.


JAORMX:
Pull-request updated, HEAD is now c6e7d1cc11381613188600833ea0dab8e7c9f303

jhrozek:
Are you sure that the migration is as easy as bumping the version? https://v0-19-x.sdk.operatorframework.io/docs/golang/project_migration_guide/ seems rather involved..

jhrozek:
Ah, sorry, I missed what you wrote in the opening comment: "Note that this doesn't migrate the operator-sdk binary we're using"

jhrozek (approved):
/lgtm

JAORMX:
Pull-request updated, HEAD is now 7e98df325b9d5baa9f6950a8ed3816f4d2197b55

build/Dockerfile
```diff
@@ -1,11 +1,11 @@
 # Step one: build compliance-operator
-FROM registry.access.redhat.com/ubi8/go-toolset as builder
+FROM golang:1.14 AS builder
 
 WORKDIR /go/src/github.com/openshift/compliance-operator
 
 ENV GOFLAGS=-mod=vendor
 
-COPY --chown=default:root . .
+COPY . .
```
jhrozek, line 8: Why was the chown removed? IIRC it was required to build using docker?

JAORMX: with the base image changing, it started to fail.

JAORMX:
Pull-request updated, HEAD is now 1c7cb929cfa2e1ce46d824287fdf1ffa36d9a68d

build/Dockerfile
jhrozek: Nevermind, building with docker seems to be still working.

jhrozek:
I tested build with both podman and docker and manually tested a bunch of things we don't test well in e2e tests and looked briefly at the operator logs. All seems well (except I hit a known race once, but I'm sure that's unrelated to this patch)

/lgtm

PR #443, opened 2020-09-14
Stop using operator-sdk build to build operator image

JAORMX:
This command has been deprecated in newer operator-sdk releases, so
let's stop using it.


jhrozek (approved):
Works fine with both docker and podman
/lgtm

PR #446, opened 2020-09-16
Handle result events generically

JAORMX:
The suite controller currently issues events for results being
available. For RHACM support, we need to do this for scansettingbindings
too, since that's the object that'll be created by RHACM as well.

this change enables such events to be created from both controllers in a
general manner.

### DIFF
--- a/pkg/controller/common/events.go
+++ b/pkg/controller/common/events.go
@@ -1,9 +1,16 @@
 package common
 
 import (
+	"fmt"
+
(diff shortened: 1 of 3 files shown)


JAORMX:
Pull-request updated, HEAD is now 0512c06980d70629917678399b89c57015b7cb7c

jhrozek (approved):
/lgtm

PR #448, opened 2020-09-16
Fix logging issues in scansettingbindings controller

JAORMX:
One error message was wrong and we were using the general logger instead
of the one with values set.

This will aide in debugability.


jhrozek (approved):
/lgtm

PR #449, opened 2020-09-16
Change aggregator's container name to something more fitting

JAORMX:
The aggregator's main container was using the name log-collector. Most
likely as a copy-paste error. This turns out a little confusing when
reading the namespace's events. So let's use something more suitable.


jhrozek (approved):
/lgtm

PR #451, opened 2020-09-17
ProfileBundle: Add contentfile to server side printing parameters

JAORMX:
This would be useful as several profile bundles may use the same content
image. So this would help easily debug and identify issues.


jhrozek:
Cluster didn't install it seems.

jhrozek (approved):
/lgtm

PR #453, opened 2020-09-18
Fix race between the aggregator and operator in processing result configMaps

jhrozek:
These two patches address a race between creating and annotating the result CM by the aggregator pod on one hand and evaluating the CM by the operator on the other hand. The first patch also fixes a logic bug in evaluating the isReady flag.
More details are in the commit messages:

compliancescan: Requeue if the result CM doesn't have the required
annotations

We've seen cases where the results configMap exists, but does not have the
required annotations yet. Instead of returning a hard error in this case,
return the isReady flag set to false in gatherResults which would cause the
processing to be requeued again in a couple of seconds.

Jira: OCPBUGSM-17900

compliancescan: Don't keep checking the scan results if the CM is not ready

If the result ConfigMap was not created by the time we check the aggregated
results, we would previously just set the isReady flag, but continue. The
code would then proceed to the getScanResult function which tries to check
the CM's annotations and because the CM doesn't even exist, getScanResult
would have returned a failure. This means the isReady flag would never be
returned, but an error would be returned instead.

This patch just shortcuts processing of the current configMap and moves on
to the next one. This way, if there is an error in the other CM's, it would
have bubbled up, but otherwise the isReady flag would have been returned as 
expected.

Jira: OCPBUGSM-17900

### DIFF
--- a/pkg/controller/compliancescan/compliancescan_controller.go
+++ b/pkg/controller/compliancescan/compliancescan_controller.go
@@ -462,7 +462,7 @@ func (r *ReconcileComplianceScan) phaseAggregatingHandler(instance *compv1alpha1
 
 	// We only wait if there are no errors.
 	if err == nil && !isReady {
-		log.Info("ConfigMap missing (not ready). Requeuing.")
+		log.Info("ConfigMap missing or not ready. Requeuing.")
 		return reconcile.Result{Requeue: true, RequeueAfter: requeueAfterDefault}, nil
 	}
 
@@ -814,6 +814,13 @@ func gatherResults(r *ReconcileComplianceScan, instance *compv1alpha1.Compliance
 			break
 		}
 
+		cmHasResult := scanResultReady(foundCM)
+		if cmHasResult == false {
+			logger.Info("Scan results not ready, retrying. If the issue persists, restart or recreate the scan", "ComplianceScan.Name", instance.Name)
+			isReady = false
+			break
+		}
+
 		// NOTE: err is only set if there is an error in the scan run
 		result, err = getScanResult(foundCM)
 
@@ -836,6 +843,13 @@ func gatherResults(r *ReconcileComplianceScan, instance *compv1alpha1.Compliance
 			// error here.
 			if err != nil {
 				logger.Info("Node has no result ConfigMap yet", "node.Name", node.Name)
+				isReady = false
+				continue
+			}
+
+			cmHasResult := scanResultReady(foundCM)
+			if cmHasResult == false {
+				logger.Info("Scan results not ready, retrying. If the issue persists, restart or recreate the scan", "ComplianceScan.Name", instance.Name)
 				isReady = false
 				continue
 			}
--- a/pkg/controller/compliancescan/compliancescan_controller.go
+++ b/pkg/controller/compliancescan/compliancescan_controller.go
@@ -215,12 +215,12 @@ func (r *ReconcileComplianceScan) phasePendingHandler(instance *compv1alpha1.Com
 		var nodes corev1.NodeList
 		var err error
 		if nodes, err = getTargetNodes(r, instance); err != nil {
-			log.Error(err, "Cannot get nodes")
+			logger.Error(err, "Cannot get nodes")
 			return reconcile.Result{}, err
 		}
 		if len(nodes.Items) == 0 {
 			warning := "No nodes matched the nodeSelector"
-			log.Info(warning)
+			logger.Info(warning)
 			r.recorder.Event(instance, corev1.EventTypeWarning, "NoMatchingNodes", warning)
 			instanceCopy := instance.DeepCopy()
 			instanceCopy.Status.Result = compv1alpha1.ResultNotApplicable
@@ -258,22 +258,22 @@ func (r *ReconcileComplianceScan) phaseLaunchingHandler(instance *compv1alpha1.C
 	}
 
 	if nodes, err = getTargetNodes(r, instance); err != nil {
-		log.Error(err, "Cannot get nodes")
+		logger.Error(err, "Cannot get nodes")
 		return reconcile.Result{}, err
 	}
 
 	if err = r.handleRootCASecret(instance, logger); err != nil {
-		log.Error(err, "Cannot create CA secret")
+		logger.Error(err, "Cannot create CA secret")
 		return reconcile.Result{}, err
 	}
 
 	if err = r.handleResultServerSecret(instance, logger); err != nil {
-		log.Error(err, "Cannot create result server cert secret")
+		logger.Error(err, "Cannot create result server cert secret")
 		return reconcile.Result{}, err
 	}
 
 	if err = r.handleResultClientSecret(instance, logger); err != nil {
-		log.Error(err, "Cannot create result client cert secret")
+		logger.Error(err, "Cannot create result client cert secret")
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -462,7 +462,7 @@ func (r *ReconcileComplianceScan) phaseAggregatingHandler(instance *compv1alpha1
 
 	// We only wait if there are no errors.
 	if err == nil && !isReady {
-		log.Info("ConfigMap missing (not ready). Requeuing.")
+		log.Info("ConfigMap missing or not ready. Requeuing.")
```
JAORMX, line 465: let's use the "logger" object instead of "log"

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -809,7 +809,16 @@ func gatherResults(r *ReconcileComplianceScan, instance *compv1alpha1.Compliance
 		// Could be a transient error, so we requeue if there's any
 		// error here.
 		if err != nil {
+			logger.Info("Platform scan has no result CM yet", "ComplianceScan.Name", instance.Name)
```
JAORMX, line 812: s/CM/ConfigMap/g

Makes it more intuitive to grep

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -809,7 +809,16 @@ func gatherResults(r *ReconcileComplianceScan, instance *compv1alpha1.Compliance
...
 			isReady = false
+			break
+		}
+
+		cmHasResult := scanResultReady(foundCM)
+		if cmHasResult == false {
+			logger.Info("Platform scan CM is not ready yet", "ComplianceScan.Name", instance.Name)
```
JAORMX, line 819: s/CM/ConfigMap/g

pkg/controller/compliancescan/scan.go
```diff
@@ -645,16 +645,21 @@ func checkScanUnknownError(cm *corev1.ConfigMap) error {
...
 
-func getScanResult(cm *corev1.ConfigMap) (compv1alpha1.ComplianceScanStatusResult, error) {
+func scanResultReady(cm *corev1.ConfigMap) bool {
 	if cm.Annotations == nil {
-		return compv1alpha1.ResultError, fmt.Errorf("the ConfigMap '%s' was missing annotations", cm.Name)
+		return false
 	}
 
-	strResult, ok := cm.Annotations[compv1alpha1.CmScanResultAnnotation]
+	_, ok := cm.Annotations[compv1alpha1.CmScanResultAnnotation]
```
JAORMX, line 653: If this won't be an error anymore, perhaps we want to log, at least.

In the case where a user would manually modify a ConfigMap, and remove the annotation, the scan would never get to ready state.

JAORMX (changes requested):

jhrozek:
On Sun, Sep 20, 2020 at 10:27:23PM -0700, Juan Osorio Robles wrote:
> @JAORMX requested changes on this pull request.
> 
> 
> 
> > @@ -462,7 +462,7 @@ func (r *ReconcileComplianceScan) phaseAggregatingHandler(instance *compv1alpha1
>  
>  	// We only wait if there are no errors.
>  	if err == nil && !isReady {
> -		log.Info("ConfigMap missing (not ready). Requeuing.")
> +		log.Info("ConfigMap missing or not ready. Requeuing.")
> 
> let's use the "logger" object instead of "log"
> 
> > @@ -809,7 +809,16 @@ func gatherResults(r *ReconcileComplianceScan, instance *compv1alpha1.Compliance
>  		// Could be a transient error, so we requeue if there's any
>  		// error here.
>  		if err != nil {
> +			logger.Info("Platform scan has no result CM yet", "ComplianceScan.Name", instance.Name)
> 
> s/CM/ConfigMap/g
> 
> Makes it more intuitive to grep
> 
> >  			isReady = false
> +			break
> +		}
> +
> +		cmHasResult := scanResultReady(foundCM)
> +		if cmHasResult == false {
> +			logger.Info("Platform scan CM is not ready yet", "ComplianceScan.Name", instance.Name)
> 
> s/CM/ConfigMap/g
> 
> >  	}
>  
> -	strResult, ok := cm.Annotations[compv1alpha1.CmScanResultAnnotation]
> +	_, ok := cm.Annotations[compv1alpha1.CmScanResultAnnotation]
> 
> If this won't be an error anymore, perhaps we want to log, at least.
> 
> In the case where a user would manually modify a ConfigMap, and remove the annotation, the scan would never get to ready state.

Yeah, I was thinking about this as well. Is there a general pattern in
Kubernetes of reconcilng a state only a limited number of times and then
giving up? We could of course annotate the CM with some internal
annotation and then give up once it's past a threshold or even just keep
a in-memory map and drop it once the CM has been processed (that
wouldn't survive operator restarts but meh). But I'm not sure if there's
anything that goes above the usual 'let's reconcile until the requested state
is reached...'

JAORMX:
> On Sun, Sep 20, 2020 at 10:27:23PM -0700, Juan Osorio Robles wrote: @JAORMX requested changes on this pull request. > @@ -462,7 +462,7 @@ func (r *ReconcileComplianceScan) phaseAggregatingHandler(instance *compv1alpha1 // We only wait if there are no errors. if err == nil && !isReady { - log.Info("ConfigMap missing (not ready). Requeuing.") + log.Info("ConfigMap missing or not ready. Requeuing.") let's use the "logger" object instead of "log" > @@ -809,7 +809,16 @@ func gatherResults(r *ReconcileComplianceScan, instance *compv1alpha1.Compliance // Could be a transient error, so we requeue if there's any // error here. if err != nil { + logger.Info("Platform scan has no result CM yet", "ComplianceScan.Name", instance.Name) s/CM/ConfigMap/g Makes it more intuitive to grep > isReady = false + break + } + + cmHasResult := scanResultReady(foundCM) + if cmHasResult == false { + logger.Info("Platform scan CM is not ready yet", "ComplianceScan.Name", instance.Name) s/CM/ConfigMap/g > } - strResult, ok := cm.Annotations[compv1alpha1.CmScanResultAnnotation] + _, ok := cm.Annotations[compv1alpha1.CmScanResultAnnotation] If this won't be an error anymore, perhaps we want to log, at least. In the case where a user would manually modify a ConfigMap, and remove the annotation, the scan would never get to ready state.
> Yeah, I was thinking about this as well. Is there a general pattern in Kubernetes of reconcilng a state only a limited number of times and then giving up? We could of course annotate the CM with some internal annotation and then give up once it's past a threshold or even just keep a in-memory map and drop it once the CM has been processed (that wouldn't survive operator restarts but meh). But I'm not sure if there's anything that goes above the usual 'let's reconcile until the requested state is reached...'

Trying to reconcile is the only thing to do. So the logic is fine IMO. However, at least we should issue a log message so admins are able to identify the issue.

pkg/controller/compliancescan/scan.go
jhrozek: I've actually reverted to the old behaviour. We now have a guard before we call `getScanResult` function to check if the CM is ready, so calling the `getScanResult` function with invalid input should throw an error as you say.

jhrozek: Hmm, wait up, I know what you meant, let me add one more logging call.

jhrozek: Actually, the call to `scanResultReady` already logs something, hope it's sufficient:
``+                       cmHasResult := scanResultReady(foundCM)                                                                                                                                                                              
+                       if cmHasResult == false {                                                                                                                                                                                            
+                               logger.Info("Node ConfigMap is not ready yet", "node.Name", node.Name)                                                                                                                                       
+                               isReady = false                                                                                                                                                                                              
+                               continue                                                                                                                                                                                                     
+                       }                                                                                                                                                                                                                    
+                         `

```

JAORMX: Right, it already looks that the node is not ready. But you don't know why. And we have no docs about it or relevant logs.

jhrozek: A new log message was added per suggestion on Slack.

JAORMX (approved):
/lgtm

Issue #461, opened 2020-10-20
multi-namespace support is broken for profilebundles


JAORMX:
ProfileBundle objects create a deployment that will parse the data stream in the container. The deployment is named after the profilebundle.

e.g. if the profilebundle is named `ocp4`, the deployment will be named `ocp4-pp`.

While this works when the operator is only listening on one namespace, this breaks if we try to make it listen on several.

We should take the namespace into account when creating these deployments.

PR #462, opened 2020-10-20
Take namespace into account for profilebundle deployment

JAORMX:
This adds the name of the namespace for creating the profilebundle
deployment object. Thus allowing us to listen on several namespaces.

Closes-Bug: #461


jhrozek (approved):
/lgtm

PR #463, opened 2020-10-21
Hardening: Set some general hardening settings for our containers

JAORMX:
This sets up some general hardening settings such as:

* Disallowing privilege escalation
* Read-only container filesystem

It also adds a read-only flag for a mount where needed.

This also mounts `/tmp` as an empty dir for the scanner containers. This
is because OpenSCAP will attempt to create a directory in `/tmp` in
order to temporarily dump the OVAL results. This way, we can mount the
directory and write to it, while still keeping the rest of the container
in a read-only filesystem.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -20,6 +20,9 @@ spec:
           - compliance-operator
           - operator
           imagePullPolicy: Always
+          securityContext:
+            readOnlyRootFilesystem: true
+            allowPrivilegeEscalation: false
           env:
             - name: WATCH_NAMESPACE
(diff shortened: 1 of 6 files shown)


JAORMX:
Pull-request updated, HEAD is now 28a3ddce730f799579ad61d242582ae8d363c352

JAORMX:
Pull-request updated, HEAD is now 870b60acd054e1d5e4aa06100ff8279ad9c85383

jhrozek (approved):
/lgtm
nice work, I had no idea that you could override directories like this.

PR #469, opened 2020-11-12
Add support for remediation patching

JAORMX:
This enables us to create partial remediations, these will merely patch
the objects that already exist in kube and ensure that they're
compliant.

If an object is created from scratch, it'll annotate the object saying
the operator created it.

If an object already exists on the cluster, it is patched.

When unapplying remediations, the operator will only delete objects that
were created by the operator to begin with. However, we currently won't
be able to roll back remediations that patched objects.

On the other hand, this patch makes sure to handle authorization issues with
remediations by surfacing any RBAC issues we encounter when
applying/unapplying remediations.
    
This handles cases where we don't have permission to get/create/patch a
certain object, and it should reflect it in the remediation's status

### DIFF
--- a/pkg/controller/complianceremediation/complianceremediation_controller.go
+++ b/pkg/controller/complianceremediation/complianceremediation_controller.go
@@ -134,32 +134,90 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
 // remediation if applicable
 func (r *ReconcileComplianceRemediation) reconcileGenericRemediation(instance *compv1alpha1.ComplianceRemediation, logger logr.Logger) error {
 	obj := instance.Spec.Current.Object
+	objectLogger := logger.WithValues("Object.Name", obj.GetName(), "Object.Namespace", obj.GetNamespace())
+
+	objectLogger.Info("Reconciling remediation object")
+
 	found := obj.DeepCopy()
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, found)
-	if instance.Spec.Apply {
-		logger.Info("Remediation will be applied", "ComplianceRemediation.Name", instance.Name)
-		if errors.IsNotFound(err) {
-			return r.client.Create(context.TODO(), obj)
-		} else if meta.IsNoMatchError(err) {
-			// If the kind is not available in the cluster, we can't retry
-			return common.NewNonRetriableCtrlError("Unable to create ComplianceRemediation. Make sure the CRD is installed: %s", err)
-		}
-		// TODO(jaosorior): If the object is found, should we update it?
-		return err
-	} else {
-		logger.Info("Remediation will be unapplied", "ComplianceRemediation.Name", instance.Name)
-	}
 
-	// unapply remediation
-	if err == nil {
-		return r.client.Delete(context.TODO(), obj)
-	} else if errors.IsNotFound(err) {
-		return nil
+	if errors.IsForbidden(err) {
+		return common.NewNonRetriableCtrlError(
+			"Unable to get fix object from ComplianceRemediation. "+
+				"Please update the compliance-operator's permissions: %s", err)
 	} else if meta.IsNoMatchError(err) {
 		// If the kind is not available in the cluster, we can't retry
-		return common.NewNonRetriableCtrlError("Unable to use ComplianceRemediation. Make sure the CRD is installed: %s", err)
+		return common.NewNonRetriableCtrlError(
+			"Unable to get fix object for ComplianceRemediation. "+
+				"Make sure the CRD is installed: %s", err)
+	} else if errors.IsNotFound(err) {
+		if instance.Spec.Apply {
+			return r.createGenericRemediation(obj, found, objectLogger)
+		}
+
+		objectLogger.Info("The object wasn't found, so no action is needed to unapply it")
+		return nil
+	} else if err != nil {
+		return err
 	}
-	return err
+
+	if instance.Spec.Apply {
+		return r.patchGenericRemediation(obj, found, objectLogger)
+	}
+
+	return r.deleteGenericRemediation(obj, found, objectLogger)
+}
+
+func (r *ReconcileComplianceRemediation) createGenericRemediation(remObj *unstructured.Unstructured, foundObj *unstructured.Unstructured, logger logr.Logger) error {
+	logger.Info("Remediation will be created")
+	compv1alpha1.AddRemediationAnnotation(remObj)
+
+	createErr := r.client.Create(context.TODO(), remObj)
+
+	if errors.IsForbidden(createErr) {
+		// If the kind is not available in the cluster, we can't retry
+		return common.NewNonRetriableCtrlError(
+			"Unable to create fix object from ComplianceRemediation. "+
+				" Please update the compliance-operator's permissions: %s", createErr)
+	}
+
+	return createErr
+}
+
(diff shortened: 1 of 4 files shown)


JAORMX:
Pull-request updated, HEAD is now 31ccc5fb0ea381a0c81b000c7e76cefabfa91697

JAORMX:
Pull-request updated, HEAD is now 24305d7c64926142b1f19240a34298f69f2ffa44

JAORMX:
Pull-request updated, HEAD is now 259aad58e942ded1f762244c99193d612c069b00

JAORMX:
Pull-request updated, HEAD is now ca4633e0d9e5f43b80f4f7d6c6fa114069ece416

JAORMX:
Pull-request updated, HEAD is now 6c3d2ce18fc5065f0af888f3488a9674a62ed8d1

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -134,32 +134,105 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
...
-		if errors.IsNotFound(err) {
-			return r.client.Create(context.TODO(), obj)
-		} else if meta.IsNoMatchError(err) {
-			// If the kind is not available in the cluster, we can't retry
-			return common.NewNonRetriableCtrlError("Unable to create ComplianceRemediation. Make sure the CRD is installed: %s", err)
+
+	if errors.IsForbidden(err) {
+		return common.NewNonRetriableCtrlError(
+			"Unable to get fix object from ComplianceRemediation. "+
+				"Please update the compliance-operator's permissions: %s", err)
```
jhrozek, line 147: The indentation seems off here.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -134,32 +134,105 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
...
+	} else if meta.IsNoMatchError(err) {
+		// If the kind is not available in the cluster, we can't retry
+		return common.NewNonRetriableCtrlError(
+			"Unable to get fix object for ComplianceRemediation. "+
```
jhrozek, line 151: Indentation

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -134,32 +134,105 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
...
+	logger.Info("Remediation will be created")
+	compv1alpha1.AddRemediationAnnotation(remObj)
+
+	createErr := r.client.Create(context.TODO(), remObj)
+
+	if errors.IsForbidden(createErr) {
+		// If the kind is not available in the cluster, we can't retry
+		return common.NewNonRetriableCtrlError(
+			"Unable to create fix object from ComplianceRemediation. "+
+				" Please update the compliance-operator's permissions: %s", createErr)
```
jhrozek, line 181: Indentation

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -134,32 +134,105 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
...
+	} else if errors.IsAlreadyExists(createErr) {
```
jhrozek, line 182: If we were asked to create an object but it exists, isn't it better to return a retriable error and let the loop cycle back into the patch branch?

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -134,32 +134,105 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
...
-	} else if meta.IsNoMatchError(err) {
+	}
+
+	return createErr
+}
+
+func (r *ReconcileComplianceRemediation) doPatchGenericRemediation(remObj *unstructured.Unstructured, foundObj *unstructured.Unstructured, logger logr.Logger) error {
+	logger.Info("Remediation patch object")
+	// ensure labels are set so we don't overwrite them
+	labels := foundObj.GetLabels()
```
jhrozek, line 192: I'm sorry, I don't understand this branch. Why are we setting the labels on the found object to an empty map? It doesn't seem like we're using the found object further on in this function or in the caller where found is scoped? Shouldn't the merge strategy take care about merging the labels and annotations as well?

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -134,32 +134,105 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
...
+		foundObj.SetLabels(annotations)
+	}
+
+	patchErr := r.client.Patch(context.TODO(), remObj, client.Merge)
+
+	if errors.IsForbidden(patchErr) {
 		// If the kind is not available in the cluster, we can't retry
-		return common.NewNonRetriableCtrlError("Unable to use ComplianceRemediation. Make sure the CRD is installed: %s", err)
+		return common.NewNonRetriableCtrlError(
+			"Unable to patch fix object from ComplianceRemediation. "+
```
jhrozek, line 210: indentation

pkg/controller/complianceremediation/complianceremediation_controller.go
jhrozek: Either way it might be nice to add a test about updating the labels, that would make the case clearer.

jhrozek (commented):
Nice work, conceptually looks OK to be, left a couple of small nits inline.

pkg/controller/complianceremediation/complianceremediation_controller.go
JAORMX: that's how it was auto-idented :/

pkg/controller/complianceremediation/complianceremediation_controller.go
JAORMX: that's a good point.

pkg/controller/complianceremediation/complianceremediation_controller.go
JAORMX: I was having issues because the labels and annotations were unset with the PATCH... but maybe it's not an issue now since I switched to the Merge strategy.

JAORMX:
Pull-request updated, HEAD is now 72f8377011b063bdbf9ec296a14c9be67dcfb55b

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -134,32 +134,92 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
...
+				"Please update the compliance-operator's permissions: %s", patchErr)
 	}
-	return err
+
+	return patchErr
+
+}
+
+func (r *ReconcileComplianceRemediation) doDeleteGenericRemediation(remObj *unstructured.Unstructured, foundObj *unstructured.Unstructured, logger logr.Logger) error {
```
mrogers950, line 205: nit: not crazy about the "do" in the function names; could just as well be deleteGenericRemediation and we assume that's what it is doing :)

JAORMX:
Pull-request updated, HEAD is now 35f29f747206231ba2c0a526ae75af998b00e2bd

JAORMX:
Pull-request updated, HEAD is now ef69ea59aef344664e73652ab7d2987eb43b8748

JAORMX:
Pull-request updated, HEAD is now ef69ea59aef344664e73652ab7d2987eb43b8748

JAORMX:
Pull-request updated, HEAD is now dd0f388c3d1f6fd7b54baf4f14020e940c788227

jhrozek (approved):
/lgtm

PR #470, opened 2020-11-16
Expose API resource collector errors as warnings

JAORMX:
This exposes errors from the API resource collector by posting them to
a warnings file.

This will be read by the collector, uploaded to a ConfigMap, and
subsequently persisted to the Custom Resource.

The errors that we expose are:

* RBAC issues (Forbidden)
* Kind/Resource issues (if they don't exist)

This way we can indicate to users that they need to revisit their
permissions or install an appropriate operator.


JAORMX:
Pull-request updated, HEAD is now 04bc0d3115bd6ed8f88d5c224d1dc0a9fb7ac0bd

JAORMX:
Pull-request updated, HEAD is now d4c26403ceef2615ddf3b64aa3d9424c9b0aaafb

JAORMX:
Pull-request updated, HEAD is now 0dc8d19e685888fe0da3bc9510b4c247eb5fe166

JAORMX:
Pull-request updated, HEAD is now 5123369c3dddbf4423c0ccb6f1d9d7253548e593

JAORMX:
Pull-request updated, HEAD is now 117a560504b00e7dc7c49156380e77b35247bbdf

JAORMX:
Pull-request updated, HEAD is now bcfb7fd10e8acb41abd5d0cffff61639655f1ad9

JAORMX:
/hold 
revisiting this entirely

JAORMX:
Pull-request updated, HEAD is now a56a1407c8068c2def791eda7fcc198662cd26c5

JAORMX:
Pull-request updated, HEAD is now 09128a67e66a08c4996f66aff3d2384a47e9ceea

JAORMX:
Pull-request updated, HEAD is now e2caf33c12ae5ac06a4a5521b766841b01bdc055

cmd/manager/resultcollector.go
```diff
@@ -385,7 +414,7 @@ func getMutualHttpsTransport(c *scapresultsConfig) (*http.Transport, error) {
 // an exit code of 2 means that the scan returned non-compliant
 // an exit code of 1 means that the scan encountered an error
 func exitCodeIsError(exitcode string) bool {
-	return exitcode != "0" && exitcode != "2"
+	return exitcode != common.OpenSCAPExitCodeCompliant && exitcode != common.OpenSCAPExitCodeNonCompliant
```
mrogers950, line 417: Why not just `return exitcode == common.OpenSCAPExitCodeError` ?

cmd/manager/scap.go
```diff
@@ -213,23 +215,28 @@ func getResourcePaths(ds *utils.XMLDocument, profile string) []string {
 	return out
 }
 
-func (c *scapContentDataStream) FetchResources() error {
-	found, err := fetch(c.client, c.resources)
+func (c *scapContentDataStream) FetchResources() ([]string, error) {
```
mrogers950, line 218: You should also need to update the FetchResources interface (surprised that build did not catch this)

mrogers950:
@JAORMX do you think we should have a mention of these warnings in the README? Or would this be mostly for rare debugging?

cmd/manager/resultcollector.go
JAORMX: Because I was planning to re-use the error output for something else. The initial patch was actually using an exit error code 3. I still think that being able to re-use that will be useful in the future.

cmd/manager/scap.go
JAORMX: Good catch, let me update it.

JAORMX:
> @JAORMX do you think we should have a mention of these warnings in the README? Or would this be mostly for rare debugging?

I'll add it to our docs

JAORMX:
Pull-request updated, HEAD is now 57934a6e0d44df703cc34621dc6fb385af8fd5b7

JAORMX:
Pull-request updated, HEAD is now 3ac06446c16ff3e143b59eed99d8f00782031b95

PR #473, opened 2020-11-18
Bug 1898819: Aggregator: Only parse ConfigMaps with results

JAORMX:
The aggregatorw as parsing all configMaps relevant to the scan. But
adding the relevant result label to the label selector that the pod uses
to fetch the configmaps, we can reduce the number of configmaps is
fetches and:

* Have cleaner aggregator logs
* Have a (slightly) more efficient aggregator


jhrozek (approved):
/lgtm

PR #474, opened 2020-11-18
Limit parallelism in our go test invocation

JAORMX:
This will limit the amount of processes to run parallel tests.


JAORMX:
Pull-request updated, HEAD is now a97ca526e8c25747c6f6bc950e93444af5d06276

jhrozek:
I'm reopening this PR because I think you were on to something. In PR #484 in the first CI run (1329722517694189568) the nodes were hitting OOM, I'm guessing because of messages like:
```
System OOM encountered, victim process: oscap, pid: 68865
```
this does not guarantee that it was openscap who was using a lot of memory, but it's very likely because the OOM killer terminates processes that will free a lot of memory...

Let's see how the test goes now, we might need to also include the patch from 484 to keep both fixes in the same PR perhaps..

jhrozek (approved):
/lgtm

PR #478, opened 2020-11-19
Add support for remediation patching (4.6 backport)

JAORMX:
This enables us to create partial remediations, these will merely patch
the objects that already exist in kube and ensure that they're
compliant.

If an object is created from scratch, it'll annotate the object saying
the operator created it.

If an object already exists on the cluster, it is patched.

When unapplying remediations, the operator will only delete objects that
were created by the operator to begin with. However, we currently won't
be able to roll back remediations that patched objects.

On the other hand, this patch makes sure to handle authorization issues with
remediations by surfacing any RBAC issues we encounter when
applying/unapplying remediations.

This handles cases where we don't have permission to get/create/patch a
certain object, and it should reflect it in the remediation's status.

### DIFF
--- a/pkg/apis/compliance/v1alpha1/complianceremediation_types.go
+++ b/pkg/apis/compliance/v1alpha1/complianceremediation_types.go
@@ -27,7 +27,8 @@ const (
 
 const (
 	// OutdatedRemediationLabel specifies that the remediation has been superseded by a newer version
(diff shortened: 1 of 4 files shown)


jhrozek:
Why do we need this in 4.6? I guess there would have to be a BZ opened?

JAORMX:
/hold

@jhrozek good point... I'll defer this until needed. (until we need this in the content...) I'll open a BZ then.

PR #480, opened 2020-11-19
E2E: Only log containers explicitly if an env var is set

jhrozek:
These days the OCP E2E logging seems to be logging containers on its
own. In an effort to reduce the number of API calls, let's only log the
container output ourselves if an environment variable is set explicitly.

### DIFF
--- a/tests/e2e/helpers.go
+++ b/tests/e2e/helpers.go
@@ -43,6 +43,7 @@ import (
 )
 
 var contentImagePath string
+var shouldLogContainerOutput bool
 
 var rhcosPb *compv1alpha1.ProfileBundle
 var ocp4Pb *compv1alpha1.ProfileBundle
@@ -58,6 +59,11 @@ func init() {
 	if contentImagePath == "" {
 		fmt.Println("Please set the 'CONTENT_IMAGE' environment variable")
 		os.Exit(1)
+	}
+
+	logContainerOutputEnv := os.Getenv("LOG_CONTAINER_OUTPUT")
+	if logContainerOutputEnv != "" {
+		shouldLogContainerOutput = true
 	}
 }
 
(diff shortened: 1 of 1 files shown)


JAORMX (approved):

jhrozek:
Note for tomorrow since it's getting late - the parallel tests that are failing start the aggregator pod, but then cycle in an endless loop of:
```
{"level":"info","ts":1605793687.0031588,"logger":"scanctrl","msg":"Reconciling ComplianceScan","Request.Namespace":"osdk-e2e-dc5e893d-80ff-4e0c-9dba-b5fc3a3524b9","Request.Name":"test-filtered-scan"}
{"level":"info","ts":1605793687.0032804,"logger":"scanctrl","msg":"Phase: Aggregating","Request.Namespace":"osdk-e2e-dc5e893d-80ff-4e0c-9dba-b5fc3a3524b9","Request.Name":"test-filtered-scan"}
{"level":"info","ts":1605793687.0037715,"logger":"scanctrl","msg":"ConfigMap missing (not ready). Requeuing.","Request.Namespace":"osdk-e2e-dc5e893d-80ff-4e0c-9dba-b5fc3a3524b9","Request.Name":"test-filtered-scan"}
```
The error seems to be coming from `phaseAggregatingHandler` where `shouldLaunchAggregator` returns `isReady=False` and `err == nil`.  Since this is a node scan, looks like we are getting an error from `getNodeScanCM` which in turn has issues looking up the result config map, because the error comes from `err := r.client.Get(context.TODO(), targetCM, foundCM)`.

In the confiMap list I only see two CMs for that test, one for `ip-10-0-136-33.us-east-2.compute.internal` and one for `ip-10-0-200-200.us-east-2.compute.internal`. I do not see a CM for `ip-10-0-187-240` at all.

I will continue digging tomorrow to see why was that CM not created.

PR #482, opened 2020-11-19
e2e: A bunch of minor code cleanups

jhrozek:
It seems like GoLand update makes some warnings more prominent since a
recent upgrade, so fixing them makes my editor a little less colorful.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -96,7 +96,7 @@ func TestE2E(t *testing.T) {
 					return err
 				}
 
-				if err := assertMustHaveParsedRules(t, f, namespace, pbName); err != nil {
+				if err := assertMustHaveParsedRules(f, pbName); err != nil {
 					return err
 				}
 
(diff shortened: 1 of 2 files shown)


jhrozek:
According to the log messages it seems like the cluster had been having some issues:
```
{"level":"info","ts":1605803360.0400896,"logger":"scanctrl","msg":"Node scan error","Request.Namespace":"osdk-e2e-1f300c44-2ee6-492b-8665-baf8eaac21b7","Request.Name":"test-suite-two-scans-with-platform-workers-scan","node.Name":"ip-10-0-237-10.us-east-2.compute.internal","errMsg":"Couldn't schedule scan pod 'openscap-pod-0580bf29d2b7533c1af7d23d813b2e3289ea158a': 0/6 nodes are available: 1 node(s) had taint {node.kubernetes.io/unreachable: }, that the pod didn't tolerate, 5 node(s) didn't match node selector."}
```
also the event log in must-gather doesn't sound too healthy, so I'm just going to kick the CI job..

PR #487, opened 2020-11-23
Fixes necessary for etcd encryption remediation

JAORMX:
This PR ads the necessary fixes for us to be able to write an automated
remediationf or the encryption provider.

This fixes:

* The aggregator, as it now can fully apply remediations other than
  machineconfigs even if no nodeSelector was used.
* The operator's clusterrole; since it now has permissions to get and modify
  the apiservers object called `cluster`


jhrozek (approved):
/lgtm

PR #488, opened 2020-11-23
Fix getting openshift user when it's something else than kubeadmin

JAORMX:
It was broken since we needed the `shell` command for Make.


jhrozek (approved):
/lgtm

PR #489, opened 2020-11-24
logcollector: Rely on the Reader interface instead of opening the files

JAORMX:
This relies on the Reader interface for all file operations in the
logcollector, instead of always reading the file and iterating through
it.

### DIFF
--- a/cmd/manager/resultcollector.go
+++ b/cmd/manager/resultcollector.go
@@ -16,12 +16,14 @@ limitations under the License.
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
@@ -178,11 +180,11 @@ func waitForResultsFile(filename string, timeout int64) *os.File {
 	return nil
 }
 
-func resultNeedsCompression(contents []byte) bool {
-	return len(contents) > 1048570
+func resultNeedsCompression(size int64) bool {
+	return size > 1048570
 }
 
-func compressResults(contents []byte) ([]byte, error) {
+func compressResults(contents io.Reader) (io.Reader, error) {
 	// Encode the contents ascii, compress it with gzip, b64encode it so it
 	// can be stored in the configmap.
 	var buffer bytes.Buffer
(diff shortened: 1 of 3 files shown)


JAORMX:
Pull-request updated, HEAD is now 3a641148effdfc604d5f365826683abeacb92d7a

JAORMX:
Pull-request updated, HEAD is now 44dcf3b037e825d310ad19cc57e03546eb717734

JAORMX:
Pull-request updated, HEAD is now 24193c2c8d0e72eb9ebfcf4d49dbcf292de97144

cmd/manager/resultcollector.go
```diff
@@ -178,52 +180,72 @@ func waitForResultsFile(filename string, timeout int64) *os.File {
...
 		return nil, err
 	}
-	w.Write([]byte(contents))
-	w.Close()
-	return buffer.Bytes(), nil
+	_, err = io.Copy(w, contents)
+	if err != nil {
+		return nil, err
+	}
+	defer w.Close()
```
jhrozek, line 199: This reads as if copy fails, then we never call the defer and never close the writer..

cmd/manager/resultcollector.go
```diff
@@ -178,52 +180,72 @@ func waitForResultsFile(filename string, timeout int64) *os.File {
...
 			log.Error(err, "Error: Compression failed")
 			return nil, err
 		}
 		rfContents.compressed = true
-		log.Info("Compressed results bytes size", "bytes", len(rfContents.contents))
+		log.Info("Compressed results")
+	} else {
+		rfContents.close = func() error {
+			return contentsfile.Close()
+		}
```
jhrozek, line 247: Nice trick!

cmd/manager/resultcollector.go
JAORMX: Good catch. Fixed

JAORMX:
Pull-request updated, HEAD is now e3227f4a74cff70e896e7ccff65282cdea202ce7

JAORMX:
Pull-request updated, HEAD is now 60b083f66e4b8337aeca95d7a3f864b32c1f7342

JAORMX:
Pull-request updated, HEAD is now b6feaf89cf756373b501d18f334ccb8344524bfd

jhrozek (commented):
Sorry I forgot to submit the review yesterday, here is one question

jhrozek:
btw the memory usage of the pod seems to have improved quite a bit looking at oc adm top pods!

pkg/utils/resultconfigmap.go
```diff
@@ -10,12 +12,25 @@ import (
...
+		_, err := io.Copy(enc, src)
+		enc.Close()
+
+		if err != nil {
+			pw.CloseWithError(err)
+		} else {
+			pw.Close()
+		}
+	}()
+	out, _ := ioutil.ReadAll(pr)
```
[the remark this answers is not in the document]
JAORMX: @jhrozek instead of reading the whole content into memory, then encoding it (which will read it again). This sets up the encoder to pass this directly to a pipe. And so we encode, then just read from the pipe. The go routine is required for working with the pipe; else it blocks.

jhrozek (approved):
/lgtm

PR #490, opened 2020-11-24
make: Don't use cache when building the openscap image

jhrozek:
We seldom rebuild the openscap image and when we do, we usually change
the contents of one of the repos that the image references, not the
image itself. Therefore it's safer even if a bit slower to use
--no-cache for the build and make sure the new packages are always
pulled from the repository.


JAORMX (approved):
/lgtm

PR #498, opened 2020-11-26
Bug 1901760: Don't fail the scansettingbinding if the same OS product is used

JAORMX:
The ScanSettingBinding controller shouldn't create a ComplianceSuite if it
detects profiles for different product. With the current logic, it would
also refuse to create a Suite if it detected more than one instance of
the same OS product. This PR fixes that.


JAORMX:
Pull-request updated, HEAD is now 2b7cf56c4d869efb89ffcc56fe81d22a56977196

jhrozek (approved):
Just marking as approved code-wise becaue @xiaojiey would be testing the functionality

JAORMX:
Pull-request updated, HEAD is now 03aed1a24e98dd150729140e08dd5a7cbf52b327

jhrozek (approved):
/lgtm

PR #499, opened 2020-11-26
Allow the api-resource-collector to get clusterlogforwarder objects

JAORMX:
This is needed to check if log forwarding is enabled.


jhrozek (approved):
/lgtm

PR #502, opened 2020-11-26
profileparser: Use a parallel worker object type

JAORMX:
This creates three go-routines, one per type of object (Profile, rule
and variables). This will en sure that this parsing is ran in parallel.


jhrozek (approved):
/lgtm

PR #503, opened 2020-11-26
e2e: Disable parallel testing

JAORMX:
This removes the e2e parallel tests in favor of a serial approach.

This is motivated by the fact that we're dealing with resource issues
and the kubelet de-scheduling pods or waiting too long to do so.

In face of this, this PR also increases the overall test timeout.


jhrozek (approved):
/lgtm
let's not waste more time debugging this

PR #504, opened 2020-11-26
e2e: Make tests that run moderate profile serial

JAORMX:
The moderate suite is very resource intensive, and so, let's use serial
tests for test-cases that use that.


jhrozek (approved):
/lgtm

PR #506, opened 2020-11-26
Use just bash-wait instead of calling pauser

jhrozek:
Submitting to see how do the e2e tests go. Still need to remove the old pauser and re-add the echo.


jhrozek:
Not anymore a WIP after a mere 3 months.

JAORMX (commented):
/lgtm

PR #507, opened 2020-11-27
Remove the replaces attribute from the CSV

jhrozek:
This seems to be causing issues during the bundle import


JAORMX (commented):
This will break upstream releases

PR #509, opened 2020-11-27
Allow api-resource-collector to view api-priority related objects

JAORMX:
A couple of objects need to be added to the api-resource-collector's
cluster-role in order to be able to scan the api-priority related
entries.


jhrozek (approved):
/lgtm
tested with a modified rule `api_server_api_priority_flowschema_catch_all` in CaC

PR #511, opened 2020-11-27
Bug 1902249: Fix tailored Platform scans

jhrozek:
We had code duplication between two functions that were setting up the
environment variables for the scans. When adding the tailoring support,
we forgot to add it to the function that sets up environment for the
platform scans.

This patch removes the code duplication and makes sure that the expected
environment is set up for both node and platform scans.

It also fixes the api-resource-collector to take into account when tailoring is required
and fetch the appropriate resources.

Closes: https://github.com/openshift/compliance-operator/pull/511

### DIFF
--- a/pkg/controller/compliancescan/config.go
+++ b/pkg/controller/compliancescan/config.go
@@ -215,7 +215,7 @@ func platformOpenScapScriptCm(name string, scan *compv1alpha1.ComplianceScan) *c
 	}
 }
 
-func defaultOpenScapEnvCm(name string, scan *compv1alpha1.ComplianceScan) *corev1.ConfigMap {
+func commonOpenScapEnvCm(name string, scan *compv1alpha1.ComplianceScan) *corev1.ConfigMap {
 	content := absContentPath(scan.Spec.Content)
 
 	cm := &corev1.ConfigMap{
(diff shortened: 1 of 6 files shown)


pkg/controller/compliancescan/config.go
```diff
@@ -248,47 +247,18 @@ func defaultOpenScapEnvCm(name string, scan *compv1alpha1.ComplianceScan) *corev
 		cm.Data[OpenScapTailoringDirEnvName] = OpenScapTailoringDir
 	}
 
-	if scan.Spec.HTTPSProxy != "" {
-		cm.Data[HTTPSProxyEnvName] = scan.Spec.HTTPSProxy
-	}
-
-	if scan.Spec.NoExternalResources {
-		cm.Data[DisconnectedInstallEnvName] = "true"
-	}
```
JAORMX, line 257: What happened to the proxy settings and external resources flag?

jhrozek: oops. I took the platform function as a base for the common one and forgot to copy over the disconnected install flags.

JAORMX:
@jhrozek might wanna remove "WIP" from the commit message, and just leave the word on the PR's title.

PR #518, opened 2020-12-01
Use skipRange in upstream release

JAORMX:
This makes it easier for us to release newer version as we only need to
track the latest version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -378,6 +378,8 @@ endif
 .PHONY: bundle
 bundle: check-operator-version operator-sdk ## Generate the bundle and packaging for the specific version (NOTE: Gotta specify the version with the OPERATOR_VERSION environment variable)
 	$(GOPATH)/bin/operator-sdk generate bundle -q --overwrite --version "$(OPERATOR_VERSION)"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -378,6 +378,8 @@ endif
 .PHONY: bundle
 bundle: check-operator-version operator-sdk ## Generate the bundle and packaging for the specific version (NOTE: Gotta specify the version with the OPERATOR_VERSION environment variable)
 	$(GOPATH)/bin/operator-sdk generate bundle -q --overwrite --version "$(OPERATOR_VERSION)"
+	sed -i '/replaces:/d' deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
+	sed -i "s/\(olm.skipRange: '>=.*\)<.*'/\1<$(OPERATOR_VERSION)'/" deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
```
mrogers950, line 382: https://github.com/openshift/compliance-operator/pull/517 adds skipRange value with double quotes, so the correct sed is
`sed -i 's/\(olm.skipRange: \">=.*\)<.*\"/\1<$(OPERATOR_VERSION)\"/'

PR #521, opened 2020-12-04
Only use latest tag when releasing the index image

JAORMX:
We're not really using the tagged releases. "latest" is the only tag
that contains the full database of releases we're releasing upstream.


jhrozek (approved):
/lgtm

PR #526, opened 2020-12-10
Track imagestream when using `deploy-local` target

JAORMX:
This ensures that subsequent updates to the imagestream get
automatically picked up by the deployment when using `deploy-local`.

So, when developing and pushing images to the cluster, it'll always use
the latest image.


jhrozek (approved):
/lgtm

PR #529, opened 2020-12-16
e2e/refactor: Use assertions instead of checking for errors everywhere

JAORMX:
this introduces the `E2EFatalf` function which will fail the test and
continue to the next one if called. This allows us to refactor some
functions and do assertions instead of having to check for errors
everywhere.

Not everything that could be refactored was refactored, but
that can be done in subsequent PRs.


jhrozek (approved):
/lgtm

PR #532, opened 2020-12-17
Bug 1909120: Always update default profilebundles on operator startup

JAORMX:
This ensures that the operator is using the ProfileBundle image that
it's supposed to be using. The intent of this patch is to help on
updates.

The workflow is as follows:

* An initial PB is created with a certain tag
* When an update happens, the Deployment rolls out and restarts the
operator
* The operator will check the contentImage reference and update (by
  patching it) if needed.

This way we ensure that folks are using supported images and not modify
the default profile bundles.


jhrozek (approved):
Looks good code-wise but I'm letting QE chime in if this helps upgrade testing.

jhrozek (approved):
/lgtm
let's merge this when CI finishes because without proper builds this is probably hard to test for our QE anyway. So let's merge the PR and create proper builds they can test and iterate on further.

PR #533, opened 2020-12-18
e2e/resiliency: Retry when trying to create subpool

JAORMX:
There have been some e2e failures related to creating the subpool we use
to do operations with remediations. This updates those operations to
retry if there's an error, and so make that more resilient.

This will also surface the errors if they happen, which we can then use
to debug CI when such issues happen.

This applied to all of the functions called when creating a sub-pool.


jhrozek (approved):
/lgtm

PR #534, opened 2020-12-18
Bug 1908990: Remove non-namespaced deployments

jhrozek:
In https://github.com/openshift/compliance-operator/pull/467 we
switched to using namespaced-names of profile parser deployments.

But we never covered the upgrade path, so the old non-namespaced names
were lingering around. This patch adds a one-off codepath that removes
the non-namespaced deployments.


JAORMX (approved):
looks good

JAORMX (commented):
/lgtm

PR #535, opened 2020-12-18
Update dependencies

JAORMX:
In hopes of not falling behind in releases and making it harder to
upgrade, this updates some tricky dependencies


JAORMX:
Pull-request updated, HEAD is now cd49392c52b6217fd99925ca8b2bdc1f1101891a

jhrozek (approved):
/lgtm
the version of gorilla/websocket we wanted to get rid of is gone from go.sum

PR #539, opened 2021-01-11
Remove outputType logic from TailoredProfile

JAORMX:
This was meant to create RHACM policies from a tailored profile. While
it was a good idea for a POC, it was never really supported, tested nor
documented. So let's remove it.


jhrozek (approved):
/lgtm

PR #540, opened 2021-01-11
profileparser: parallelize rule and variable parsing

JAORMX:
This introduces a simple benchmark test that runs the profile parser
1000 times.

Initially, this resulted in:

    Average Time: 0.173s ± 0.019s

This PR then parallelizes the parsing of rules and variables, using 4
workers for this. With the PR, the 1000 sample run gave:

    Average Time: 0.130s ± 0.014s

Which is ~24% improvement.


jhrozek (approved):
/lgtm

PR #543, opened 2021-01-20
refactor: Introduce scan type handler

JAORMX:
This splits the scan handling in a per-type object. This way we won't
have the big switch code blocks that we used to, and this will be done
in object-specific functions.

The node scan type handler will gather the relevant nodes on the
validation phase (which now happens before any phase in the reconcile
loop) and so it'll always have that reference of those objects when it
needs it.

This also adds the extra validation that if a node is marked as Unschedulable,
the scan will issue an event and won't proceed.

### DIFF
--- a/pkg/apis/compliance/v1alpha1/compliancescan_types.go
+++ b/pkg/apis/compliance/v1alpha1/compliancescan_types.go
@@ -1,7 +1,7 @@
 package v1alpha1
 
 import (
-	"fmt"
+	"errors"
 	"strings"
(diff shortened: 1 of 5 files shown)


JAORMX:
/retest

another CI issue

JAORMX:
Pull-request updated, HEAD is now 0043c922aac7de584d4478969dde6fe0437c8d19

jhrozek (approved):
/lgtm

PR #546, opened 2021-01-25
Bug 1919098: Add annotations to better deal with remediations

JAORMX:
This adds a couple of remediations that ComplianceSuites react to in
order to better deal with ComplianceRemediations:

* `compliance.openshift.io/apply-remediations`: Applies all of the
remediations created by a ComplianceSuite's Scans. It's similar to the
`AutoApplyRemediations` flag, but it's meant to be a one-shot operation,
which removes the annotation once it's done applying.

* `compliance.openshift.io/remove-outdated`: Can be used if either the
`apply-remediations` annotation is used, or the `autoApplyRemediations`
flag is enabled. It will automatically remove the outdated remediations
objects in order to update them (leaving the current ones to be applied).


jhrozek (commented):
See one comment inline, generally looks good

JAORMX:
/hold

@xiaojiey will test this first.

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -502,13 +529,18 @@ func (r *ReconcileComplianceSuite) applyGenericRemediation(rem compv1alpha1.Comp
...
 func (r *ReconcileComplianceSuite) applyMcfgRemediationAndPausePool(rem compv1alpha1.ComplianceRemediation,
+	suite *compv1alpha1.ComplianceSuite,
 	pool *mcfgv1.MachineConfigPool, logger logr.Logger) error {
 	remCopy := rem.DeepCopy()
 	// Only pause pools where the pool wasn't paused before and
 	// the remediation hasn't been applied
 	if !pool.Spec.Paused {
 		logger.Info("Pausing pool", "MachineConfigPool.Name", pool.Name)
 		pool.Spec.Paused = true
+		if remediationNeedsOutdatedRemoval(remCopy, suite) {
```
jhrozek, line 540: Why is the outdated remediation only touched when unpausing the pool? Wouldn't it mean that only the first outdated MC remediation in a pool gets hits this codepath? in general I think it would be cleaner to have all the code that touches the remediation on one place.

jhrozek (commented):
forgot to add the comment properly earlier

pkg/controller/compliancesuite/compliancesuite_controller.go
JAORMX: That was a mistake from my side. Good catch!

pkg/apis/compliance/v1alpha1/compliancesuite_types.go
```diff
@@ -175,3 +186,32 @@ func (s *ComplianceSuite) IsResultAvailable() bool {
 	result := s.LowestCommonResult()
 	return result != "" && result != ResultNotAvailable
 }
+
+// ApplyRemediations returns whether the ComplianceSuite requires
+// that the CoplianceRemediations that were generated from it be
+// applied.
+func (s *ComplianceSuite) ApplyRemediations() bool {
```
mrogers950, line 193: nit: the name ApplyRemediations implies a direct action as such.. Might be clearer as HasApplyRemediations or similar.

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -459,15 +437,16 @@ func (r *ReconcileComplianceSuite) reconcileRemediations(suite *compv1alpha1.Com
 		return reconcile.Result{}, nil
 	}
 
+	logger.Info("All scans are in Done phase. Post-procesing remediations")
```
mrogers950, line 440: s/procesing/processing

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -485,13 +464,61 @@ func (r *ReconcileComplianceSuite) reconcileRemediations(suite *compv1alpha1.Com
...
+	suite *compv1alpha1.ComplianceSuite,
+	scan *compv1alpha1.ComplianceScan,
+	mcfgpools *mcfgv1.MachineConfigPoolList,
+	affectedMcfgPools map[string]*mcfgv1.MachineConfigPool,
+	logger logr.Logger) error {
+
+	if utils.IsMachineConfig(rem.Spec.Current.Object) {
+		// get affected pool
+		pool, affectedPoolExists := r.getAffectedMcfgPool(scan, mcfgpools)
+		// we only ned to operate on pools that are affected
```
mrogers950, line 492: s/ned/need (this was left from the existing code)

mrogers950:
changes look good
/lgtm

JAORMX:
Pull-request updated, HEAD is now 9914a0c49ed26eadf72f241323b75842215c7442

PR #547, opened 2021-01-26
Add clarification about supported OpenShift version

### LINKED ISSUE #542: most platform checks fail because of missing configmap
Hi all

Not sure whether this is the right place to ask such a question. Feel free to redirect me to other channels if needed.

I run an OpenShift cluster 4.3 on IBM Cloud which is built on top of RHEL nodes. 

I installed the compliance operator and wanted to assess the platform compliance posture. Pretty much all of the ocp4 checks fail, so I started investigating. 
From the `ssg-ocp4-ds.xml` file I saw that e.g. the api-server checks are all based around the evaluation of a configmap `config` in different namespaces i.e. `oc get configmap config -n openshift-apiserver -ojson | jq -r '.data["config.yaml"]' | jq '.apiServerArguments["audit-log-path"]'`. The problem is, a configmap called `config` does not exist on my cluster, in any namespace. 

Any clue what I'm doing wrong or what I'm missing?

JAORMX:
We only support 4.6 up, so we should clarify this in the docs.

Closes #542

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,3 +332,5 @@ Platform scans
 
 Current testing has been done on OpenShift (OCP). The project is open to
 getting other platforms tested, so volunteers are needed for this.
+
+The current supported versions of OpenShift are 4.6 and up.


jhrozek (approved):
/lgtm

PR #548, opened 2021-01-27
Don't build binary with `-race` flag

JAORMX:
Per the docs[1], this flag is what's been causing out excessive memory
usage on the compliance-operator side. Let's remove it.

[1] https://golang.org/doc/articles/race_detector.html#Runtime_Overheads

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,7 +148,7 @@ test-broken-content-image:
 build: fmt manager ## Build the compliance-operator binary
 
 manager:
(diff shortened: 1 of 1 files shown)


jhrozek:
This is of course correct and a nice catch, but what do you think about leaving the flag in for the CI runs or at least for unit tests?

JAORMX:
@jhrozek let's turn it on for unit tests... but I didn't enable it just yet cause it seems we do have a race condition on the unit tests. I'll fix that on a subsequent PR.

jhrozek (approved):
/lgtm

Issue #555, opened 2021-02-04
ocp4-cis scan: "warnings: the server could not find the requested resource"


darrenfoong:
(Follow-up to an e-mail with @JAORMX)

Setup: OpenShift 4.6.13 on AWS with Compliance Operator 0.1.24 from Operator Hub.

I created a ScanSettingBinding with the following profiles:

```
profiles:
  - apiGroup: compliance.openshift.io/v1alpha1
    kind: Profile
    name: ocp4-cis
  - apiGroup: compliance.openshift.io/v1alpha1
    kind: Profile
    name: ocp4-cis-node
```

The scans finished running and I looked at the ComplianceSuite YAML.

Actual:

```
  scanStatuses:
    - name: ocp4-cis
      phase: DONE
      result: NON-COMPLIANT
      resultsStorage:
        name: ocp4-cis
        namespace: openshift-compliance
      warnings: the server could not find the requested resource # <--- interesting line
    - name: ocp4-cis-node-worker
      phase: DONE
      result: NON-COMPLIANT
      resultsStorage:
        name: ocp4-cis-node-worker
        namespace: openshift-compliance
    - name: ocp4-cis-node-master
      phase: DONE
      result: NON-COMPLIANT
      resultsStorage:
        name: ocp4-cis-node-master
        namespace: openshift-compliance
```

Expected: no warnings.

PR #562, opened 2021-02-05
Output relevant object in warnings

JAORMX:
If an object couldn't be fetched (for whatever reason) the warning was
fairly cryptic. This exposes the URI that was tried, so users can debug
more easily what the issue is.

Closes: https://github.com/openshift/compliance-operator/issues/555


jhrozek (approved):
/lgtm

PR #567, opened 2021-02-11
api: Surface a new attribute instructions that contains OCIL

jhrozek:
Extends the ComplianceCheckResult API object with a new attribute
instructions that contains the OCIL attribute from OpenScap rule.

This attribute contains the manual checks for cases when a rule cannot
be evaluated with a OVAL check.

### DIFF
--- a/deploy/crds/compliance.openshift.io_compliancecheckresults_crd.yaml
+++ b/deploy/crds/compliance.openshift.io_compliancecheckresults_crd.yaml
@@ -35,6 +35,11 @@ spec:
           id:
             description: A unique identifier of a check
             type: string
+          instructions:
+            description: How to evaluate if the rule status manually. If no automatic
+              test is present, the rule status will be MANUAL and the administrator
+              should follow these instructions.
+            type: string
           kind:
             description: 'Kind is a string value representing the REST resource this
               object represents. Servers may infer this from the endpoint the client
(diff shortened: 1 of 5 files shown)


deploy/crds/compliance.openshift.io_compliancecheckresults_crd.yaml
```diff
@@ -35,6 +35,9 @@ spec:
           id:
             description: A unique identifier of a check
             type: string
+          instructions:
+            description: Manual check steps in case the rule can't be evaluated automatically
```
JAORMX, line 39: Does this only show if there rule can't be evaluated automatically?

jhrozek:
On Thu, Feb 11, 2021 at 09:47:04AM -0800, Juan Osorio Robles wrote:
> @JAORMX commented on this pull request.
> 
> 
> 
> > @@ -35,6 +35,9 @@ spec:
>            id:
>              description: A unique identifier of a check
>              type: string
> +          instructions:
> +            description: Manual check steps in case the rule can't be evaluated automatically
> 
> Does this only show if there rule can't be evaluated automatically? 

No, always.

JAORMX:
> On Thu, Feb 11, 2021 at 09:47:04AM -0800, Juan Osorio Robles wrote:
> > @JAORMX commented on this pull request.
> > 
> > 
> > 
> > > @@ -35,6 +35,9 @@ spec:
> >            id:
> >              description: A unique identifier of a check
> >              type: string
> > +          instructions:
> > +            description: Manual check steps in case the rule can't be evaluated automatically
> > 
> > Does this only show if there rule can't be evaluated automatically? 
> 
> No, always.
> 

Let's change the description, with the current way it's written out sounds like that key will only have text if automatic remediation wasn't possible

PR #568, opened 2021-02-11
Add an explicit MANUAL compliancecheck result status

jhrozek:
We were mapping both notchecked and info coming from the XCCFD results
to INFO in compliancecheckresult. This is confusing as users want to
generally know what do they need to check manually.

Therefore let's add a separate MANUAL status for checks that couldn't
have been evaluated by SCAP.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -1479,11 +1479,26 @@ func TestE2E(t *testing.T) {
 						Namespace: namespace,
 					},
 					ID:       "xccdf_org.ssgproject.content_rule_wireless_disable_in_bios",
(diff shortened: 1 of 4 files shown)


tests/e2e/e2e_test.go
```diff
@@ -1479,7 +1479,7 @@ func TestE2E(t *testing.T) {
 						Namespace: namespace,
 					},
 					ID:       "xccdf_org.ssgproject.content_rule_wireless_disable_in_bios",
-					Status:   compv1alpha1.CheckResultInfo,
+					Status:   compv1alpha1.CheckResultManual,
```
JAORMX, line 1482: Do we have any other rules that would be INFO still? We should cover that too

PR #572, opened 2021-02-16
Label ComplianceCheck CRs that have automated remediations

jhrozek:
Our users have a hard time figuring out which ComplianceCheck findings
can be remediated automatically. Let's add a label so that they can
filter out checks that have a remediation.

Jira: [CMP-855](https://issues.redhat.com/browse/CMP-855)

### DIFF
--- a/pkg/apis/compliance/v1alpha1/compliancecheckresult_types.go
+++ b/pkg/apis/compliance/v1alpha1/compliancecheckresult_types.go
@@ -14,6 +14,11 @@ type ComplianceCheckStatus string
 const ComplianceCheckResultStatusLabel = "compliance.openshift.io/check-status"
 const ComplianceCheckResultSeverityLabel = "compliance.openshift.io/check-severity"
 
+// ComplianceCheckResultLabel defines a label that will be included in the
+// ComplianceCheckResult objects. It indicates whether the result has an automated
+// remediation or not.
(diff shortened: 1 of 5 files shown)


pkg/apis/compliance/v1alpha1/compliancecheckresult_types.go
```diff
@@ -14,6 +14,11 @@ type ComplianceCheckStatus string
 const ComplianceCheckResultStatusLabel = "compliance.openshift.io/check-status"
 const ComplianceCheckResultSeverityLabel = "compliance.openshift.io/check-severity"
 
+// ComplianceCheckResultLabel defines a label that will be included in the
+// ComplianceCheckResult objects. It indicates whether the result has an automated
+// remediation or not.
+const ComplianceCheckResultHasRemediation = "compliance.openshift.io/automated-remediation"
```
JAORMX, line 20: Can you add docs about this? Also about its usage;

e.g.

* How do I search for results that have automated remediations?
* How do I search for results without automated remediations?

JAORMX (changes requested):

pkg/apis/compliance/v1alpha1/compliancecheckresult_types.go
jhrozek: Done

jhrozek:
You know what, let me also add a test that ensures that the check for an existing remediation is labelled correctly.
/hold

PR #573, opened 2021-02-19
Bug 1919311: Use a separate SA for resultserver

jhrozek:
resultserver was relying on the "default" SA to apply the restricted SCC on
the resultserver pod. While this is OK in the general case, in environments
that bound the "default" SA to some other SCC than restricuted, this would
have broken the assumption by applying a different SCC and not propagating
settings like fsGroup to the pod.

Jira: [OCPBUGSM-23632](https://issues.redhat.com/browse/OCPBUGSM-23632)

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -220,6 +220,20 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
+  name: resultserver
+rules:
+- apiGroups:
+  - security.openshift.io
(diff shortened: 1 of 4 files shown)


JAORMX (approved):

jhrozek:
huh, what's going on with the unit tests? They work fine for me..
/test unit

JAORMX:
This PR LGTM. I'll wait for @xiaojiey or @pdhamdhe to test this and add the `lgtm` label.

PR #575, opened 2021-02-22
Add dependency handling for remediations

JAORMX:
This adds the ability for the compliance operator to handle dependencies
in remediations.

These remediations are defined in the content through kubernetes
annotations which look as follows:

```
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  annotations:
    complianceascode.io/depends-on: {{{ XCCDF ID }}}
  spec:
    config:
...
```

Multiple XCCDF ID's can be specified by separating with comma's.

these XCCDF ID's refer to rules that the remediation depends on, these
rules will take the form of ComplianceCheckResults.

The annotation will be read in the operator, and persisted to the
remediation objects with the following format:

```
  annotations:
    compliance.openshift.io/depends-on: {{{ XCCDF ID }}}
```

When reconciling the remediation, the operator will only apply it if all
the dependencies are met. If they are not, it won't apply the
remediation and output a relevant event. If they are, it'll annotate the
remediation with the following key: `compliance.openshift.io/dependencies-met`

If a remediation has un-met dependencies, this also adds a label to the
remediation object that allows users to filter the remediations
appropraitely: `compliance.openshift.io/has-unmet-dependencies`

Co-Authored-By: Jakub Hrozek <jhrozek@redhat.com>

### DIFF
--- a/cmd/manager/aggregator.go
+++ b/cmd/manager/aggregator.go
@@ -383,7 +383,7 @@ func createResults(crClient *complianceCrClient, scan *compv1alpha1.ComplianceSc
 
 		// The state even if set in the object would have been overwritten by the call to
 		// spec update, so we keep the state separately in a variable
-		stateUpdate := pr.Remediation.Status.ApplicationState
+		stateUpdate := compv1alpha1.RemediationPending
 
 		remkey := getObjKey(pr.Remediation.GetName(), pr.Remediation.GetNamespace())
 		foundRemediation := &compv1alpha1.ComplianceRemediation{}
@@ -434,9 +434,9 @@ func createResults(crClient *complianceCrClient, scan *compv1alpha1.ComplianceSc
 		}
 
 		// Update the status as needed
-		if stateUpdate == compv1alpha1.RemediationOutdated {
-			if err := updateRemediationState(crClient, pr.Remediation, stateUpdate); err != nil {
-				return nil
+		if remExists {
+			if err := updateRemediationStatus(crClient, pr.Remediation, stateUpdate); err != nil {
+				return err
 			}
 		}
 	}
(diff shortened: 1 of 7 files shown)


jhrozek:
I pushed a follow up, which is mostly untested past make and make check, because my cluster expired and it's too late to wait for another one :-)

jhrozek:
This is the diff of the tests:
```
diff --git a/pkg/controller/complianceremediation/complianceremediation_controller_test.go b/pkg/controller/complianceremediation/complianceremediation_controller_test.go
index 5dcff57a..2f37c6f8 100644
--- a/pkg/controller/complianceremediation/complianceremediation_controller_test.go
+++ b/pkg/controller/complianceremediation/complianceremediation_controller_test.go
@@ -262,6 +262,10 @@ var _ = Describe("Testing complianceremediation controller", func() {
                                // reflects an admin having removed it.
                                err = reconciler.client.Update(context.TODO(), remediationinstance)
                                Expect(err).NotTo(HaveOccurred())
+                               // mock that the remediation was applied
+                               remediationinstance.Status.ApplicationState = compv1alpha1.RemediationApplied
+                               err = reconciler.client.Status().Update(context.TODO(), remediationinstance)
+                               Expect(err).NotTo(HaveOccurred())
                        })

                        It("should remove the outdated remediation label", func() {
```

We were hitting this new patch in the remediation controller:
```

	if remediationInstance.Status.ApplicationState == "" {
		rCopy := remediationInstance.DeepCopy()
		rCopy.Status.ApplicationState = compv1alpha1.RemediationPending
		if updErr := r.client.Status().Update(context.TODO(), rCopy); updErr != nil {
			return reconcile.Result{}, updErr
		}
	}
```

That caused the Reconcile loop to shortcut sooner than we expected.

JAORMX:
@jhrozek i think this is good to go. What do you think?

jhrozek (approved):
/lgtm

PR #576, opened 2021-02-23
Fix profileparser race condition

JAORMX:
The tests weren't properly parallelized, so this fixes that by
aggregating results through a channel.


jhrozek (approved):
/lgtm

PR #578, opened 2021-02-24
Release v0.1.27

JAORMX:
* Dependency handling for remediations
* resultserver now has its own ServiceAccount so we can bind it to a specific SCC.
* ComplianceChecks with automated remediations are now appropriately labeled.


jhrozek (approved):
/lgtm

PR #584, opened 2021-03-02
Add conditions to relevant objects

JAORMX:
Objects that users are meant to interact with more often now have
appropraite conditions:

* ProfileBundle: Has a "Ready" condition that indicates that a bundle
  has been processed successfully.

* ScanSettingBinding: Has a "Ready" condition that indicates that the
  binding has already generated a relevant ComplianceSuite. This object
  now also has an `outputRef` in the status, which indicates what
  ComplianceSuite was generated.

* ComplianceSuite: Has a "Ready" condition that indicates if the suite
  has results. There is also a "Processing" condition that indicates if
  the suite is running or not.

These conditions make it easier to interact with these objects using
standard kubectl/oc. We now can do:

```
oc wait --for condition=ready scansettingbinding <name>
```

To wait for a certain condition to be met.


jhrozek (approved):
This version works for me and the conditions appear sane
/lgtm

PR #586, opened 2021-03-04
Add short names to all CRDs

JAORMX:
This is a usability enhancement, as folks won't need to write the whole
object name any more, and instead can just refer to the short names.

e.g. instead of writing

```
oc get compliancesuites
```

folks can just do

`oc get suites` or `oc get suite`


jhrozek (approved):
The names all make sense and seem to be working fine.

PR #591, opened 2021-03-08
Fix possible nil dereference with ScanSettingBinding

mrogers950:
It appears we might hit this condition before OutputRef exists.


JAORMX:
oops! Good catch!

/lgtm

PR #593, opened 2021-03-09
Set resource requests and limits for ProfileBundle workload

JAORMX:
This is a hardening measure and ensures that the operator doesn't become
a noisy neighbour.


jhrozek (approved):
These limits make sense
/lgtm

PR #594, opened 2021-03-09
Add resource requests and limits for suite rerunner

JAORMX:
This is a hardening measure and ensures that the operator doesn't become
a noisy neighbour.


jhrozek (approved):
/lgtm

PR #595, opened 2021-03-09
Create "default-auto-apply" ScanSetting

JAORMX:
This ScanSetting will be automatically deployed by the operator, and
allows folks to just let their scans auto apply remediations and also
auto update them. This is more usable than always having to create
a new ScanSetting object or modify the default one.

### DIFF
--- a/cmd/manager/operator.go
+++ b/cmd/manager/operator.go
@@ -59,7 +59,11 @@ var (
 		"rhcos4",
 		"ocp4",
 	}
-	defaultScanSettingsName = "default"
+)
+
+const (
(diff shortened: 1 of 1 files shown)


cmd/manager/operator.go
```diff
@@ -59,7 +59,11 @@ var (
 		"rhcos4",
 		"ocp4",
 	}
+)
+
+const (
 	defaultScanSettingsName = "default"
+	applyScanSettingsName   = "apply"
```
mrogers950, line 66: I was thinking since this is just a variant of default, it might be clearer as `default-apply` or `default-auto-apply`

PR #604, opened 2021-03-22
Replace xmldom by xmlquery for profile parsing

JAORMX:
xmldom seems to be a stale project, and its current issues present some
usability limitations for us:

* Excessive resource usage
* Inability to get full xml content from sub-nodes.

The second point results in us getting malformed descriptions in some
rules.

This aims to address that by replacing the XML parsing [for another
library that's more active](https://github.com/antchfx/xmlquery).

This is only used when generated profiles, rules and variables.

### DIFF
--- a/cmd/manager/profileparser.go
+++ b/cmd/manager/profileparser.go
@@ -8,9 +8,9 @@ import (
 	"os"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
+	"github.com/antchfx/xmlquery"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	"github.com/spf13/cobra"
-	"github.com/subchen/go-xmldom"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -42,11 +42,6 @@ func defineProfileParserFlags(cmd *cobra.Command) {
 	// Add flags registered by imported packages (e.g. glog and
 	// controller-runtime)
 	flags.AddGoFlagSet(flag.CommandLine)
-}
-
-// XMLDocument is a wrapper that keeps the interface XML-parser-agnostic
-type XMLDocument struct {
-	*xmldom.Document
 }
 
 func newParserConfig(cmd *cobra.Command) *profileparser.ParserConfig {
@@ -140,7 +135,7 @@ func runProfileParser(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 	bufContentFile := bufio.NewReader(contentFile)
-	contentDom, err := xmldom.Parse(bufContentFile)
+	contentDom, err := xmlquery.Parse(bufContentFile)
 	if err != nil {
 		log.Error(err, "Couldn't read the content XML")
 		updateProfileBundleStatus(pcfg, pb, fmt.Errorf("Couldn't read content XML: %s", err))
--- a/pkg/profileparser/profileparser.go
+++ b/pkg/profileparser/profileparser.go
@@ -11,9 +11,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
+	"github.com/antchfx/xmlquery"
 	cmpv1alpha1 "github.com/openshift/compliance-operator/pkg/apis/compliance/v1alpha1"
 	"github.com/openshift/compliance-operator/pkg/xccdf"
-	"github.com/subchen/go-xmldom"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
@@ -53,7 +53,7 @@ func GetPrefixedName(pbName, objName string) string {
 	return pbName + "-" + objName
 }
 
-func ParseBundle(contentDom *xmldom.Document, pb *cmpv1alpha1.ProfileBundle, pcfg *ParserConfig) error {
+func ParseBundle(contentDom *xmlquery.Node, pb *cmpv1alpha1.ProfileBundle, pcfg *ParserConfig) error {
 	// One go routine per type
 	errors := make(chan error)
 	done := make(chan string)
@@ -264,13 +264,12 @@ func deleteIfNotCurrentDigest(client runtimeclient.Client, imageDigest string, i
 	return client.Delete(context.TODO(), item)
 }
 
-func getVariableType(varNode *xmldom.Node) cmpv1alpha1.VariableType {
-	typeAttr := varNode.GetAttribute("type")
-	if typeAttr == nil {
-		return cmpv1alpha1.VarTypeString
-	}
+func getVariableType(varNode *xmlquery.Node) cmpv1alpha1.VariableType {
+	typeAttr := varNode.SelectAttr("type")
 
-	switch typeAttr.Value {
+	switch typeAttr {
+	case "":
+		return cmpv1alpha1.VarTypeString
 	case "string":
 		return cmpv1alpha1.VarTypeString
 	case "number":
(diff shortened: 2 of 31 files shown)


pkg/profileparser/profileparser.go
```diff
@@ -294,19 +293,19 @@ func ParseProfilesAndDo(contentDom *xmldom.Document, pb *cmpv1alpha1.ProfileBund
...
+	profileObjs := xmlquery.Find(profileRoot, "//xccdf-1.2:Profile")
 	for _, profileObj := range profileObjs {
 
-		id := profileObj.GetAttributeValue("id")
+		id := profileObj.SelectAttr("id")
 		if id == "" {
 			return LogAndReturnError("no id in profile")
 		}
-		title := profileObj.FindOneByName("title")
+		title := profileObj.SelectElement("xccdf-1.2:title")
```
mrogers950, line 304: nit - let's use a constant for the "xccdf-1.2:" prefix

pkg/profileparser/profileparser.go
```diff
@@ -376,15 +375,14 @@ func parseProfileFromNode(profileRoot *xmldom.Node, pb *cmpv1alpha1.ProfileBundl
 	return nil
 }
 
-func getProductTypeAndName(root *xmldom.Node, defaultType cmpv1alpha1.ComplianceScanType, defaultName string) (cmpv1alpha1.ComplianceScanType, string) {
+func getProductTypeAndName(root *xmlquery.Node, defaultType cmpv1alpha1.ComplianceScanType, defaultName string) (cmpv1alpha1.ComplianceScanType, string) {
 	// Loop ourselves b/c we are only interested in direct children
```
mrogers950, line 379: This comment doesn't fit anymore

cmd/manager/profileparser.go
```diff
@@ -8,6 +8,7 @@ import (
 	"os"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
+	"github.com/antchfx/xmlquery"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	"github.com/spf13/cobra"
 	"github.com/subchen/go-xmldom"
```
jhrozek, line 14: go-xmldom is now only used in a struct definition that itself is unused. I wonder if we can get rid of it? There are some other uses that are still used for real, do you plan on getting rid of them as well?

pkg/profileparser/profileparser.go
```diff
@@ -376,15 +375,13 @@ func parseProfileFromNode(profileRoot *xmldom.Node, pb *cmpv1alpha1.ProfileBundl
...
 
-func getProductTypeAndName(root *xmldom.Node, defaultType cmpv1alpha1.ComplianceScanType, defaultName string) (cmpv1alpha1.ComplianceScanType, string) {
-	// Loop ourselves b/c we are only interested in direct children
-	for _, child := range root.Children {
-		if child.Name == "platform" {
-			return parseProductTypeAndName(child.GetAttributeValue("idref"), defaultType, defaultName)
-		}
-	}
+func getProductTypeAndName(root *xmlquery.Node, defaultType cmpv1alpha1.ComplianceScanType, defaultName string) (cmpv1alpha1.ComplianceScanType, string) {
+	p := root.SelectElement("xccdf-1.2:platform")
```
jhrozek, line 379: So this part is tricky. See that getProducTypeAndName is used twice, this is IIRC because each profile can redefine the CPE. That's why the original code used to only check for direct children and not just search for any platform. You could do that (I think) by searching for `/platform`. But the thing is that I don't know if any profile we ship does this -- so maybe it's just additional complexity we don't need to have in the code. IIRC I was just coding by following the standard back then.

jhrozek (commented):
There are two comments inline. In addition, I wonder:
 - did you test the XPath performance of the new library? The old one was quite slow, that's why we often just loop and search for what we need ourselves. The code might be more compact using XPath queries, OTOH I don't know how correct this library as and..maybe we shouldn't fix what is not broken
 - do you plan on replacing the other uses of the older library?

cmd/manager/profileparser.go
JAORMX: > go-xmldom is now only used in a struct definition that itself is unused. I wonder if we can get rid of it?

Good point, gotta delete that one!

> There are some other uses that are still used for real, do you plan on getting rid of them as well?
Not in this iteration, maybe in the future. The thing that I'm concerned about is that xmldom disregards XML namespaces, which is how we coded this whole thing; xmlquery cares about namespaces. In some old instances of the ARF, I noticed the namespace was randomized (they set something like `ns10` and stuff like this) so we might not be able to use xmlquery for ARF parsing unless that gets fixed in OpenSCAP.

pkg/profileparser/profileparser.go
JAORMX: We do this with `ocp4-cis-node`, where we say that applies for a node and not the platform. That's a good point! I need to verify that one.

pkg/profileparser/profileparser.go
JAORMX: I wonder how the compiler would optimize this... would it do the sprintf operation on each instance? or does it optimize it to just set the resulting string?

pkg/profileparser/profileparser.go
JAORMX: Seems to me like it parsed it correctly:
```
apiVersion: compliance.openshift.io/v1alpha1
description: This profile defines a baseline that aligns to the Center for Internet
  Security®&#xA;Red Hat OpenShift Container Platform 4 Benchmark™, V0.3, currently
  unreleased.&#xA;&#xA;This profile includes Center for Internet Security®&#xA;Red
  Hat OpenShift Container Platform 4 CIS Benchmarks™ content.&#xA;&#xA;Note that this
  part of the profile is meant to run on the Operating System that&#xA;Red Hat OpenShift
  Container Platform 4 runs on top of.&#xA;&#xA;This profile is applicable to OpenShift
  versions 4.6 and greater.
id: xccdf_org.ssgproject.content_profile_cis-node
kind: Profile
metadata:
  annotations:
    compliance.openshift.io/image-digest: pb-ocp4hx7kq
    compliance.openshift.io/product: redhat_openshift_container_platform_node_4
    compliance.openshift.io/product-type: Node
  creationTimestamp: "2021-03-23T06:52:56Z"
  generation: 1
  labels:
    compliance.openshift.io/profile-bundle: ocp4
  name: ocp4-cis-node
  namespace: openshift-compliance
```

JAORMX:
> There are two comments inline. In addition, I wonder:
> 
>     * did you test the XPath performance of the new library? The old one was quite slow, that's why we often just loop and search for what we need ourselves. The code might be more compact using XPath queries, OTOH I don't know how correct this library as and..maybe we shouldn't fix what is not broken

I didn't test XPath performance, I based this change on maintainability (the other one doesn't seem to be maintained while xquery seems to be). I'd use some help performance testing this.

> 
>     * do you plan on replacing the other uses of the older library?

I'd like to, but there are some questions I still have on the ARF namespacing. So that will be done in another PR.

First I was thinking of wrapping all of the XML queries into a single interface. Then just have the two implementations there. Finally, when the time comes, we could easily replace it. But I was hoping to do that in another PR. If you folks think it's a good idea, I could get that into this PR.

JAORMX:
Pull-request updated, HEAD is now 7578d12dcdfbd052927dd97886b1465a071ad959

JAORMX:
the immediate benefit I want to get out of this is finally getting appropriate descriptions in our objects, which was not happening with xmldom.

jhrozek:
can you please check if the availableFixes still render well for you with this patch? I'm seeing this locally:
```
grep -c -i availablefixes /tmp/rules.new /tmp/rules.old
/tmp/rules.new:0
/tmp/rules.old:434
```
rules.old is generated with master rules.new with your patch. btw the remediation are still created correctly, we don't use the availableFixes attribute there..

JAORMX:
/hold

I'll verify the avilablefixes

JAORMX:
> can you please check if the availableFixes still render well for you with this patch? I'm seeing this locally:
> 
> ```
> grep -c -i availablefixes /tmp/rules.new /tmp/rules.old
> /tmp/rules.new:0
> /tmp/rules.old:434
> ```
> 
> rules.old is generated with master rules.new with your patch. btw the remediation are still created correctly, we don't use the availableFixes attribute there..

You're right, this introduces a regression on that key.

JAORMX:
@jhrozek pushed a fix

JAORMX:
@mrogers950 could you revisit this?

pkg/profileparser/profileparser.go
mrogers950: To my knowledge the '+' concatenation allocates a new string and combines the two, and performs worse as the strings become larger, so for short strings it would be quicker than an sprintf. But keeping it like it is now would be even better, so let's ignore this nit.

pkg/profileparser/profileparser_test.go
```diff
@@ -579,6 +582,9 @@ var _ = Describe("Testing parse rules", func() {
 
 		stdParser := newStandardParser()
 		nonce := names.SimpleNameGenerator.GenerateName("pb-")
+		zaplog, _ := zap.NewDevelopment()
+		log = zapr.NewLogger(zaplog)
```
mrogers950, line 586: This is unused? What's it do?

JAORMX: It sets the global log instance just in the test to Development. Thus printing the reconciler's log messages when running tests in verbose mode.

JAORMX: At some point we should move away from those log instances in global variables and instantiate them as part of the reconcile object.

mrogers950: Ah, sounds good.

PR #605, opened 2021-03-22
Make pauser pod limit be 15Mi

JAORMX:
In some cases, the kubelet was complaining that it was too little for
the limit, this fixes that.


jhrozek:
one of the tests was slower by 3 miliseconds than our limit :-)

jhrozek (approved):
/lgtm

PR #620, opened 2021-04-13
Switch to using go 1.16

JAORMX:


jhrozek:
Why is this on hold? (I'd like us to switch to 1.16 also in FIO to cover one BZ that just needs a rebuild..)

JAORMX:
It's on hold because I don't think we can backport this to 4.7 and 4.6.... or can we?

jhrozek:
On Tue, Apr 20, 2021 at 11:35:22PM -0700, Juan Osorio Robles wrote:
> It's on hold because I don't think we can backport this to 4.7 and 4.6.... or can we?
> 

As long as the new go version is only used for upstream then why not?
Downstream chooses its own builder image.

JAORMX:
Pull-request updated, HEAD is now 7ff8320de0996c0265ab6f4e8d09e80a9eaedfb8

PR #622, opened 2021-04-15
Add management workload annotations

JAORMX:
In support of the workload partitioning feature (openshift/enhancements#703),
we need to add annotations to all management pods and namespaces so they can be
properly identified and assigned to segregated management cores on
clusters configured to do so.

The proposed workloads to only run in the management-designated
resources are the following:

* Compliance Opertor Deployment: Runs reconcile loops and schedules
  everyting. Only meant for management and shouldn't run on
  non-management resources.

* Result Server: Gets the full ARF raw results and positions them into the
  Persistent Volume.

* Aggregator: Gets a subset of the results and builds ComplianceCheckResult
  and ComplianceRemediation objects.

* Platform scan: This type of scan queries the kube API before running
  OpenSCAP to evaluate them. There's no need for this to run on the
  non-management resources.

* Suite-rerunner: is a one-off job that ensures that the ComplianceSuite
  is re-run at the configured time.

* ProfileBundle's profileparser: parses the data stream and builds
  profiles, rules and variable kubernetes objects out of it.

Node type of scans was purposefully left out of this as there may be the possibility
of nodes not being configured with these management resources, and thus the
Kubelet wouldn't schedule them appropriately as it should.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -11,6 +11,8 @@ spec:
     metadata:
       labels:
         name: compliance-operator
+      annotations:
+        workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
     spec:
       serviceAccountName: compliance-operator
       containers:
--- a/pkg/controller/compliancescan/aggregator.go
+++ b/pkg/controller/compliancescan/aggregator.go
@@ -37,6 +37,9 @@ func newAggregatorPod(scanInstance *compv1alpha1.ComplianceScan, logger logr.Log
 			Name:      podName,
 			Namespace: common.GetComplianceOperatorNamespace(),
 			Labels:    podLabels,
+			Annotations: map[string]string{
+				"workload.openshift.io/management": `{"effect": "PreferredDuringScheduling"}`,
+			},
 		},
 		Spec: corev1.PodSpec{
 			ServiceAccountName: aggregatorSA,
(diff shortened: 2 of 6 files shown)


jhrozek:
The general idea of everything but node scans seems good to me, but it also seems like CI doesn't like the annotations:
```
[36mINFO[0m[2021-04-15T09:05:54Z] Error: failed to overwrite operator image in the namespaced manifest: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.ObjectMeta: v1.ObjectMeta.Labels: ReadString: expects " or n, but found {, error found in #10 byte of ...|tations":{"workload.|..., bigger context ...|,"template":{"metadata":{"labels":{"annotations":{"workload.openshift.io/management":"{\"effect\": \|... 
```

JAORMX:
> The general idea of everything but node scans seems good to me, but it also seems like CI doesn't like the annotations:
> 
> ```
> �[36mINFO�[0m[2021-04-15T09:05:54Z] Error: failed to overwrite operator image in the namespaced manifest: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.ObjectMeta: v1.ObjectMeta.Labels: ReadString: expects " or n, but found {, error found in #10 byte of ...|tations":{"workload.|..., bigger context ...|,"template":{"metadata":{"labels":{"annotations":{"workload.openshift.io/management":"{\"effect\": \|... 
> ```

that was because I accidentally put the annotations under the labels. that's fixed.

jhrozek (approved):

PR #623, opened 2021-04-15
Make use of the HTTPS_PROXY environment variable

jhrozek:
OpenShift allows the administrator to set a global proxy by editing the
proxy/cluster object:
    https://github.com/openshift/api/blob/master/config/v1/types_proxy.go
The OLM can in turn watch for the proxy objects and set the HTTPS_PROXY
environment variable in the CO deployment.

This patch sets the proxy environment variable for the openscap pods
based on the value of the HTTPS_PROXY environment variable passed for
the operator, thus taking advantage of the global proxy settings.

The HTTPSProxy member of the ComplianceScanSettings structure is
retained, just an attribute is added that the OCP proxy is preferred.

Jira: [CMP-846](https://issues.redhat.com/browse/CMP-846)

### DIFF
--- a/pkg/apis/compliance/v1alpha1/compliancescan_types.go
+++ b/pkg/apis/compliance/v1alpha1/compliancescan_types.go
@@ -155,6 +155,7 @@ type ComplianceScanSettings struct {
 	// resources could be, for instance, CVE feeds. This is useful for disconnected
 	// installations without access to a proxy.
 	NoExternalResources bool `json:"noExternalResources,omitempty"`
+	// It is recommended to set the proxy via the config.openshift.io/Proxy object
 	// Defines a proxy for the scan to get external resources from. This is useful for
 	// disconnected installations with access to a proxy.
 	HTTPSProxy string `json:"httpsProxy,omitempty"`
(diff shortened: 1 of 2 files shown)


jhrozek:
@xiaojiey as part of our testing of disconnected installations, do we have some environment in house with a proxy where I could test this PR properly?

JAORMX (approved):
Looks trivial enough. Let me know when I should merge this.

jhrozek:
> Looks trivial enough. Let me know when I should merge this.

Please merge this, I tested at least that the variable gets populated into the openscap CM which is what the old method did:
```
oc describe cm workers-cis-openscap-env-map
Name:         workers-cis-openscap-env-map
Namespace:    openshift-compliance
Labels:       compliance.openshift.io/scan-name=workers-cis
              complianceoperator.openshift.io/scan-script=
Annotations:  <none>

Data
====
HTTPS_PROXY:
----
104.236.80.199:8080
```

JAORMX (approved):
/lgtm

JAORMX:
In a subsequent PR, can you add the OLM annotation to the CSV?

PR #624, opened 2021-04-16
Enhance TailoredProfile validation

JAORMX:
This enhances the TailoredProfile validation by checking that all the
rules belong to the same ProfileBundle. This way, users are not able to
craft invalid profiles that would reference objects not existent in the
data stream.

### DIFF
--- a/pkg/controller/tailoredprofile/tailoredprofile_controller.go
+++ b/pkg/controller/tailoredprofile/tailoredprofile_controller.go
@@ -5,11 +5,11 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	"github.com/openshift/compliance-operator/pkg/controller/common"
 	"github.com/openshift/compliance-operator/pkg/xccdf"
 
-	compliancev1alpha1 "github.com/openshift/compliance-operator/pkg/apis/compliance/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


JAORMX:
Pull-request updated, HEAD is now 359294d129fa0a3f2bb6bbdef9313d48da3ef668

pkg/controller/tailoredprofile/tailoredprofile_controller.go
```diff
@@ -191,85 +166,125 @@ func (r *ReconcileTailoredProfile) validateTailoredProfile(tp *compliancev1alpha
 	return true, nil
 }
 
-func (r *ReconcileTailoredProfile) getRulesFromSelections(tp *compliancev1alpha1.TailoredProfile) (map[string]*compliancev1alpha1.Rule, bool, error) {
-	rules := make(map[string]*compliancev1alpha1.Rule)
+// getProfileInfo gets the needed profile info (profile and profile bundle) from a tailored profile
+func (r *ReconcileTailoredProfile) getProfileInfo(tp *cmpv1alpha1.TailoredProfile) (*cmpv1alpha1.Profile, *cmpv1alpha1.ProfileBundle, error) {
```
mrogers950, line 170: getProfileInfoFromExtends() has the same parameters, is this one needed?

JAORMX: It isn't. at a previous implementation, I had gotten this to work with "profiles from scratch" but ended removing that. Let me clean this up.

PR #625, opened 2021-04-21
Move relevant workloads to be scheduled on the master nodes only

JAORMX:
This ensures that the worker nodes are free for customer workloads only
and thus we only use master resources.

The main use-cases for this are:

* Hypershift, where the master nodes are in different infrastructure (maybe even a different cloud) than the worker nodes.
* Remote worker nodes (results should only reside on the master).


jhrozek (approved):
/lgtm
(for anyone wondering, the operator itself would already only run on master nodes per toleratoins and nodeSelector in `deployment.yaml`)

PR #626, opened 2021-04-22
Limit resource usage for all workloads

JAORMX:
now that our profiles aren't using as many resources as they used to,
let's add appropriate requests and limits for all resources.

The max CPU usage that a workload is allowed to have for the compliance
operator is 100m, this is with the intent of leaving resources for other
workloads so this would work in single-node deployments.


jhrozek (approved):
/lgtm

PR #627, opened 2021-04-22
Update most dependencies

JAORMX:
While we're not updating to new major versions of the kube APIs and
operator-sdk, this does update the minor ones. For other dependencies
we do try to udpate to the latest.

This also makes the necessary code changes to be able to upgrade the
dependencies.


JAORMX:
Pull-request updated, HEAD is now 240656d06f40cef71815264fe58b45cf33833199

PR #629, opened 2021-04-22
Update gosec to v2.7.0

JAORMX:
This updates gosec and also fixes most warnings it gives.

Most of them were the following

```
Implicit memory aliasing in for loop.
```


pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -590,13 +592,14 @@ func (r *ReconcileComplianceSuite) applyMcfgRemediationAndPausePool(rem compv1al
 	return nil
 }
 
-func (r *ReconcileComplianceSuite) getAffectedMcfgPool(scan *compv1alpha1.ComplianceScan, mcfgpools *mcfgv1.MachineConfigPoolList) (mcfgv1.MachineConfigPool, bool) {
-	for _, pool := range mcfgpools.Items {
-		if utils.McfgPoolLabelMatches(scan.Spec.NodeSelector, &pool) {
+func (r *ReconcileComplianceSuite) getAffectedMcfgPool(scan *compv1alpha1.ComplianceScan, mcfgpools *mcfgv1.MachineConfigPoolList) (*mcfgv1.MachineConfigPool, bool) {
```
jhrozek, line 595: Since we're already changing the signature of this function to (ptr, bool) why not drop the bool and check if the pointer is non-nil instead as a flag if the pool exists?

jhrozek (commented):
Nit about the signature of `getAffectedMcfgPool` otherwise LGTM

pkg/controller/compliancesuite/compliancesuite_controller.go
JAORMX: done

jhrozek (approved):
/lgtm

PR #630, opened 2021-04-26
Release v0.1.32

JAORMX:
* Set requests and limits for all workloads
* Update golang dependencies
* Set nodeSelector for most workloads to only run on master nodes
* Fix TailoredProfile validation issues
* Most workloads have appropriate annotations to be scheduled on management resources


jhrozek (approved):
/lgtm

PR #644, opened 2021-05-18
Add CHANGELOG.md and make changelog target

mrogers950:
Add an initial changelog and a make changelog target

### DIFF
--- a/utils/update_changelog.sh
+++ b/utils/update_changelog.sh
@@ -1,0 +1,17 @@
+#!/bin/bash -e
+# usage: ./update_changelog.sh <release number>
+release=$1
+cl=CHANGELOG.md
+today=$(date +%F)
+header=$(head -n7 ${cl})
+older=$(tail -n+7 ${cl})
+
+# get all commits on head since the latest tag
+changes=$(git log --no-merges --pretty=format:"- %s" $(git describe --tags --abbrev=0 HEAD^)..HEAD)
+
+echo "${header}" > ${cl}
+echo "" >> ${cl}
+echo "## [${release}] - ${today}" >> ${cl}
+echo "### Changes" >> ${cl}
+echo "${changes}" >> ${cl}
+echo "${older}" >> ${cl}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic
+Versioning](https://semver.org/spec/v2.0.0.html).
+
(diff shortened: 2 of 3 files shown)


jhrozek:
This looks great!
Do you think we could add a subtarget to make release that would grep Changelog.md for the current tag to make sure we didn't forget to add at least something?

mrogers950:
> This looks great!
> Do you think we could add a subtarget to make release that would grep Changelog.md for the current tag to make sure we didn't forget to add at least something?
Will do, I was also looking to update CHANGELOG.md automatically with git log via a target.

mrogers950:
The script I added for make release will result in a changelog update like this example release:

```
Author: Matt Rogers <mrogers@redhat.com>
Date:   Tue May 18 11:55:40 2021 -0400

	Release v0.2.30

diff --git a/CHANGELOG.md b/CHANGELOG.md
index b4853467..1958b344 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.30] - 2021-05-18
+### Changes
+- Changes to changelog
+- Add CHANGELOG.md and create-release Github Action
+- Update go.uber.org/zap dependency
+- Add permission to api-resource-collector to read MCs
+- Convert XML from CaC content to markdown in the k8s objects
+- Allow the api-resource collector to read ComplianceSuite objects
+- Die xmldom! Die!
+- Set the operators.openshift.io/infrastructure-features:proxy-aware annotation
+- Make use of the HTTPS_PROXY environment variable
+
 ## [0.1.32] - 2021-04-26
 ### Changes
 - Add Workload management annotations

```

jhrozek (approved):
looks good to me, if the openshift org allows running GH actions, feel free to merge

utils/update_changelog.sh
```diff
@@ -0,0 +1,17 @@
...
+release=$1
+cl=CHANGELOG.md
+today=$(date +%F)
+header=$(head -n7 ${cl})
+older=$(tail -n+7 ${cl})
+
+# get all commits on head since the latest tag
+changes=$(git log --no-merges --pretty=format:"- %s" $(git describe --tags --abbrev=0 HEAD^)..HEAD)
+
+echo "${header}" > ${cl}
```
JAORMX, line 12: This seems to overwrite the changelog. So... we're not able to keep the older entries then?

mrogers950: sorry the script is ugly (I bash program with a hammer), it does keep the older entries (see the echo of ${older} below)

JAORMX: let's shipit then! :smile:

PR #646, opened 2021-05-20
Allow api-resource-collector to read PrometheusRules and oauthclients

jhrozek:
This is needed to cover several FedRAMP rules.


JAORMX (approved):
/lgtm

PR #650, opened 2021-06-01
Remove unused const definitions

JAORMX:
There are a couple of unused const definitions that were left overs from
the migration off the xmldom library. This removes them.


jhrozek (approved):
/lgtm

PR #651, opened 2021-06-02
Release v0.1.34

JAORMX:
### Changes
- Switch to using go 1.16
- Remove unused const definitions
- Update dependencies
- RBAC: Allow api-resource-collector to list FIO objects

### DIFF
--- a/deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
@@ -159,13 +159,13 @@ metadata:
       ]
     capabilities: Seamless Upgrades
     categories: Monitoring,Security
-    olm.skipRange: '>=0.1.17 <0.1.33'
+    olm.skipRange: '>=0.1.17 <0.1.34'
     operatorframework.io/cluster-monitoring: "true"
     operatorframework.io/suggested-namespace: openshift-compliance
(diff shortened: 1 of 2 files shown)


deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
```diff
@@ -1406,8 +1407,8 @@ spec:
   maintainers:
   - email: support@redhat.com
     name: Red Hat Support
-  maturity: alpha
+  maturity: stable
```
jhrozek, line 1410: what is the reasoning behind this change? Can we claim maturity stable even if we ship alpha APIs?

JAORMX: uhm... good question! I honestly don't know. I consider the operator to be stable, and we've actually had quite stable APIs for a while. But you're right, maybe this is not appropriate cause we still have alpha CRDs.

JAORMX: Fixed

jhrozek: btw I couldn't find any explanation of what this flag means on operator-sdk homepage and oc explain just says it's a string..so I'm fine changing to stable since it just seems like an UI element. I filed https://issues.redhat.com/browse/CMP-986 so we can move our APIs beyond alpha.

jhrozek (approved):
/lgtm

PR #654, opened 2021-06-07
Collect all ocp-api-endpoint elements

jhrozek:
Some rules need to collect several resources and correlate them or run a
check on several results. For example OAuth token inactivity can be set
either in the OAuth server object or any of the OAuth clients.

This patch extends gathering the API resource URLs to all of those that
are included in the warnings elemement instead of the first one to
support cases like above.

Jira: [CMP-775](https://issues.redhat.com/browse/CMP-775)

### DIFF
--- a/cmd/manager/scap.go
+++ b/cmd/manager/scap.go
@@ -164,7 +164,7 @@ func (c *scapContentDataStream) FigureResources(profile string) error {
 //
 //  <warning category="general" lang="en-US"><code class="ocp-api-endpoint">/apis/config.openshift.io/v1/oauths/cluster
 //  </code></warning>
-func getPathFromWarningXML(in *xmlquery.Node) string {
+func getPathFromWarningXML(in *xmlquery.Node) []string {
 	DBG("Parsing warning %s", in.OutputXML(false))
 	return utils.GetPathFromWarningXML(in)
 }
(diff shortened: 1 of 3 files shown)


JAORMX:
> Some rules need to collect several resources and correlate them or run a
> check on several results. For example OAuth token inactivity can be set
> either in the OAuth server object or any of the OAuth clients.
> 
> This patch extends gathering the API resource URLs to all of those that
> are included in the warnings elemement instead of the first one to
> support cases like above.
> 
> Jira: [CMP-775](https://issues.redhat.com/browse/CMP-775)

Just to verify that I understood this correctly. The idea is to now have one warning that may contain several paths, right?

jhrozek:
> > Some rules need to collect several resources and correlate them or run a
> > check on several results. For example OAuth token inactivity can be set
> > either in the OAuth server object or any of the OAuth clients.
> > This patch extends gathering the API resource URLs to all of those that
> > are included in the warnings elemement instead of the first one to
> > support cases like above.
> > Jira: [CMP-775](https://issues.redhat.com/browse/CMP-775)
> 
> Just to verify that I understood this correctly. The idea is to now have one warning that may contain several paths, right?

Right, that's implemented with https://github.com/ComplianceAsCode/content/pull/7089

JAORMX:
/lgtm

@jhrozek thanks for the pointer!

PR #658, opened 2021-06-22
Enable kubernetes resource filtering using `jq` syntax

JAORMX:
This enables filtering of YAML objects using the jq syntax. This
effectively enables us to filter objects such as namespaces with more
complex metrics than we had before.

e.g. give me namespaces that don't start with "openshift" or "kube-"

This is done by relying on yet another `<code>` tag in the content,
which contains a class called "ocp-api-filter" which denotes a filter.
The aforementioned filter will also contain a unique ID as a class,
which will be an "id" tag in the original URL's code tag.

Finally, the plan is to use a URI fragment to differentiate the object
from other non-filtered ones.

Here's an example of the content:

```
<xccdf-1.2:warning xml:lang="en-US" category="general">This rule's check operates on the cluster configuration dump.
Therefore, you need to use a tool that can query the OCP API, retrieve the
<html:code class="ocp-api-endpoint" id="rule-id">/apis/fileintegrity.openshift.io/v1alpha1/fileintegrities?limit=5</html:code>
API endpoint, pipe it to the jq command with the following filter:
<html:code class="ocp-api-filter filter-rule-id">'.items[]'</html:code>,
and persist it to the local
<html:code class="ocp-dump-location" id="dump-rule-id"><xccdf-1.2:sub idref="xccdf_org.ssgproject.content_value_ocp_data_root" use="legacy"/>/apis/fileintegrity.openshift.io/v1alpha1/fileintegrities?limit=5#rule-id</html:code> file.
</xccdf-1.2:warning>
```

The dump location now has a `#rule-id` suffix (containing an actual ID), that differentiates it from
a regular dump. This way we are able to keep full objects separate from filtered ones.


JAORMX:
Pull-request updated, HEAD is now 97ae58ec497a2cca8c30004b4432f133a51c3a34

JAORMX:
Pull-request updated, HEAD is now 0bac9a2ffa99576efb0ed6cd2e881e57c17abdec

JAORMX:
Pull-request updated, HEAD is now 08fac5cce46739e20947f9a864fe811344408687

JAORMX:
Pull-request updated, HEAD is now 08fac5cce46739e20947f9a864fe811344408687

cmd/manager/scap.go
```diff
@@ -311,6 +323,34 @@ func fetch(client *kubernetes.Clientset, objects []utils.ResourcePath) (map[stri
...
+	}
+	obj := map[string]interface{}{}
+	unmarshallErr := json.Unmarshal(rawobj, &obj)
+	if unmarshallErr != nil {
+		return nil, fmt.Errorf("Error unmarshalling json: %w", unmarshallErr)
+	}
+	iter := fltr.RunWithContext(ctx, obj)
+	v, ok := iter.Next()
+	if !ok {
+		DBG("No result from filter, returning original object.")
```
jhrozek, line 339: This debug statement is confusing..btw I agree with not returning anything, just an error, it is important to fail hard when filtering fails.

cmd/manager/scap.go
```diff
@@ -311,6 +323,34 @@ func fetch(client *kubernetes.Clientset, objects []utils.ResourcePath) (map[stri
...
+	if fltrErr != nil {
+		return nil, fmt.Errorf("could not create filter '%s': %w", filter, fltrErr)
+	}
+	obj := map[string]interface{}{}
+	unmarshallErr := json.Unmarshal(rawobj, &obj)
+	if unmarshallErr != nil {
+		return nil, fmt.Errorf("Error unmarshalling json: %w", unmarshallErr)
+	}
+	iter := fltr.RunWithContext(ctx, obj)
+	v, ok := iter.Next()
```
jhrozek, line 337: So what about the case when the filter would return multiple values? Should we try running iter.Next() and fail if the result is not a single object? Because right now if the filter yields multiple objects we only return the first one (multiple objects is AFAIK something else than an array..)

jhrozek (commented):
Two nits inline, but in general I think this is really good!

cmd/manager/scap.go
JAORMX: That's right, we have the limitation that we can only support one output object and that's deliberate. Multiple objects won't lead to a parseable YAML file, and thus are quite useless to us in that state.

cmd/manager/scap.go
JAORMX: aha! The original idea was to not fail here. Indeed this debug message no longer applies. Let me fix it.

JAORMX:
Pull-request updated, HEAD is now 90935544b9c20be2955d7357f8da06a357040ac2

cmd/manager/scap.go
jhrozek: Right, that was my question, should we run another get from the iterator to assert that we only get one result? But feel free to add this in another PR or not at all of you disagree.

jhrozek (approved):
/lgtm

PR #661, opened 2021-06-28
This fixes the issue where a empty remediation yaml file is present

Vincent056:
Prevent creating ComplianceRemediations with empty remediation objects.

### DIFF
--- a/cmd/manager/aggregator.go
+++ b/cmd/manager/aggregator.go
@@ -378,7 +378,7 @@ func createResults(crClient *complianceCrClient, scan *compv1alpha1.ComplianceSc
 		if canCreate, why := canCreateRemediation(scan, remTargetObj); !canCreate {
 			// Only issue event once.
 			if !remediationNotPossibleEventIssued {
-				log.Info(why + " Remediation:" + crkey.Name)
+				log.Info(why, "Remediation", crkey.Name)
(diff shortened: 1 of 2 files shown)


cmd/manager/aggregator.go
```diff
@@ -373,8 +378,8 @@ func createResults(crClient *complianceCrClient, scan *compv1alpha1.ComplianceSc
 		if canCreate, why := canCreateRemediation(scan, remTargetObj); !canCreate {
 			// Only issue event once.
 			if !remediationNotPossibleEventIssued {
-				log.Info(why)
-				crClient.recorder.Event(scan, v1.EventTypeWarning, "CannotRemediate", why)
+				log.Info(why + " Remediation:" + crkey.Name)
```
JAORMX, line 381: Let's use the `"key": "value"` format to log this, this way it'll show up as JSON in the logs:

It would look like this:

```
log.Info(why. "Remediation", crkey.Name)
```

JAORMX (changes requested):
Great! And thanks for adding the unit test. The log and the test look great! So this PR is almost ready to merge.

I added one small comment about the log statement, and once that's fixed this is ready to go.

Thanks for the PR!

Vincent056 (commented):
That looks good!

JAORMX:
I'll wait for the tests to pass then merge this.

JAORMX (approved):

PR #662, opened 2021-06-28
Issue warning if filter issues more than one object

JAORMX:
We only really take one object into account when filtering results. This
can be tricky for content writes and we should ideally tell them
something is wrong. Thus, we issue a warning for such cases.


jhrozek (approved):
/lgtm
Thanks!

PR #665, opened 2021-07-01
The remediation templating design document

Vincent056:
It is still a work in progress, please let me know what you guys think, thanks!

### DIFF
--- a/doc/remediation-templating.md
+++ b/doc/remediation-templating.md
@@ -1,0 +1,386 @@
+# Remediation template proposed flow design
+This page describes the design and implementation of the remediation templating
+support in the compliance operator.
+
+## Goals
+We want to accomplish:
+   * The remediation content can use variables that SCAP supports instead of a pre-set 
+     fixed value
+   * During a scan where an administrator set custom variables in the tailored profile, it
+     is possible to create `ComplianceRemediation` objects based on the given custom values
+   * As a content writer, it is possible to use XCCDF value from rules in Kubernetes remediation
+   * Add additional remediation state that shows "Needs-Review" for not set and no default values 
+     remediation
+
+## High-level overview
+
+The compliance operator allows an administrator to extend a pre-existing profile, 
+to enable, disable rules and set values that come from the ProfileBundle. When 
+this happens, a CR `TailoredProfile` will be generated. Currently the `ComplianceScan`
+will use the tailored rules and variables from `TailoredProfile` CR, however,
+the `ComplianceRemediation` will now take custom values from `TailoredProfile`.
+Here we are going to implement that feature.
+
+### ComplianceAsCode remediation rule format
+
+
+In ComplianceAsCode, when the SCAP remediation content is generated, normally we will 
+have fixed remediation. In this proposal, we are going to change the remediation rule
+formate for the Kubernetes to include a marked reference variable to reference related
+XCCDF values from rules. 
+
+The following example is parts of two YAML files from ComplianceAsCode content, 
+the first one is the rule, and the second is the remediation:
+```yaml
+documentation_complete: true
+title: 'Set SSH Idle Timeout Interval'
+...
+requires:
+    - sshd_set_keepalive
+
+ocil_clause: 'it is commented out or not configured properly'
+
+ocil: |-
+    Run the following command to see what the timeout interval is:
+    <pre>$ sudo grep ClientAliveInterval /etc/ssh/sshd_config</pre>
+    If properly configured, the output should be:
+    <pre>ClientAliveInterval {{{ xccdf_value("sshd_idle_timeout_value") }}}</pre>
+...
+```
+As showing in the rule YAML file above, an XCCDF value for SSH Idle Timeout Interval is set.
+
+In the remediation YAML file, we can modify that Jinja Macro to pass the name of the 
+XCCDF variable and its value `sshd_idle_timeout_value`, and then the Macro will substitute the related 
+`ClientAliveInterval` values to `VAR_SSHD_IDLE_TIMEOUT_VALUE` as defined in the ocil:
+```yaml
+---
+# platform = multi_platform_ocp,multi_platform_rhcos
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+{{{ kubernetes_sshd_set("ClientAliveInterval","sshd_idle_timeout_value") }}}
+```
+
+And the machine config object data will be something like below before it gets url-encoded.
+```jinja
+...
+PrintLastLog yes
+#TCPKeepAlive yes
+PermitUserEnvironment no
+Compression no
+ClientAliveInterval 600
+ClientAliveCountMax VAR_SSHD_IDLE_TIMEOUT_VALUE
+
+#UseDNS no
+#PidFile /var/run/sshd.pid
+#MaxStartups 10:30:100
+#PermitTunnel no
+...
+```
+Since the url encode doesn't encode ASCII alpha-numerical nor underscore characters, and
+we can later substitute the place holder after the url encoding from `VAR_SSHD_IDLE_TIMEOUT_VALUE`
+to `{{sshd_idle_timeout_value | findXCCDFValue | urlquery}}`, so that it can be post-process them with the appropriate
+escaped instances in the compliance operator. The following code is the example of final result:
+```xml
+<fix rule="sshd_set_idle_timeout" complexity="low" disruption="low" reboot="false" strategy="restrict">---
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+spec:
+  config:
+    ignition:
+      version: 3.1.0
+    storage:
+      files:
+      - contents:
+          source: data:, `PrintLastLog%20yes%0A%23TCPKeepAlive%20yes%0APermitUserEnvironment%20no%0ACompression
+          %20no%0AClientAliveInterval%20{{"sshd_idle_timeout_value" | findXCCDFValue | urlquery}}%0AClientAliveCountMax%200%0A%0A%23UseDNS%2
+          0no%0A%23PidFile%20%2Fvar%2Frun%2Fsshd.pid%0A%23MaxStartups%2010%3A30%3A100%0A%23PermitTunnel%20no`
+        mode: 0600
+        path: /etc/ssh/sshd_config
+        overwrite: true
+</fix>
+```
+
+There will be some change in the aggregator since we brought a marker XCCDF variable into
+the remediation content, we need to substitute all the placeholder with corresponding values
+before we create the `ComplianceRemediation` object. we need to fetch all the variables that 
+are in the scan result `ConfigMap` for all the nodes with consistency check(Note that all the 
+tailored set values are parsed to the resulting CM), and then since we have XCCDF values marked 
+using `{{}}` in the remediation rules there is a Go Library called `Template` that we can 
+use to match, url-encode, and substitute the value with the XCCDF variable set in the Tailored Profile 
+from the result `ConfigMap` and then to create the `ComplianceRemediation` object with 
+corresponding values.
+
+Proof of concept using golang templating https://golang.org/pkg/text/template/#hdr-Functions 
+```go
+var value_dic = map[string]string{
+	"the_value_you_want_to_define":  "3600",
+	"the_value_you_want_to_define2": "1111",
+	"the_value_you_want_to_define3": "2222",
+	"the_value_you_want_to_define4": "3333",
+}
+var xccdf_variable = []string{} // list of variable was used to render the MachineConfig
+var missing_value = []string{} // list contain all missing value variables
+
+func findXCCDFValue(format string) string {
+	value, ok := value_dic[format]
+	if (ok && value != "xccdf-value-require-input") {
+    xccdf_variable = append(xccdf_variable, format) //append name of xccdf value used
+		return value
+	} else {
+    missing_value = append(missing_value, format) //append missing value variable name to the missing value list
+		return "Value-Not-Set"
+	}
+
+}
+
+func main() {
+	MachineConfigCM := "data:,...%0APrintMotd%20no%0A%0APrintLastLog%20yes%0A%23TCPKeepAlive%20yes%0APermitUserEnvironment%20no%0ACompression%20no%0AClientAliveInterval%20{{\"the_value_you_want_to_define65\" | findXCCDFValue | urlquery}}%0AClientAliveCountMax%200%0A%23UseDNS%20no%0A%23PidFile%20/var/run/sshd.pid%0A%23MaxStartups%2010%3A30%3A100%0A%23PermitTunnel%20no%0A%23ChrootDirectory%20none%0A%23..."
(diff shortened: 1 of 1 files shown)


JAORMX (changes requested):

doc/remediation-templating.md
```diff
@@ -0,0 +1,207 @@
...
+```
+#### Content writer user-flow
+
+As a content writer:
+It is now possible to write remediation content for kubenetes and use the XCCDF values from
+the rules file.
+
+Contiune with this rule as an example:
+
+```yaml
```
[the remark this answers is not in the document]
Vincent056: I have made some change to include the kube yaml for the remediation object. I will add more on the "what to put on the remediation when they want to take a variable into use." part.

doc/remediation-templating.md
```diff
@@ -0,0 +1,207 @@
...
+
+#UseDNS no
+#PidFile /var/run/sshd.pid
+#MaxStartups 10:30:100
+#PermitTunnel no
+...
+```
+Since the url encode doesn't encode ASCII alpha-numerical nor underscore characters, and
+we can later change the place holder after the url encoding from `VAR_SSHD_IDLE_TIMEOUT_VALUE`
+to `{{sshd_idle_timeout_value}}`, so that it can be post-process them with the appropriate
```
jhrozek, line 81: Can you please explicitly mention which component will perform this substitution?

doc/remediation-templating.md
```diff
@@ -0,0 +1,207 @@
...
+        mode: 0600
+        path: /etc/ssh/sshd_config
+        overwrite: true
+</fix>
+```
+
+There will be some change in the aggregator since we brought a marker XCCDF variable into
+the remediation content, we need to replace all the placeholder with corresponding values
+before we create the `ComplianceRemediation` object. we need to fetch all the variables that 
+are in the scan result `ConfigMap` for all the nodes with consistency check(Note that all the
```
jhrozek, line 106: The variable values also need to be url-encoded right?

jhrozek (commented):
Thank you very much for the writeup, I think it shows a very nice progress. In addition to the two small questions inline, I wonder if you have considered the case where the variable is not defined by the administrator and at the same time cannot have a reasonable default? In the CaC project, there are some rules for chronyd that need the administrator to specify a list of NTPD servers for example. Would it be possible to e.g. add another remediation state that would indicate to the administrator that some variables must be defined first?

mrogers950:
Looking good so far! I was wondering about the same case that Jakub mentions, where some remediations have to be treated as a  "needs-review" before being applied, and what might be a good way to expose that to the admin..

doc/remediation-templating.md
Vincent056: I am thinking about add a function in the Macro to do that substitution, before we create the remediation xml file.

doc/remediation-templating.md
Vincent056: Yes, the corresponding values needed to be url-encoded as well before we substitute to the XCCDF value marker.

Vincent056:
@jhrozek @mrogers950 @JAORMX Yes. I agree with you guys. We can add an additional status called “needs-review” for the remediation. So that will solve the situation where the remediation value is unspecified and doesn’t come with a default value. I will put change in the design document. Thanks for your reviews. I really appreciate your inputs.

doc/remediation-templating.md
JAORMX: Right, so from `VAR_SSHD_IDLE_TIMEOUT_VALUE` to `{{sshd_idle_timeout_value}}` it would be part of the content building. But it would be good to know where the substitution from `{{sshd_idle_timeout_value}}` to the actual value happens as well

doc/remediation-templating.md
JAORMX: This could be done via golang templating https://golang.org/pkg/text/template/#hdr-Functions so we could safely add some functions to the template and the Compliance Operator could evaluate them

doc/remediation-templating.md
```diff
@@ -0,0 +1,247 @@
...
+      kind: MachineConfig
+      spec:
+        config:
+          ignition:
+            version: 3.1.0
+          storage:
+            files:
+              - contents:
+                  source: >-
+                    data:,...%0APrintMotd%20no%0A%0APrintLastLog%20yes%0A%23TCPKeepAlive%20yes%0APermitUserEnvironment%20no%0ACompression%20no%0AClientAliveInterval%203600%0AClientAliveCountMax%200%0A%23UseDNS%20no%0A%23PidFile%20/var/run/sshd.pid%0A%23MaxStartups%2010%3A30%3A100%0A%23PermitTunnel%20no%0A%23ChrootDirectory%20none%0A%23...
```
JAORMX, line 239: I don't see the template here

doc/remediation-templating.md
Vincent056: Yes, so after we have the remediation xml data with `{{some-value-need-get-replaced}}` , in the aggregator, before we create the remediation, we need to look up  `some-value-need-get-replaced` against the result `config-map` and we can then replace the value with the `Set-Value` in result `config-map`

Vincent056 (commented):
{{"the_value_you_want_to_define" | findValue | urlquery}} can be potential format

doc/remediation-template.md
```diff
@@ -0,0 +1,342 @@
...
+	if err != nil {
+		panic(err)
+	}
+	t.Execute(os.Stdout, "nothinghere")
+}
+```
+
+
+In the case, where an administrator has to set a value for the remediation but did not set one,
+`ComplianceRemediation` status should be changed into `Needs-Review`
```
jhrozek: The remediation controller should then prevent the remediation from being applied so that nobody can wreck the cluster by applying a remediation that is missing some required values.

doc/remediation-template.md
```diff
@@ -0,0 +1,342 @@
...
+	"the_value_you_want_to_define3": "2222",
+	"the_value_you_want_to_define4": "3333",
+}
+
+func findValue(format string) string {
+	value, ok := value_dic[format]
+	if ok {
+		return value
+	} else {
+		return "NoValue"
```
jhrozek: I guess that during parsing you can just annotate the object with missing variable names and the remediation controller would then look if there are any unset variables and if yes, set the state to needs review or such?

jhrozek: btw since setting the variables can only be done in the tailored profile, at least we don't have to worry about re-rendering the remediation after the initial scan. We should probably output an event that tells the admin to define variables x,y,z and re-run the scan.

doc/remediation-template.md
Vincent056: I think that's a good point, we can add a check in remediation controller to do that. Also, I have a question on what values to put for the missing values remediation when we create them, do you have any recommendations?

Vincent056:
@JAORMX @mrogers950 @jhrozek  Sorry guys, I made some mistake while squash the commits, the correct squashed commits is this one: 
https://github.com/openshift/compliance-operator/pull/665/commits/3480b01500e99fd9bb58a8ae6fb95c9d606e5df9

jhrozek (commented):
Looking really good now! Just a couple of follow up notes.
Also, would you mind squash the commits to one? See e.g. [this post](https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec)
I usually just rebase with `git rebase -i origin/master`, aliased to `git rbi`

doc/remediation-templating.md
```diff
@@ -0,0 +1,342 @@
...
+		return value
+	} else {
+		return "NoValue"
+	}
+
+}
+
+func main() {
+	MachineConfigCM := "data:,...%0APrintMotd%20no%0A%0APrintLastLog%20yes%0A%23TCPKeepAlive%20yes%0APermitUserEnvironment%20no%0ACompression%20no%0AClientAliveInterval%20{{\"the_value_you_want_to_define65\" | findValue | urlquery}}%0AClientAliveCountMax%200%0A%23UseDNS%20no%0A%23PidFile%20/var/run/sshd.pid%0A%23MaxStartups%2010%3A30%3A100%0A%23PermitTunnel%20no%0A%23ChrootDirectory%20none%0A%23..."
+	t, err := template.New("text").Funcs(template.FuncMap{"findValue": findValue}).
```
JAORMX, line 136: I like the usage of FuncMap here! I hadn't thought about using this function but it makes a lot of sense!

JAORMX: I'd say this should have a more specific name, such as `findXCCDFValue` or something of the sort. @Vincent056 @jhrozek @mrogers950 thoughts on this?

doc/remediation-templating.md
```diff
@@ -0,0 +1,342 @@
...
+```yaml
+apiVersion: compliance.openshift.io/v1alpha1
+kind: ComplianceRemediation
+metadata:
+  selfLink: >-
+    /apis/compliance.openshift.io/v1alpha1/namespaces/openshift-compliance/complianceremediations/nist-moderate-modified-master-sshd-set-idle-timeout
+  resourceVersion: '35343'
+  name: nist-moderate-modified-master-sshd-set-idle-timeout
+  uid: 085f51e4-19bb-4478-9398-df409858dce6
+  creationTimestamp: '2021-07-01T19:06:43Z'
```
JAORMX, line 265: From a user perspective, it would be nice to indicate that a variable was used to render the `MachineConfig`. We could add an annotation or something of the sort to indicate that. What do you think?

JAORMX (changes requested):
Small suggestions, but overall I very much like the idea about templating support!

doc/remediation-template.md
jhrozek: I think you could take a similar approach as is currently used for dependencies between remediations, see e.g. how are labels like `RemediationDependenciesMetAnnotation` used. So you'd have an annotation whose values would be either all variables or just the missing ones and I guess it would also be useful to have a label (annotations are not indexed and cannot be used in searches, labels can) that would be set if some variables are missing so that people can list remediations with missing variable values.

Then on the controller side, if a remediation is to be applied and it has label/annotation set that says that a variable expansion is missing, refuse to apply it, see e.g. `handleUnmetDependencies` in the remediation controller now.

doc/remediation-templating.md
jhrozek: Yes, I was going back an forth between listing all variables or only those that are missing but because remediations are so hard to read we should probably annotate with all variables used. There won't be that many in most cases.

doc/remediation-templating.md
mrogers950: Yes, I agree with that.

Vincent056: I am going to change to that on the design doc

doc/remediation-templating.md
```diff
@@ -0,0 +1,342 @@
...
+
+#UseDNS no
+#PidFile /var/run/sshd.pid
+#MaxStartups 10:30:100
+#PermitTunnel no
+...
+```
+Since the url encode doesn't encode ASCII alpha-numerical nor underscore characters, and
+we can later substitute the place holder after the url encoding from `VAR_SSHD_IDLE_TIMEOUT_VALUE`
+to `{{sshd_idle_timeout_value | findValue | urlquery}}`, so that it can be post-process them with the appropriate
```
mrogers950, line 83: I don't follow the urlquery part of it. Is it for an already URL encoded value?

Vincent056: Since the `MachineConfig` object is url-encoded and the value for "sshd_idle_timeout_value" is not, we need to encoded that value as well before we substitute it into the `MachineConfig` object.

doc/remediation-templating.md
```diff
@@ -0,0 +1,342 @@
...
+
+In the case, where an administrator has to set a value for the remediation but did not set one,
+`ComplianceRemediation` status should be changed into `Needs-Review`
+
+### Proposed user-flow with examples
+
+#### Administrator user-flow
+Run the operator first, and then assume that an administrator wants
+to do a scan based on nist-moderate with a custom `sshd-idle-timeout-value` set to 
+`3600` instead of `300` as the default value. We will create a `TailoredProfile`:
```
mrogers950, line 154: The scenario that I am wondering about is if the content writer can leave an empty default (or to make things easier to parse, having an extra variable field such as "critical" or "review-needed"), making the remediation require that some value selection has to be made. 

It might be something that is reflected back to the user when a check tries to run initially, not necessarily just when the remediation is created. Then the admin sets the value, and runs the scan again.

doc/remediation-template.md
Vincent056: That's good point, we can add a `compliance.openshift.io/has-not-set-value-variable` label and `compliance.openshift.io/not-set-xccdf-values` , and when a user do a tailored scan, and did not set a value for the xccdf value that didn't come with a default value, we can put the missing variable to that label and annotation.

doc/remediation-templating.md
Vincent056: That's a interesting point, so will it be something for example, a rule that can only be used in a tailored profile since a value input is needed from the user. Maybe we can add a special value to the default selection so that in the operator we know that rule have to have a value set.

JAORMX: I like the idea. So, you'd detect that a variable is being used, check if the value of the variable is `require-input` or `xccdf-value-require-input` (let's make it verbose so there's no possibility for it to clash with something else) and no apply the remediation if that's the case?

mrogers950: A reserved value is cool with me.

doc/remediation-templating.md
mrogers950: Ah, I see now.

doc/remediation-templating.md
```diff
@@ -0,0 +1,352 @@
...
+ocil: |-
+    Run the following command to see what the timeout interval is:
+    <pre>$ sudo grep ClientAliveInterval /etc/ssh/sshd_config</pre>
+    If properly configured, the output should be:
+    <pre>ClientAliveInterval {{{ xccdf_value("sshd_idle_timeout_value") }}}</pre>
+...
+```
+As showing in the rule YAML file above, an XCCDF value for SSH Idle Timeout Interval is set.
+
+In the remediation YAML file, we can modify that Marco to pass the name of the
```
JAORMX, line 52: nit: s/Marco/macro/

doc/remediation-templating.md
JAORMX: could you change it? seems that it's still `findValue` as opposed to a more specific name

JAORMX (changes requested):
Almost there! Just missing some minor comments and the final comment about adding the variables used in a ComplianceRemediation as annotations so users can know they're being used.

doc/remediation-templating.md
Vincent056: Thanks for the correction, it is now ready for the review! I think we can define `missing_value = []string{}`, and add all the name of missing value variables here, and then to check if this missing_value is empty, if not we set label to the `missing_value` so that we know all the missing variables

JAORMX (approved):
I have no more comments on my side. This LGTM.

JAORMX:
@mrogers950 can you take a look and merge if you feel its ready?

mrogers950:
Looks great!

PR #666, opened 2021-07-01
tailoring: Update the tailoring CM on changes

jhrozek:
We used to only create the tailoring CM, not update it. This makes a
tight test loop harder as you'd need to remove the tailoredProfile every
time.

JIRA: [CMP-1008](https://issues.redhat.com/browse/CMP-1008)


JAORMX (approved):
/lgtm

PR #668, opened 2021-07-15
Fix instructions on building VMs

Vincent056:
I added some instructions on using VM to replace the vagrant VM setup


jhrozek (approved):
/lgtm
I think that using Vagrant by default is a good choice, this is targeting users who can't build the content on their local machine in the first place.

PR #670, opened 2021-07-21
Initial controller metrics

mrogers950:
This adds the plumbing and `compliance_operator_compliance_scan_status_total`

image


jhrozek:
Thanks for the PR. Except for looking at the code, do you have a pointer to some howto on metrics so that I can try out the code? Should I just magically see the metrics in the OCP UI once the new operator version is running and some scans have ran?

mrogers950:
> Thanks for the PR. Except for looking at the code, do you have a pointer to some howto on metrics so that I can try out the code? Should I just magically see the metrics in the OCP UI once the new operator version is running and some scans have ran?
That's right :) I can add to the PR/docs a mention that the operator sets up the monitoring service so it is scraped to the cluster monitoring dashboard automatically.

You can also check out the service directly like this (basically what we will use in ci)
```
$ oc run --rm -i --restart=Never --image=registry.fedoraproject.org/fedora-minimal:latest -n openshift-compliance test-metrics -- bash -c 'curl -ks -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" https://metrics.openshift-compliance.svc:8585/metrics-co' | grep compliance
If you don't see a command prompt, try pressing enter.
Error attaching, falling back to logs: unable to upgrade connection: container test-metrics not found in pod test-metrics_openshift-compliance
# HELP compliance_operator_compliance_scan_status_total A counter for the total number of updates to the status of a ComplianceScan
# TYPE compliance_operator_compliance_scan_status_total counter
compliance_operator_compliance_scan_status_total{name="ocp4-moderate",phase="AGGREGATING",result="NOT-AVAILABLE"} 3
compliance_operator_compliance_scan_status_total{name="ocp4-moderate",phase="DONE",result="NON-COMPLIANT"} 1
compliance_operator_compliance_scan_status_total{name="ocp4-moderate",phase="LAUNCHING",result="NOT-AVAILABLE"} 1
compliance_operator_compliance_scan_status_total{name="ocp4-moderate",phase="PENDING",result="NOT-AVAILABLE"} 1
compliance_operator_compliance_scan_status_total{name="ocp4-moderate",phase="RUNNING",result="NOT-AVAILABLE"} 1
compliance_operator_compliance_scan_status_total{name="rhcos4-moderate-master",phase="AGGREGATING",result="NOT-AVAILABLE"} 5
compliance_operator_compliance_scan_status_total{name="rhcos4-moderate-master",phase="DONE",result="NON-COMPLIANT"} 1
compliance_operator_compliance_scan_status_total{name="rhcos4-moderate-master",phase="LAUNCHING",result="NOT-AVAILABLE"} 1
compliance_operator_compliance_scan_status_total{name="rhcos4-moderate-master",phase="PENDING",result="NOT-AVAILABLE"} 1
compliance_operator_compliance_scan_status_total{name="rhcos4-moderate-master",phase="RUNNING",result="NOT-AVAILABLE"} 1
compliance_operator_compliance_scan_status_total{name="rhcos4-moderate-worker",phase="AGGREGATING",result="NOT-AVAILABLE"} 4
compliance_operator_compliance_scan_status_total{name="rhcos4-moderate-worker",phase="DONE",result="NON-COMPLIANT"} 1
compliance_operator_compliance_scan_status_total{name="rhcos4-moderate-worker",phase="LAUNCHING",result="NOT-AVAILABLE"} 1
compliance_operator_compliance_scan_status_total{name="rhcos4-moderate-worker",phase="PENDING",result="NOT-AVAILABLE"} 1
compliance_operator_compliance_scan_status_total{name="rhcos4-moderate-worker",phase="RUNNING",result="NOT-AVAILABLE"} 1
```

cmd/manager/operator.go
```diff
@@ -52,13 +57,18 @@ func init() {
...
 var (
-	metricsHost               = "0.0.0.0"
-	metricsPort         int32 = 8383
-	operatorMetricsPort int32 = 8686
-	defaultProducts           = []string{
+	metricsHost                 = "0.0.0.0"
+	metricsServiceName          = "metrics"
+	metricsPort           int32 = 8383
+	operatorMetricsPort   int32 = 8686
+	controllerMetricsPort int32 = 8585
```
jhrozek, line 64: This constant seems unused?

cmd/manager/operator.go
```diff
@@ -221,31 +241,16 @@ func addMetrics(ctx context.Context, cfg *rest.Config) {
...
-	}
-
-	// Create Service object to expose the metrics port(s).
-	service, err := metrics.CreateMetricsService(ctx, cfg, servicePorts)
+	// Fetch the metrics service created during the deployment.
+	metricsService, err := kClient.CoreV1().Services(operatorNs).Get(ctx, metricsServiceName, metav1.GetOptions{})
 	if err != nil {
-		log.Info("Could not create metrics Service", "error", err.Error())
+		log.Error(err, "Error getting metrics service")
+		os.Exit(1)
```
jhrozek, line 248: Does this need to be a hard error? (Both yes and no are fine; see my other comment about that error being unchecked when starting up the metrics controller..the only reason I'm commenting is that I want to understand which piece we consider critical and which optional)

deploy/role.yaml
```diff
@@ -1168,3 +1169,20 @@ rules:
...
+metadata:
+  name: prometheus-k8s
+  namespace: openshift-compliance
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - services
+      - endpoints
+      - pods
```
jhrozek, line 1184: Why does prometheus need to list pods (I understand services and endpoints)?

deploy/role_binding.yaml
```diff
@@ -110,3 +110,17 @@ roleRef:
...
+  name: prometheus-k8s
+  namespace: openshift-compliance
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: openshift-monitoring
```
jhrozek, line 126: Would this work also if openshift-monitoring was not installed?

deploy/service_account.yaml
```diff
@@ -32,3 +32,63 @@ apiVersion: v1
...
+kind: ClusterRoleBinding
+metadata:
+  name: co-metrics-client
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: co-metrics-client
+subjects:
+  - kind: ServiceAccount
+    name: default
```
jhrozek, line 57: Why are we using the default SA? Is it so that anybody could scrape the metrics?

deploy/service_account.yaml
```diff
@@ -32,3 +32,63 @@ apiVersion: v1
...
+    namespace: openshift-compliance
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: metrics-token
```
jhrozek, line 63: I don't see this secret being used anywhere

jhrozek (commented):
Mostly looks good -- I left a couple of questions inline. Please keep in mind that some of the questions come from my ignorance of metrics or prometheus :-)

jhrozek:
Oh btw the unit tests fail also for me locally. I bet it's the `met *metrics.Metrics` parameter being unset for the existing controllers.

cmd/manager/operator.go
mrogers950: Yeah, will fix

cmd/manager/operator.go
mrogers950: I'm on the fence about all metrics pieces being a hard requirement or not. On one hand, it will be annoying to the user to be put in the spot where they have to troubleshoot why metrics aren't showing, even though the operator is running. If we crash it is more obvious..

deploy/role.yaml
mrogers950: not sure exactly, I will investigate

deploy/role_binding.yaml
mrogers950: It's possible to create a rolebinding that references a nonexistant SA, if that's what you mean

jhrozek: yes, thanks

deploy/role.yaml
jhrozek: it's not that big of a deal since it's read-only -- I was just wondering

cmd/manager/operator.go
jhrozek: CO is (at least at the moment) an OCP-only operator and frankly I don't see it changing anytime soon. Personally I'm more in favor of requiring the critical functionality like metrics, provided the deps are there by default and just crash if something goes sideways.
We can always relax the requirement later if we get bug reports..

deploy/service_account.yaml
mrogers950: It's for the manual CLI testing with `oc run`. Running the workload to scrape the metrics would use default, and use the token specified below.

It's probably better to not have this permission set enabled on deployment, but let it be created by the CI harness for its use in CI.

pkg/controller/metrics/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	tlsConfig = libgocrypto.SecureTLSConfig(tlsConfig)
+	server := &http.Server{
+		Addr:      ":8585",
+		TLSConfig: tlsConfig,
+	}
+
+	err := server.ListenAndServeTLS("/var/run/secrets/serving-cert/tls.crt", "/var/run/secrets/serving-cert/tls.key")
+	if err != nil {
+		m.log.Error(err, "Metrics service failed")
```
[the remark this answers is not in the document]
mrogers950: will do

cmd/manager/operator.go
mrogers950: good point, i'll fix

mrogers950:
> Oh btw the unit tests fail also for me locally. I bet it's the `met *metrics.Metrics` parameter being unset for the existing controllers.

That's the reason. I think I'll have to add some stuff to the metrics fakeimpl pieces to work with those.

cmd/manager/operator.go
```diff
@@ -326,6 +330,53 @@ func ensureDefaultScanSettings(ctx context.Context, crclient client.Client, name
...
+	if err != nil {
+		return err
+	}
+	if !ok {
+		log.Info("Install prometheus-operator in your cluster to create ServiceMonitor objects")
+		return nil
+	}
+
+	serviceMonitor := metrics.GenerateServiceMonitor(service)
+	for i, _ := range serviceMonitor.Spec.Endpoints {
```
jhrozek, line 348: You can just omit the second loop variable instead of ignoring it with the blank assignment.

pkg/controller/metrics/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+	}).Inc()
+	if len(status.ErrorMessage) > 0 {
+		m.metrics.metricComplianceScanError.With(prometheus.Labels{
+			metricLabelScanName:  name,
+			metricLabelScanError: status.ErrorMessage,
+		}).Inc()
+	}
+}
+
+// IncComplianceScanStatus also increments error if necessary
```
jhrozek, line 151: copy pasta in the comment :-)

jhrozek (commented):
Looking good! Two nits, but I think the PR can be merged pretty much anytime. I'll reschedule CI which failed just because some installation error

cmd/manager/operator.go
mrogers950: fixed

pkg/controller/metrics/metrics.go
mrogers950: fixed

jhrozek (approved):
/lgtm

In general this version looks good. I think we can still do some small improvements (as you said, remove the default SA RBAC binding, maybe add some tests), but let's merge the PR and iterate on it.

PR #672, opened 2021-07-25
Add supoort for remediation templating for operator

Vincent056:
I think it is mostly done for the operator part, still need to do the consistent check, I have some questions on the value-required annotation that content developer use, since it will be already handled if a value is not found while parsing from result CM.


Vincent056:
I made some changes to templating format, I need to update the design doc. initially, in the proposal it was `{{the_value |findXCCDFValue| urlencode}}`, now it is `{{.the_value | urlencode}} `since it is a little tricky to pass multiple values to the pipeline, for the `findXCCDFValue` to find `the_value`, we need to give ResultCM to search that value.

Vincent056:
Please let me know what you guys think @mrogers950 @jhrozek, thanks.

pkg/utils/parse_arf_result.go
```diff
@@ -453,10 +484,66 @@ func remediationsFromString(scheme *runtime.Scheme, name string, namespace strin
...
+		return remContent, valuesUsedList, valuesMissingList, err //Error creating template
+	}
+	buf := &bytes.Buffer{}
+
+	errMsg := t.Execute(buf, resultValues) //Will return error if a value is not found in XCCDF
+	fixedText := buf.String()
+	if errMsg == nil {
+		valuesUsedList = getParsedValueName(t)
+		return fixedText, valuesUsedList, valuesMissingList, errMsg
+	} else if strings.Contains(errMsg.Error(), "map has no entry") {
```
jhrozek, line 511: this makes me a bit nervous..are there no error types returned from the template engine? Depending on a regex returned seems especially brittle..
In `getParsedValueName` you iterate the tree, would it be possible to get the missing values this way, too?

pkg/utils/parse_arf_result.go
```diff
@@ -453,10 +484,66 @@ func remediationsFromString(scheme *runtime.Scheme, name string, namespace strin
...
+		return fixedText, valuesUsedList, valuesMissingList, errMsg
+	}
+	return fixedText, valuesUsedList, valuesMissingList, errMsg
+}
+
+func getParsedValueName(t *template.Template) []string {
+	return ListTemplFields(t)
+}
+
+func ListTemplFields(t *template.Template) []string {
```
jhrozek, line 522: I guess this function doesn't have to be exported.

pkg/utils/parse_arf_result.go
```diff
@@ -389,11 +400,19 @@ func remediationFromFixElement(scheme *runtime.Scheme, fix *xmlquery.Node, scanN
...
+	return remediationsFromString(scheme, remName, namespace, fix.InnerText(), resultValues)
 }
 
-func remediationsFromString(scheme *runtime.Scheme, name string, namespace string, fixContent string) []*compv1alpha1.ComplianceRemediation {
-	objs, err := ReadObjectsFromYAML(strings.NewReader(fixContent))
+func remediationsFromString(scheme *runtime.Scheme, name string, namespace string, fixContent string, resultValues map[string]string) []*compv1alpha1.ComplianceRemediation {
+	//ToDO find and substitute the value
+	fixWithValue, valuesUsedList, notFoundValueList, parsingError := parseValues(fixContent, resultValues)
+
+	if (len(notFoundValueList) == 0) && (len(valuesUsedList) == 0) && parsingError != nil {
```
jhrozek, line 410: the braces seem to be extra since you're and-ing three conditions, but I wonder if there's a condition where you want to continue if parsingError is nil?

jhrozek (commented):
So far I scrolled through the code and left comments. Is there a test datastream to test the changes and see how the code works? I think we might want to add some test input to the unit tests as well..

pkg/apis/compliance/v1alpha1/complianceremediation_types.go
```diff
@@ -45,12 +46,16 @@ const (
...
 	OutdatedRemediationLabel               = "complianceoperator.openshift.io/outdated-remediation"
 	RemediationHasUnmetDependenciesLabel   = "compliance.openshift.io/has-unmet-dependencies"
+	RemediationUnsetValueLabel             = "compliance.openshift.io/has-unset-variable"
 	RemediationCreatedByOperatorAnnotation = "compliance.openshift.io/remediation"
 	RemediationDependencyAnnotation        = "compliance.openshift.io/depends-on"
 	RemediationObjectDependencyAnnotation  = "compliance.openshift.io/depends-on-obj"
 	RemediationDependenciesMetAnnotation   = "compliance.openshift.io/dependencies-met"
 	RemediationOptionalAnnotation          = "compliance.openshift.io/optional"
 	RemediationEnforcementTypeAnnotation   = "compliance.openshift.io/enforcement-type"
+	RemediationValueRequiredAnnotation     = "compliance.openshift.io/value-reqired"
```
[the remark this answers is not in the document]
Vincent056: > So far I scrolled through the code and left comments. Is there a test datastream to test the changes and see how the code works? I think we might want to add some test input to the unit tests as well..

I modified the test datastream to test out some different  situations, but I didn't put them on the PR, I will put them with some Ginkgo Tests.

pkg/utils/parse_arf_result.go
Vincent056: Actually, I was going to implement an error message, but I haven't decided what to put, as right now, if there is an error during any time at parsing, it will return nil and skip creating the remediaton,

pkg/utils/parse_arf_result.go
Vincent056: When the templating is not used then the parsingError will be nil, and It will continue, did I understand your question correctly?

pkg/utils/parse_arf_result.go
Vincent056: I will fix that

pkg/utils/parse_arf_result.go
```diff
@@ -506,6 +593,43 @@ func handleDependencyAnnotation(u *unstructured.Unstructured, annotations map[st
...
+			trimedValuesList = append(trimedValuesList, trimedValueMatch[0])
+		}
+	}
+	annotations[compv1alpha1.RemediationValueUsedAnnotation] = strings.Join(trimedValuesList, ",")
+	return annotations
+}
+
+func handleNotFoundValue(NotFoundValues []string, annotations map[string]string) map[string]string {
+
+	annotations[compv1alpha1.RemediationUnsetValueAnnotation] = strings.Join(NotFoundValues, ",")
```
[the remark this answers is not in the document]
Vincent056: No, it will not, only when a formated remediation is used, and there is no found value in the Result Config Map,
It is checked here, 

`if len(notFoundValueList) > 0 {
			annotations = handleNotFoundValue(notFoundValueList, annotations)
		}`

And it will only annotate the first not found value, since it is a parsing error, it will stop looking for values from the template. However, it will still show all used values, (the value that is going to be parsed will show in the value used)

pkg/utils/parse_arf_result.go
Vincent056: The `getParsedValueName` tree is before its execution, the the parsing will stop upon Notfound value, so the parsing tree will have all the values to be parsed.

Vincent056: I think I found a way around it, I am going to compare the Parsing values list with the found-value list to get the missing value list

Vincent056:
@jhrozek I have changed how I parse the value, now it will save all not-found-values to the annotation, as well as the values that were parsed, found, and used in the remediation. I also added a unit test for the parsing to check if the remediation will have correct annotations.

pkg/apis/compliance/v1alpha1/complianceremediation_types.go
```diff
@@ -45,12 +46,16 @@ const (
...
 	RemediationHasUnmetDependenciesLabel   = "compliance.openshift.io/has-unmet-dependencies"
+	RemediationUnsetValueLabel             = "compliance.openshift.io/has-unset-variable"
 	RemediationCreatedByOperatorAnnotation = "compliance.openshift.io/remediation"
 	RemediationDependencyAnnotation        = "compliance.openshift.io/depends-on"
 	RemediationObjectDependencyAnnotation  = "compliance.openshift.io/depends-on-obj"
 	RemediationDependenciesMetAnnotation   = "compliance.openshift.io/dependencies-met"
 	RemediationOptionalAnnotation          = "compliance.openshift.io/optional"
 	RemediationEnforcementTypeAnnotation   = "compliance.openshift.io/enforcement-type"
+	RemediationValueRequiredAnnotation     = "compliance.openshift.io/value-required"
+	RemediationUnsetValueAnnotation        = "compliance.openshift.io/unset-variable"
```
mrogers950, line 57: this might need to be `RemediationUnsetVariableAnnotation` to match the label

pkg/apis/compliance/v1alpha1/complianceremediation_types.go
```diff
@@ -185,6 +190,14 @@ func (r *ComplianceRemediation) HasUnmetDependencies() bool {
 	return (hasDependencies || hasObjDependencies) && !dependenciesMet
 }
 
+func (r *ComplianceRemediation) HasNotSetValues() bool {
```
mrogers950, line 193: I think for consistency sake this should be `HasUnsetValues`

pkg/utils/parse_arf_result.go
```diff
@@ -453,10 +484,95 @@ func remediationsFromString(scheme *runtime.Scheme, name string, namespace strin
 	return rems
 }
 
+func parseValues(remContent string, resultValues map[string]string) (string, []string, []string, error) {
```
mrogers950, line 487: Please add a leading comment for this function, particularly to describe the return values.

pkg/utils/parse_arf_result.go
```diff
@@ -453,10 +484,95 @@ func remediationsFromString(scheme *runtime.Scheme, name string, namespace strin
...
+
+	if err != nil {
+		return remContent, valuesUsedList, valuesMissingList, err //Error creating template
+	}
+	buf := &bytes.Buffer{}
+	err = t.Execute(buf, resultValues)
+	fixedText := buf.String()
+
+	if err != nil {
+		// Wrongly formated remediation context
```
mrogers950, line 502: Wondering if this comment is something that might be useful to include in the returned err.
Also, on this function can you remove the extra spacing?

pkg/utils/parse_arf_result.go
```diff
@@ -453,10 +484,95 @@ func remediationsFromString(scheme *runtime.Scheme, name string, namespace strin
...
+
+	valueToBeTrimmed := listNodeFields(t.Tree.Root, nil)
+
+	return trimToValue(valueToBeTrimmed)
+}
+
+//trim {{value | urlquery}} list to value list
+func trimToValue(listToBeTrimmed []string) []string {
+
+	trimedValuesList := listToBeTrimmed[:0]
```
mrogers950, line 536: typo in the resulting variable

pkg/utils/parse_arf_result.go
```diff
@@ -506,6 +622,37 @@ func handleDependencyAnnotation(u *unstructured.Unstructured, annotations map[st
...
+func handleValueUsed(valuesList []string, annotations map[string]string) map[string]string {
+
+	annotations[compv1alpha1.RemediationValueUsedAnnotation] = strings.Join(valuesList, ",")
+
+	return annotations
+}
+
+func handleNotFoundValue(notFoundValues []string, annotations map[string]string) map[string]string {
+
+	annotations[compv1alpha1.RemediationUnsetValueAnnotation] = strings.Join(notFoundValues, ",")
```
mrogers950, line 634: Oh, my earlier suggestion for changing RemediationUnsetVariableAnnotation might not be right, but instead change the annotation name to `unset-value`?

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -145,6 +145,15 @@ func (r *ReconcileComplianceRemediation) Reconcile(request reconcile.Request) (r
...
 
+	if remediationInstance.HasNotSetValues() {
+		res, valueErr := r.handleUnsetValues(remediationInstance, reqLogger)
+		if res.Requeue || valueErr != nil {
+			return res, valueErr
+		}
+	} else {
+		reconcileErr = r.reconcileRemediation(remediationInstance, reqLogger)
+	}
+
```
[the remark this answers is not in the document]
Vincent056: Hi Jakub, I have changed to this `if remediationInstance.Spec.Apply && remediationInstance.HasUnmetDependencies() {
		reqLogger.Info("Has unmet dependencies. Requeuing")
		return r.handleUnmetDependencies(remediationInstance, reqLogger)

	}

	if remediationInstance.Spec.Apply && remediationInstance.HasUnsetValues() {
		reqLogger.Info("Has unset values. Requeuing")
		return r.handleUnsetValues(remediationInstance, reqLogger)
	}`

Does this look ok to you? Thanks

Vincent056 (commented):
Hi @mrogers950 @jhrozek, I have addressed all the PR comments in the code, I am wondering if you guys can do an another review, Thanks

Vincent056:
For `RemediationValueRequiredAnnotation  = "compliance.openshift.io/value-required"`, I am thinking to let the content creator to use this annotation during creating the remediation, and in order for to check if the value was provided during a scan, We can check if the XCCDF variable value was found in the ResultCM and if it is equal to the default value. In order to pass `RemediationValueRequiredAnnotation` check the custom required variable's value has to be different from the default one. What do you guys think? This way, it won't break the existing scan since we still have a default value to use. During the scan, and only stop remediation from applying.

Vincent056:
> For `RemediationValueRequiredAnnotation = "compliance.openshift.io/value-required"`, I am thinking to let the content creator to use this annotation during creating the remediation, and in order for to check if the value was provided during creating the remediation, We can check if the `XCCDF variable `value was found in the `ResultCM` and if it is equal to the default value. We can make the requirement that In order to pass `RemediationValueRequiredAnnotation` check the custom required variable's value has to be different from the default one. What do you guys think? This way, it won't break the existing scan since we still have a default value to use. The scan will work with or without having a Custom Value, and it will only stop that missing required value annotation's remediation from applying. I haven't finished the code for this part yet, let me know if this idea sounds good to you guys! @mrogers950 @jhrozek

I decided to add check in the aggregator, first to check value-required annotation, and then get the tailored profile, to check if that custom value was set.

mrogers950:
@Vincent056 awesome, I'll review your latest changes soon. I'll probably need to check the PR out locally and see it run (not that familiar with remediation/variable code myself)

Vincent056:
> @Vincent056 awesome, I'll review your latest changes soon. I'll probably need to check the PR out locally and see it run (not that familiar with remediation/variable code myself)
Thanks for that! Let me know if you have any questions. By the way, I just noticed that we are using dash instead of underscore that is from the data-stream for variable name, should I can change '_' to '-' to match that?

mrogers950:
> > @Vincent056 awesome, I'll review your latest changes soon. I'll probably need to check the PR out locally and see it run (not that familiar with remediation/variable code myself)
> 
> Thanks for that! Let me know if you have any questions. By the way, I just noticed that we are using dash instead of underscore that is from the data-stream for variable name, should I can change '_' to '-' to match that?

Yes, that makes sense for consistency sake.

pkg/utils/parse_arf_result.go
```diff
@@ -15,17 +19,19 @@ import (
...
+	rulePrefix                      = "xccdf_org.ssgproject.content_rule_"
+	valuePrefix                     = "xccdf_org.ssgproject.content_value_"
+	questionnaireSuffix             = "_ocil:questionnaire:1"
+	questionSuffix                  = "_question:question:1"
+	dependencyAnnotationKey         = "complianceascode.io/depends-on"
+	kubeDependencyAnnotationKey     = "complianceascode.io/depends-on-obj"
+	remediationTypeAnnotationKey    = "complianceascode.io/remediation-type"
+	enforcementTypeAnnotationKey    = "complianceascode.io/enforcement-type"
+	optionalAnnotationKey           = "complianceascode.io/optional"
+	valueInputRequiredAnnotationKey = "complianceascode.io/value-input-reqired"
```
mrogers950, line 34: typo of `required` in the annotation string

pkg/utils/parse_arf_result.go
```diff
@@ -450,13 +484,82 @@ func remediationsFromString(scheme *runtime.Scheme, name string, namespace strin
...
+func parseValues(remContent string, resultValues map[string]string) (string, []string, []string, error) {
+	t, err := template.New("").Option("missingkey=zero").Parse(remContent)
+	var valuesUsedList []string
+	var valuesMissingList []string
+	var valuesParsedList []string
+	if err != nil {
+		fmt.Println(err)
+		errStr := "wrongly formated remediation context." + err.Error()
+		err = errors.New(errStr)
+		return remContent, valuesUsedList, valuesMissingList, err //Error creating template // Wrongly formated remediation context
```
mrogers950, line 502: Suggestion, this err block could probably be simplified quite a bit, i.e.,
```
if err != nil {
  return remContent, valuesUsedList, valuesMissingList, errors.New(fmt.Sprintf("error parsing remediation: %v", err))
}
```

pkg/utils/parse_arf_result.go
Vincent056: Hi, Matt, thanks for the review, I have corrected that, I also added a unit test to test the `complianceascode.io/value-input-reqired`.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -369,6 +379,51 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
...
+			}
+		}
+		return reconcile.Result{}, nil
+	}
+	return reconcile.Result{}, nil
+}
+
+func (r *ReconcileComplianceRemediation) countXCCDFUnsetValues(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) (int, error) {
+	var nNotSetValues int
+	not_set_values := rem.Annotations[compv1alpha1.RemediationUnsetValueAnnotation]
```
mrogers950, line 418: Rather than use underscores for a similarly named variable, it would be better to have different names, (such as numberUnset and unsetValues)

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -369,6 +379,51 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
...
+	}
+
+	rCopy := rem.DeepCopy()
+	labels := rCopy.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+
+	if nNotSetValues > 0 {
+		if _, ok := labels[compv1alpha1.RemediationUnsetValueLabel]; !ok {
```
mrogers950, line 403: In this case that remediations are parsed and have unset variables initially; suppose then the admin sets the variable and applies the remediation, would this label and RemediationUnsetValueAnnotation persist?

Vincent056: This one is tricky, if the admin did set a variable value later on, should we trigger a rescan instead for that specify rule/might be more than one rule, however in order to do that we have to first associate all the variable with rules , and then a new remediation will be generated based on the new variable.

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: Yeah, that's a good idea

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -369,6 +379,51 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
...
+	}
+	return reconcile.Result{}, nil
+}
+
+func (r *ReconcileComplianceRemediation) countXCCDFUnsetValues(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) (int, error) {
+	var nNotSetValues int
+	not_set_values := rem.Annotations[compv1alpha1.RemediationUnsetValueAnnotation]
+
+	for _, not_set_values := range strings.Split(not_set_values, ",") {
+		logger.Info("Remediation has unset values, cannot apply", "ComplianceRemediation.Name", rem.Name)
```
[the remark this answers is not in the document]
Vincent056: Sound good, I will change that.

pkg/controller/complianceremediation/complianceremediation_controller.go
mrogers950: Hmm, I don't think we should worry about triggering a rescan right away when the variable is set, but I want to let Jakub or Ozz weigh in on the handling of this part. My thinking of the user flow would be:
1. Scan + remediate attempt
2. Discover one or more remediation needs attention, from remediation status
3. Look at remediation and find the required variable, and set the value
4. re-run scan manually (or wait for the next scheduled scan)
The result of the second scan is that the remediations are updated with the variable setting.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -369,6 +381,115 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
 	return nMissingDeps, nil
 }
 
+func (r *ReconcileComplianceRemediation) handleValueRequired(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) (reconcile.Result, error) {
+	rCopy := rem.DeepCopy()
```
mrogers950, line 385: Just a point of optimization wrt DeepCopy; it would be better to make the DeepCopy closer to the call to Update. Ideally under the same condition block, so a no-op run of the function won't make the copy.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -369,6 +381,115 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
...
+		}
+		return false, nil
+	}
+	//scan did not have a tailored profile
+	return false, nil
+}
+
+func (r *ReconcileComplianceRemediation) handleUnsetValues(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) (reconcile.Result, error) {
+	nNotSetValues := r.countXCCDFUnsetValues(rem, logger)
+	rCopy := rem.DeepCopy()
```
mrogers950, line 463: Same advice wrt DeepCopy as above

JAORMX:
@mrogers950 @Vincent056 what's the status of this? Is this ready for review?

pkg/apis/compliance/v1alpha1/complianceremediation_types.go
```diff
@@ -185,6 +191,37 @@ func (r *ComplianceRemediation) HasUnmetDependencies() bool {
...
+}
+
+func (r *ComplianceRemediation) HasValueUsed() bool {
+
+	a := r.GetAnnotations()
+	if len(a) == 0 {
+		return false
+	}
+	_, HasValueUsed := a[RemediationValueUsedAnnotation]
+	return HasValueUsed
```
JAORMX, line 221: There is quite a bit of repetition on these functions now. At this point it would be worth having a common one that takes the annotation to be checked and returns true or false if the annotation is set or not.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -369,6 +381,116 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
 	return nMissingDeps, nil
 }
 
+func (r *ReconcileComplianceRemediation) handleValueRequired(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) (reconcile.Result, error) {
```
JAORMX, line 384: I'd like to see a unit test for this function, that would ensure this works as expected and help us reviewers understand better what this does.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -369,6 +381,116 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
...
+				return true, nil
+			}
+		}
+		return false, nil
+	}
+	//scan did not have a tailored profile
+	return false, nil
+}
+
+func (r *ReconcileComplianceRemediation) handleUnsetValues(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) (reconcile.Result, error) {
```
JAORMX, line 461: I'd like to see a unit test for this function, that would ensure this works as expected and help us reviewers understand better what this does.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -369,6 +381,116 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
...
 }
 
+func (r *ReconcileComplianceRemediation) handleValueRequired(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) (reconcile.Result, error) {
+	annotations := rem.GetAnnotations()
+	labels := rem.GetLabels()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	requiredValues := annotations[compv1alpha1.RemediationValueRequiredAnnotation]
+	requiredValuesList := strings.Split(requiredValues, ",")
```
JAORMX, line 391: How do we handle empty strings here? Sounds like something a user may set accidentally.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -136,11 +137,22 @@ func (r *ReconcileComplianceRemediation) Reconcile(request reconcile.Request) (r
 	}
 
 	var reconcileErr error
-	if remediationInstance.HasUnmetDependencies() {
+
+	if remediationInstance.HasUnmetDependencies() || remediationInstance.HasUnsetValues() || remediationInstance.HasValueRequired() {
```
JAORMX, line 141: moving these to individual `if` branches would help with readability. e.g.

```
if remediationInstance.HasUnmetDependencies() {
...
} else if remediationInstance.HasUnsetValues() {
...
} else if remediationInstance.HasValueRequired() {
...
} else {
...
```

This will also help ensure that we don't depend on things set on `remediationInstance` which we shouldn't be modifying directly.

pkg/controller/complianceremediation/complianceremediation_controller.go
JAORMX: As far as the user flow is concerned, the user would need to modify the TailoredProfile in order to set the variable. While most variables are already set by default, this case is for variables that are empty and thus need setting. @Vincent056 perhaps in the PR description you could add an example of one of these variables that need setting? Anyway, the workflow that you described sounds correct.

JAORMX (changes requested):

Vincent056 (commented):
> @mrogers950 @Vincent056 what's the status of this? Is this ready for review?

Welcome back! yes, the code is ready for the review!

pkg/apis/compliance/v1alpha1/complianceremediation_types.go
Vincent056: That's a good advice! So can it be something like this?

`func (r *ComplianceRemediation) HasAnnotation(ann string) bool {

	a := r.GetAnnotations()
	if len(a) == 0 {
		return false
	}
	_, hasAnnotation := a[ann]
	return hasAnnotation

}`

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: Sounds good! I will add a unit test for this.

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: Maybe I should handle the empty string during the parsing?

pkg/apis/compliance/v1alpha1/complianceremediation_types.go
JAORMX: looks good

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: I think that's correct, there is one more case, when there is an value-required annotation, and the user did not set values for those value-required variables, it will also trigger that Need-Review Status.  I will add examples for that.

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: I will add a unit test for that!

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: This will skip the `if remediationInstance.HasValueRequired()` when there is unset-values

Vincent056: I mean it will skip the next block if the previous ones are true, so the other blocks doesn't get handled

Vincent056: I think I found a way around

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: It has been addressed, I will skip all the empty string during the looping

Vincent056:
Hi @JAORMX, I have addressed those comments and ready for another review.

pkg/controller/complianceremediation/complianceremediation_controller.go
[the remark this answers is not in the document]
jhrozek: Maybe the context around this (old) comment has changed, but why don't we use:
```
if remediationInstance.HasUnmetDependencies() {
 ...
} else if remediationInstance.HasAnnotation(compv1alpha1.RemediationUnsetValueAnnotation) {
 ...
} else if remediationInstance.HasAnnotation(compv1alpha1.RemediationValueRequiredAnnotation) {
 ...
} else {
 reconcileErr = r.reconcileRemediation(remediationInstance, reqLogger)
}
```
?

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -378,6 +395,123 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
...
+		}
+		return false, nil
+	}
+	//scan did not have a tailored profile
+	return false, nil
+}
+
+func (r *ReconcileComplianceRemediation) handleUnsetValues(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) (bool, reconcile.Result, error) {
+	nNotSetValues := r.countXCCDFUnsetValues(rem, logger)
+	hasUpdate := false
```
jhrozek, line 483: Do we really need the `hasUpdate` variable here? it seems like there's only one place where it's set to true and its value is returned right afterwards, could we get away with instead using a `true` or `false` constant?

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -378,6 +395,123 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
...
+				return true, nil
+			}
+		}
+		return false, nil
+	}
+	//scan did not have a tailored profile
+	return false, nil
+}
+
+func (r *ReconcileComplianceRemediation) handleUnsetValues(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) (bool, reconcile.Result, error) {
```
jhrozek, line 481: Do we need to return the `reconcile.Result` variable from this function? it seems like it's set to `reconcile.Result{}` in all branches?
Note that the reason that `handleUnmetDependencies` returns the `reconcile.Result` type is that in some branches, that function returns `reconcile.Result{Requeue: true, RequeueAfter: defaultDependencyRequeueTime}`, IOW, reconcile, but with a delay.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -378,6 +395,123 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
...
+			}
+			hasUpdate = true
+			return hasUpdate, reconcile.Result{}, nil
+		}
+		return hasUpdate, reconcile.Result{}, nil
+	}
+	return hasUpdate, reconcile.Result{}, nil
+}
+
+func (r *ReconcileComplianceRemediation) countXCCDFUnsetValues(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) int {
```
jhrozek, line 507: logger is unused in this function. I guess it also doesn't have to me a method of `r` but it's not harmful either and we haven't been super consistent with how we use functions and methods.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -378,6 +395,123 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
...
+	notSetValues := rem.Annotations[compv1alpha1.RemediationUnsetValueAnnotation]
+	if notSetValues == "" {
+		return 0
+	}
+	return len(notSetValues)
```
jhrozek, line 512: did you really want to return the length of the annotation or maybe split it by `,` ?

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -378,6 +395,123 @@ func (r *ReconcileComplianceRemediation) countKubeUnmetDependencies(rem *compv1a
 	return nMissingDeps, nil
 }
 
+func (r *ReconcileComplianceRemediation) handleValueRequired(rem *compv1alpha1.ComplianceRemediation, logger logr.Logger) (bool, reconcile.Result, error) {
+	annotations := rem.GetAnnotations()
+	labels := rem.GetLabels()
+	hasUpdate := false
```
jhrozek, line 401: Same comment about potentially not needing the `hasUpdate` variable and not having to return the `reconcile.Result` as I had with `handleUnsetValues`

pkg/utils/parse_arf_result.go
```diff
@@ -450,13 +488,79 @@ func remediationsFromString(scheme *runtime.Scheme, name string, namespace strin
...
+//This function will take orginal remediation content, and a list of all values found in the configMap
+//It will processed and substitue the value in remediation content, and return processed Remediation content
+//The return will be Processed-Remdiation Content, Value-Used List, Un-Set List, and err if possible
+func parseValues(remContent string, resultValues map[string]string) (string, []string, []string, error) {
+	t, err := template.New("").Option("missingkey=zero").Parse(remContent)
+	var valuesUsedList []string
+	var valuesMissingList []string
+	var valuesParsedList []string
+	if err != nil {
+		return remContent, valuesUsedList, valuesMissingList, errors.New("wrongly formatted remediation context: " + err.Error()) //Error creating template // Wrongly formatted remediation context
```
jhrozek, line 503: Wouldn't it be nicer to use `errors.Wrap` instead of concatenating error strings?

pkg/utils/parse_arf_result.go
```diff
@@ -450,13 +488,79 @@ func remediationsFromString(scheme *runtime.Scheme, name string, namespace strin
...
+	}
+	buf := &bytes.Buffer{}
+	err = t.Execute(buf, resultValues)
+	fixedText := buf.String()
```
jhrozek, line 507: fixedText appears to be used only in the error handling, maybe move it there?

jhrozek (commented):
I left a couple of comments inline, starting to look really good!

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: That's a good idea, I will fix that

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: Thanks for pointing out! Yeah I was meant to return length split by `,`

pkg/controller/complianceremediation/complianceremediation_controller.go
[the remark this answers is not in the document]
Vincent056: Because there are situations where we have both unset-values annotation and value required Annotation,  and since the second will always be true, it will never reach the third block.

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: I am going to fix it

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: That's a good catch, we can remove that return as well!

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: I think we can remove that logger here since it is not used

pkg/utils/parse_arf_result.go
Vincent056: Sounds good, I will do that.

Vincent056:
Hey @jhrozek, thank you very much for the review, I have addressed you comments in the last commit.

pkg/utils/parse_arf_result.go
Vincent056: Actually I think `fixedText` is used for other returns as well. the `fixedText` here in the error handling block is just a placeholder. The `fixedText` is parsed remediation context with value instead of variable name from datastream.

JAORMX:
The e2e test failed with the following...

```
     helpers.go:258: expected all remediations are unapplied when scan finishes 
```
Seems related... but let's retry and see if it's replicable

JAORMX (approved):
This LGTM. @mrogers950 @jhrozek please take a look as well

pkg/utils/parse_arf_result.go
jhrozek: sorry, I missed it, you're right

jhrozek (commented):
I'll refrain from clicking the Approve buton so that the PR doesn't get merged yet in case prow decides, but I'm happy with the current version.
ACK.

PR #674, opened 2021-07-28
doc: note that rolling back images in ProfileBundle is not well supported

jhrozek:
Let's just document that this is the case.

Jira: [CMP-962](https://issues.redhat.com/browse/CMP-962)


jhrozek:
Does not need BZ, this is just docs. Merging.

PR #677, opened 2021-08-02
Add a must-gather plugin

mrogers950:
I was looking for something in the vanilla must-gather and realized we needed support - so here it is.
@Vincent056 can you review?


Vincent056:
I will do that, thanks

Vincent056 (commented):
I have reviewed the code, and I think it all makes sense to me!

PR #681, opened 2021-08-11
Release v0.1.38

mrogers950:
### Changes
- e2e: aggregating/NA metric value
- Bug 1990836: Move metrics service creation back into operator startup
- Add fetch-git-tags make target
- Add a must-gather plugin


jhrozek (approved):
/lgtm

PR #684, opened 2021-08-20
Add permission for operator to remediate prometheusrule objects

JAORMX:
This is needed for the AU-5 control which requires an alert that looks
for audit log errors.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -174,6 +174,17 @@ rules:
   - watch
   - update
   - patch
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - prometheusrules
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - create
+  - patch
 # Enforcement types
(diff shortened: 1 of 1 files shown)


JAORMX:
Pull-request updated, HEAD is now 6a0e67d146fc70208b6abb06b72f6abd2079ab7d

deploy/role.yaml
```diff
@@ -174,6 +174,17 @@ rules:
...
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - prometheusrules
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - create
```
jhrozek, line 186: Why create and update though? Are we going to create an automated remediation?

JAORMX: Yes. See https://github.com/ComplianceAsCode/content/pull/7427

jhrozek (approved):

PR #687, opened 2021-08-25
Make default scanTolerations more tolerant

gsleeman:
To allow scan scheduling by default for worker and master nodes with non-default taints

### DIFF
--- a/deploy/crds/compliance.openshift.io_compliancescans_crd.yaml
+++ b/deploy/crds/compliance.openshift.io_compliancescans_crd.yaml
@@ -130,13 +130,11 @@ spec:
                 type: string
               scanTolerations:
                 default:
-                - effect: NoSchedule
-                  key: node-role.kubernetes.io/master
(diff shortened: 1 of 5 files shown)


JAORMX (commented):
Thanks for the PR! We'll check this out

JAORMX:
@gsleeman to move this forward, let's comment (for now) the test case `TestNodeSchedulingErrorFailsTheScan` from the e2e tests. You can find it here: https://github.com/openshift/compliance-operator/blob/master/tests/e2e/e2e_test.go#L1830

Basically, that test tries to verify that if you can't schedule a scan on all relevant nodes, the scan should fail... We're actually thinking about changing that functionality given that this may not be the wanted behavior in some scenarios (e.g. when having more ephemeral nodes). So... let's comment it for now and I'll work on turning it back on once we come up with a solution for the aforementioned case.

JAORMX (approved):
/lgtm
/approve

PR #688, opened 2021-08-26
Bug 1969620: compliancescan: Fill the <target> element and the urn:xccdf:fact:identifier for node checks

jhrozek:
The <target> element ought to contain the hostname of the system being 
scanned, but for the node scan results produced by oscap-chroot, it was
just saying unknown. This is because openscap normally reads /etc/hostname, 
which is not present on RHCOS.

While there is a bug filed against openscap
(https://bugzilla.redhat.com/show_bug.cgi?id=1977668) to read alternative
sources, let's work around that issue in the meantime by overriding the
<target> element ourselves during scan results post-processing.

Also, because the node host name is not that useful outside baremetal 
on-prem deployments, let's also put the node name into the <fact> with
"id=urn:xccdf:fact:identifier". This is done by setting the 
OSCAP_EVALUATION_TARGET environment variable. But in order to do so, we 
need to stop using oscap-chroot and start using oscap directly with the 
OSCAP_PROBE_ROOT variable set to "chroot://" ourselves because the 
oscap-chroot wrapper script also sets OSCAP_EVALUATION_TARGET to a value on
its own.

Platform check results are not changed in any way, because it is not 
possible to find the cluster name internally, without having the client 
context (see e.g. https://github.com/kubernetes/kubernetes/issues/44954)

Related: [OCPBUGSM-30237](https://issues.redhat.com/browse/OCPBUGSM-30237)

### DIFF
--- a/pkg/controller/compliancescan/config.go
+++ b/pkg/controller/compliancescan/config.go
@@ -77,8 +77,9 @@ if [ -z $HOSTROOT ]; then
 		oscap xccdf eval \
 	)
 else
+	export OSCAP_PROBE_ROOT="$HOSTROOT"
 	cmd=(
-		oscap-chroot $HOSTROOT xccdf eval \
+		oscap xccdf eval \
 	)
 fi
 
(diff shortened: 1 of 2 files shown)


jhrozek:
> @jhrozek thanks for checking into this... I think it's not worth the effort as you put it. We might be introducing too much tech debt for something that might not bring as much value. Is this good to go as-is?

I think yes, the only question is about tests, but I could add them in a separate PR as well. The intent being that the scan produces the expected number of CMs and a PV with some files.

PR #692, opened 2021-08-27
Restart profileparser on failures

jhrozek:
We used to be too permisive about failures in the profileparser, which
worked fine in the beginning, but doesn't work so well since we started
annotating the objects with the bundle SHA digest during an update and delete
any objects that were not annotated with that digest value - in case
an update operation failed because of a transient error, we would have
skipped annotating that object and ultimately treat it as removed and
delete it. This was happening especially during upgrades where the API
server might be unreachable at some point.

This patch exits the profileparser with a non-zero code on failures as
well as propagates the errors from the client operations to the error
channels so that they are eventually returned to the binary and used for
the error code. This ensures that the profileparser restarts and
hopefully goes through after the restart.

Jira: [OCPBUGSM-32701](https://issues.redhat.com/browse/OCPBUGSM-32701)

### DIFF
--- a/cmd/manager/profileparser.go
+++ b/cmd/manager/profileparser.go
@@ -151,6 +151,11 @@ func runProfileParser(cmd *cobra.Command, args []string) {
 	// to valid
 	updateProfileBundleStatus(pcfg, pb, err)
 
+	if err != nil {
+		log.Error(err, "Parsing the bundle failed, will restart the container")
+		os.Exit(1)
+	}
+
 	if closeErr := contentFile.Close(); closeErr != nil {
 		log.Error(err, "Couldn't close the content file")
 	}
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -399,6 +399,100 @@ func TestE2E(t *testing.T) {
 			},
 		},
 		testExecution{
+			Name:       "TestParsingErrorRestartsParserInitContainer",
+			IsParallel: true,
+			TestFn: func(t *testing.T, f *framework.Framework, ctx *framework.Context, mcTctx *mcTestCtx, namespace string) error {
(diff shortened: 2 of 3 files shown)


cmd/manager/profileparser.go
```diff
@@ -154,4 +154,9 @@ func runProfileParser(cmd *cobra.Command, args []string) {
 	if closeErr := contentFile.Close(); closeErr != nil {
 		log.Error(err, "Couldn't close the content file")
 	}
+
+	if err != nil {
```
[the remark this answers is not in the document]
jhrozek: Done.

jhrozek:
> uhm... the only way I can think of testing this is by having a container image with invalid content.

Right, let's see if I can craft one. Putting this on hold.
/hold

jhrozek:
test added, no longer holding

tests/e2e/e2e_test.go
```diff
@@ -398,6 +398,100 @@ func TestE2E(t *testing.T) {
...
+				}
+
+				if err := waitForProfileBundleStatus(t, f, namespace, pbName, compv1alpha1.DataStreamInvalid); err != nil {
+					return err
+				}
+
+				// list the pods with profilebundle=pbName
+				var lastErr error
+				timeouterr := wait.Poll(retryInterval, timeout, func() (bool, error) {
+					podList := &v1.PodList{}
```
JAORMX, line 434: s/v1/corev1/

JAORMX (changes requested):

tests/e2e/e2e_test.go
jhrozek: oops, IDE optimization fail

JAORMX (approved):
/lgtm

PR #693, opened 2021-08-31
Associate variable with compliance check result

Vincent056:
Added a new attribute `ValuesUsed` into `ComplianceCheckResult` that contains a string slice with all the associate settable variables for a rule

### DIFF
--- a/cmd/manager/aggregator.go
+++ b/cmd/manager/aggregator.go
@@ -322,6 +322,10 @@ func getCheckResultLabels(pr *utils.ParseResult, resultLabels map[string]string,
 	labels[compv1alpha1.SuiteLabel] = scan.Labels[compv1alpha1.SuiteLabel]
 	labels[compv1alpha1.ComplianceCheckResultStatusLabel] = string(pr.CheckResult.Status)
 	labels[compv1alpha1.ComplianceCheckResultSeverityLabel] = string(pr.CheckResult.Severity)
+	if len(pr.CheckResult.ValuesUsed) > 0 {
+		labels[compv1alpha1.ComplianceCheckResultValueLabel] = ""
+	}
+
 	if pr.Remediations != nil {
 		labels[compv1alpha1.ComplianceCheckResultHasRemediation] = ""
 	}
@@ -336,7 +340,6 @@ func getCheckResultLabels(pr *utils.ParseResult, resultLabels map[string]string,
 func getCheckResultAnnotations(cr *compv1alpha1.ComplianceCheckResult, resultAnnotations map[string]string) map[string]string {
 	annotations := make(map[string]string)
 	annotations[compv1alpha1.ComplianceCheckResultRuleAnnotation] = cr.IDToDNSFriendlyName()
-
 	for k, v := range resultAnnotations {
 		annotations[k] = v
 	}
--- a/doc/crds.md
+++ b/doc/crds.md
@@ -251,6 +251,8 @@ Where:
       properly.
 	* **NOTAPPLICABLE**: Which indicates that the check didn't run because it is not
       applicable or not selected.
+ * **valuesUsed**: a list of settable variables associated with the rule scan result,
+  a user can set these variables in a tailored profile.
 
 This object is owned by the scan that created it, as seen in the
 `ownerReferences` field.
--- a/pkg/apis/compliance/v1alpha1/compliancecheckresult_types.go
+++ b/pkg/apis/compliance/v1alpha1/compliancecheckresult_types.go
@@ -13,6 +13,7 @@ type ComplianceCheckStatus string
 // way.
 const ComplianceCheckResultStatusLabel = "compliance.openshift.io/check-status"
 const ComplianceCheckResultSeverityLabel = "compliance.openshift.io/check-severity"
+const ComplianceCheckResultValueLabel = "compliance.openshift.io/check-has-value"
 
 // ComplianceCheckResultLabel defines a label that will be included in the
 // ComplianceCheckResult objects. It indicates whether the result has an automated
@@ -92,6 +93,8 @@ type ComplianceCheckResult struct {
 	// Any warnings that the user should be aware about.
 	// +nullable
 	Warnings []string `json:"warnings,omitempty"`
+	// It stores a list of values used by the check
+	ValuesUsed []string `json:"valuesUsed,omitempty"`
 }
 
 // IDToDNSFriendlyName gets the ID from the scan and returns a DNS
--- a/pkg/utils/parse_arf_result.go
+++ b/pkg/utils/parse_arf_result.go
@@ -6,6 +6,7 @@ import (
 	"html/template"
 	"io"
 	"regexp"
+	"sort"
 	"strings"
 	"text/template/parse"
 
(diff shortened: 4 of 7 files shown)


Vincent056:
@jhrozek I went with the safe path, I noticed there are more than one states in `oval:states` for certain oval tests, and I also found out there are some variables in oval:objects that have used variables. What I did is to generate a hash map map each  oval test to all its associate variables, by going through each test's objects and states. And then during `complianceCheckResult` creation, I have a function to lookup value from the generated hash map.

Vincent056:
@jhrozek @mrogers950 @JAORMX Hi guys, can you guys do a review when you free up? Thanks!

pkg/apis/compliance/v1alpha1/compliancecheckresult_types.go
```diff
@@ -85,7 +86,8 @@ type ComplianceCheckResult struct {
 	// The severity of a check status
 	Severity ComplianceCheckResultSeverity `json:"severity"`
 	// A human-readable check description, what and why it does
-	Description string `json:"description,omitempty"`
+	ValueUsed   []string `json:"valuesUsed,omitempty"`
```
JAORMX, line 89: Given that this is a list, I suggest using `ValuesUsed`, denoting the possibility of several values.

pkg/apis/compliance/v1alpha1/compliancecheckresult_types.go
```diff
@@ -85,7 +86,8 @@ type ComplianceCheckResult struct {
...
+	Description string   `json:"description,omitempty"`
```
JAORMX, line 90: the comment for the description got lost and is now on top of `ValueUsed`, let's keep it on top of description as that comment is used to generate the CRD.

JAORMX (changes requested):
I don't see the CRD changes, seems that you're yet to run `make generate`

cmd/manager/aggregator.go
```diff
@@ -322,6 +322,10 @@ func getCheckResultLabels(pr *utils.ParseResult, resultLabels map[string]string,
 	labels[compv1alpha1.SuiteLabel] = scan.Labels[compv1alpha1.SuiteLabel]
 	labels[compv1alpha1.ComplianceCheckResultStatusLabel] = string(pr.CheckResult.Status)
 	labels[compv1alpha1.ComplianceCheckResultSeverityLabel] = string(pr.CheckResult.Severity)
+	if len(pr.CheckResult.ValueUsed) > 0 {
+		labels[compv1alpha1.ComplianceCheckResultValueLabel] = ""
+	}
```
JAORMX, line 327: I'm not sure a label is needed for this if we're adding the `ValuesUsed` key to the CRD itself. Is there any specific use-case in mind for adding this label?

cmd/manager/aggregator.go
```diff
@@ -336,7 +340,9 @@ func getCheckResultLabels(pr *utils.ParseResult, resultLabels map[string]string,
 func getCheckResultAnnotations(cr *compv1alpha1.ComplianceCheckResult, resultAnnotations map[string]string) map[string]string {
 	annotations := make(map[string]string)
 	annotations[compv1alpha1.ComplianceCheckResultRuleAnnotation] = cr.IDToDNSFriendlyName()
-
+	if len(cr.ValueUsed) > 0 {
+		annotations[compv1alpha1.ComplianceCheckResultValueAnnotation] = strings.Join(cr.ValueUsed, ",")
+	}
```
JAORMX, line 345: I'm not sure annotation is needed for this if we're adding the ValuesUsed key to the CRD itself. Is there any specific use-case in mind for adding this annotation? If we want this information to be on the metadata somehow, I would suggest skipping the label and sticking only to the annotation. But then again, I still think it's enough if we have it on the `ValuesUsed` key on the CRD.

pkg/utils/parse_arf_result.go
```diff
@@ -126,7 +134,152 @@ func newRuleHashTable(dsDom *xmlquery.Node) nodeByIdHashTable {
 func newOcilQuestionTable(dsDom *xmlquery.Node) nodeByIdHashTable {
 	return newHashTableFromRootAndQuery(dsDom, "//ds:component/ocil:ocil", "//ocil:boolean_question")
 }
+func newStateHashTable(dsDom *xmlquery.Node) nodeByIdHashTable {
+	return newHashTableFromRootAndQuery(dsDom, "//ds:component/oval-def:oval_definitions/oval-def:states", "*")
+}
+
+func newObjHashTable(dsDom *xmlquery.Node) nodeByIdHashTable {
+	return newHashTableFromRootAndQuery(dsDom, "//ds:component/oval-def:oval_definitions/oval-def:objects", "*")
+}
```
JAORMX, line 143: uhm... let's stay put to the e2e tests, I have the feeling these three operations will be very memory intensive.

Adding a big table of sates, objects and values from a 9Mb (as is the RCHOS4 datastream) file might result in high memory usage. If this is the case, we might want to increase our memory limits for the aggregator pod. I'm fine with using more memory since this is good for performance, and this is a one-shot operation (you don't do compliance checks every minute).

pkg/utils/parse_arf_result_test.go
```diff
@@ -166,6 +166,54 @@ var _ = Describe("XCCDF parser", func() {
...
+						break
+					}
+				}
+			})
+
+			It("Should have the expected value List", func() {
+				Expect(len(check.ValueUsed)).To(Equal(1))
+				Expect(check.ValueUsed[0]).To(Equal(expValue))
+			})
+		})
```
JAORMX, line 214: Let's do several tests, one for a rule where you know there's no values used, then one that has several values (more than one). I think that would be enough coverage.

JAORMX (changes requested):

cmd/manager/aggregator.go
Vincent056: That's a good idea, I was testing if it has worked since the `ValuesUsed` did not show up, I will deleted it!

pkg/utils/parse_arf_result_test.go
Vincent056: I have added two more tests, one with 2 variables and one without any

pkg/utils/parse_arf_result.go
Vincent056: Sounds good! I added one more table to lookup test for rules, the previous method had some corner cases, now it is fixed

pkg/apis/compliance/v1alpha1/compliancecheckresult_types.go
Vincent056: I have fixed that and added comments to the `ValuesUsed` for descriptions

cmd/manager/aggregator.go
Vincent056: I have changed the `ValueUsed` to `ValuesUsed`, and deleted the annotations, I am going to keep the label so that it will be easier to search if a `ComplianceCheckResult` has used variables

JAORMX:
@Vincent056 the failures seem consistent, perhaps it's now time to investigate.

JAORMX:
```
        --- FAIL: TestE2E/Parallel_tests/TestSuiteWithContentThatDoesNotMatch (220.03s)
        --- FAIL: TestE2E/Serial_tests/TestSuiteScan (105.04s)
        --- FAIL: TestE2E/Serial_tests/TestPlatformAndNodeSuiteScan (110.04s) 
```

JAORMX (changes requested):
Let's add relevant documentation too

Vincent056:
@JAORMX before we set the `ValuesUsed`, I checked each variable against generated `set-values` hash table from `ResultConfigMap`

JAORMX:
> @JAORMX before we set the `ValuesUsed`, I checked each variable against generated `set-values` hash table from `ResultConfigMap`, and the rule with two variables test was removed, because I couldn’t find a rule with two variable for moderated after only includes settable variables.

Got it, makes sense. I think code-wise, this is good to go. Let's just add the docs for the new entry in the CRD and we can merge this.

doc/crds.md
```diff
@@ -251,6 +251,8 @@ Where:
       properly.
 	* **NOTAPPLICABLE**: Which indicates that the check didn't run because it is not
       applicable or not selected.
+ * **valueUsed**: a list of settable varaibles associated with the rule scan result,
+  a user can set these varaibles in a tailord profile.
```
JAORMX, line 255: s/valuesed/valuesUsed/
s/varaibles/variables/g
s/tailord/tailored/

JAORMX (changes requested):

doc/crds.md
Vincent056: Haha thanks for pointing out the typo

JAORMX (approved):
/lgtm
/approve

PR #694, opened 2021-09-02
Bug 1999374: Use ClusterRole/ClusterRoleBinding for monitoring permissions

mrogers950:
OLM doesn't support deploying Roles/RoleBindings that reference SAs outside of
the operator namespace. To allow for out-of-the-box monitoring when deploying
from OLM, this changes the permission objects to a ClusterRole and
ClusterRoleBinding.

Also:

- Add a `test-catalog` make target for local-cluster testing of the index.
- Add documentation for the current prometheus metrics.


mrogers950:
@JAORMX @jhrozek Some updates after more CSV testing:

Placing the rolebinding directly into manifests/ allows OLM to deploy the rolebinding without stepping on the subject namespace. This is good. Unfortunately, it also means that OLM does not update the reference to the correct role (which is created dynamically by OLM).

```
$ oc get rolebinding/prometheus-k8s-monitoring -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  creationTimestamp: "2021-09-07T18:33:22Z"
  name: prometheus-k8s-monitoring
  namespace: openshift-compliance
  resourceVersion: "70279"
  uid: a9e2de08-0ce5-4208-a1b5-f00d7c2c086b
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: prometheus-k8s
subjects:
- kind: ServiceAccount
  name: prometheus-k8s
  namespace: openshift-monitoring

$ oc get role/prometheus-k8s
Error from server (NotFound): roles.rbac.authorization.k8s.io "prometheus-k8s" not found
```

The OLM deployed role is instead:
```
$ oc get role
NAME                                                            CREATED AT
compliance-operator.v0.1.39                                     2021-09-07T18:33:21Z
compliance-operator.v0.1.39-api-resource-collector-5bfdbcc6d9   2021-09-07T18:33:23Z
compliance-operator.v0.1.39-compliance-operator-748c4bc79b      2021-09-07T18:33:25Z
compliance-operator.v0.1.39-profileparser-74c8c5d475            2021-09-07T18:33:24Z
compliance-operator.v0.1.39-prometheus-k8s-ff6bfd85c            2021-09-07T18:33:24Z  <----
compliance-operator.v0.1.39-remediation-aggregator-565566f444   2021-09-07T18:33:23Z
compliance-operator.v0.1.39-rerunner-6f85c6c95                  2021-09-07T18:33:24Z
compliance-operator.v0.1.39-resultscollector-5476bdc8bc         2021-09-07T18:33:25Z
compliance-operator.v0.1.39-resultserver-5f6bdc4687             2021-09-07T18:33:23Z
```

It's not possible to patch a rolebinding's roleRef, so the fix from this point is to manually create the Role under the expected name (I added a README blurb for this).

jhrozek:
> @JAORMX @jhrozek Some updates after more CSV testing:

[...]

> It's not possible to patch a rolebinding's roleRef, so the fix from this point is to manually create the Role under the expected name (I added a README blurb for this).

I think this is fine at least for now, but I wonder if we're the only ones with this problem. Have you asked the operator-sdk/OLM developers if others have had the same problem as well?

README.md
```diff
@@ -353,3 +353,59 @@ or support is available at `quay.io/compliance-operator/must-gather:latest`:
...
+
+The cluster-monitoring logs will display an error for the operator namespace:
+```
+$ oc -n openshift-monitoring logs -c prometheus prometheus-k8s-0 | grep openshift-compliance
+...
+level=error ts=2021-09-07T19:09:08.598Z caller=klog.go:96 component=k8s_client_runtime func=ErrorDepth msg="github.com/prometheus/prometheus/discovery/kubernetes/kubernetes.go:429: Failed to watch *v1.Endpoints: failed to list *v1.Endpoints: endpoints is forbidden: User \"system:serviceaccount:openshift-monitoring:prometheus-k8s\" cannot list resource \"endpoints\" in API group \"\" in the namespace \"openshift-compliance\": RBAC: role.rbac.authorization.k8s.io \"prometheus-k8s\" not found"
+```
+
+To fix this, as the cluster-admin, create the prometheus-k8s Role inside of the operator namespace:
+```
```
jhrozek, line 392: So we can't put the Role under `deploy/` because it would have been picked up into the bundle, right? I'm just wondering if we could put the Role into a manifest under SCM control and instead point to `oc create` invocation which might be a bit more user friendly and maybe even possible to use in a test later on?

mrogers950:
Updated again. With using a ClusterRole/ClusterRoleBinding permission set, it works without requiring a manual step (this is the same approach cluster-logging uses until OLM can support RoleBindings in the way we need)

jhrozek (approved):
/lgtm
thank you, I think this is much better for usability

PR #697, opened 2021-09-06
e2e: Migrate TestScanProducesRemediations to use ScanSettingBinding

JAORMX:
This migrates the test to use ScanSettingBindings and TailoredProfiles,
which is what we recommend users do. It also stops using the
`rhcos4-moderate` profile which is very time and resource intensive, and
instead uses the `ocp4-moderate` which is quite complete but lighter.

The profile change also fixes a fluke. The moderate profile doesn't
produce all remediations with the "applicable" state, as some need
dependencies. So we couldn't even do this test properly with that
profile anyway. So the profile change is needed.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -451,46 +451,62 @@ func TestE2E(t *testing.T) {
 			},
 		},
 		testExecution{
-			Name: "TestScanProducesRemediations",
-			// NOTE(jaosorior): This was made a serial test because it runs the long-running, resource-taking and
-			// big AF moderate profile
-			IsParallel: false,
+			Name:       "TestScanProducesRemediations",
+			IsParallel: true,
 			TestFn: func(t *testing.T, f *framework.Framework, ctx *framework.Context, mcTctx *mcTestCtx, namespace string) error {
-				scanName := getObjNameFromTest(t)
-				selectWorkers := map[string]string{
-					"node-role.kubernetes.io/worker": "",
-				}
-				testScan := &compv1alpha1.ComplianceScan{
+				bindingName := getObjNameFromTest(t)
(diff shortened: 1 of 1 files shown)


tests/e2e/e2e_test.go
```diff
@@ -451,46 +451,59 @@ func TestE2E(t *testing.T) {
...
+			Name:       "TestScanProducesRemediations",
+			IsParallel: true,
 			TestFn: func(t *testing.T, f *framework.Framework, ctx *framework.Context, mcTctx *mcTestCtx, namespace string) error {
-				scanName := getObjNameFromTest(t)
-				selectWorkers := map[string]string{
-					"node-role.kubernetes.io/worker": "",
+				bindingName := getObjNameFromTest(t)
+				tpName := getObjNameFromTest(t)
+
+				tp := &compv1alpha1.TailoredProfile{
```
jhrozek, line 460: Can you add a comment explaining why we're adding a TP that only extends and doesn't enable or disable any rules?

jhrozek: (And remind me why we're doing that, I know you told me some time ago but I forgot...)

jhrozek (commented):
One nit inline, otherwise looks good

JAORMX:
Pull-request updated, HEAD is now 5c1b2fc9b324a08178d0db3c8ed11512329cfaed

tests/e2e/e2e_test.go
JAORMX: @jhrozek done. Mostly it's just us wanting the scans to be unique and for this not to re-use names. This is not a problem in usual deployments, but mostly for our test cases.

jhrozek (approved):
/lgtm

PR #698, opened 2021-09-07
Remove Variable Selection check for tailor profile 

Vincent056:
To enable the use of custom variable for remediation templating, we need to remove the Variable Selection check so that user can use any values for variables doing both scan and remediation


jhrozek (commented):
The code looks good, but please amend the commit message `git commit --amend` to reflect what the PR really does.

Vincent056:
> The code looks good, but please amend the commit message `git commit --amend` to reflect what the PR really does.

Thanks for that! I have correct the commit message

jhrozek (approved):
/lgtm

PR #699, opened 2021-09-08
Disallow empty titles and descriptions for tailored profiles

JAORMX:
These settings are actually required by the tailored profile, so let's
instead validate them appropriately instead of silently failing.


jhrozek (approved):
/lgtm

PR #702, opened 2021-09-14
Modify openscap build in Makefile

Vincent056:
Added openscap build to e2e-ci, deploy-local and image-to-cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,7 @@ else ifeq ($(RUNTIME), docker)
 endif
 
 # Temporary
-OPENSCAP_DEFAULT_IMAGE_TAG=1.3.4
-RELATED_IMAGE_OPENSCAP_TAG?=$(OPENSCAP_DEFAULT_IMAGE_TAG)
+RELATED_IMAGE_OPENSCAP_TAG?=latest
 
 # Image path to use. Set this if you want to use a specific path for building
 # or your e2e tests. This is overwritten if we bulid the image and push it to
@@ -137,7 +136,7 @@ operator-image:
 
 .PHONY: openscap-image
 openscap-image:
-	$(RUNTIME) build --no-cache -t $(RELATED_IMAGE_OPENSCAP_PATH):$(TAG) $(OPENSCAP_DOCKER_CONTEXT)
+	$(RUNTIME) build --no-cache -t $(RELATED_IMAGE_OPENSCAP_PATH):$(RELATED_IMAGE_OPENSCAP_TAG) $(OPENSCAP_DOCKER_CONTEXT)
 
 .PHONY: bundle-image
 bundle-image:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -244,13 +244,13 @@ test-benchmark: ## Run the benchmark tests -- Note that this can only be ran for
 e2e: namespace tear-down operator-sdk image-to-cluster openshift-user deploy-crds ## Run the end-to-end tests
 	@echo "WARNING: This will temporarily modify deploy/operator.yaml"
 	@echo "Replacing workload references in deploy/operator.yaml"
-	@$(SED) 's%$(IMAGE_REPO)/$(RELATED_IMAGE_OPENSCAP_NAME):$(OPENSCAP_DEFAULT_IMAGE_TAG)%$(RELATED_IMAGE_OPENSCAP_PATH):$(RELATED_IMAGE_OPENSCAP_TAG)%' deploy/operator.yaml
+	@$(SED) 's%$(IMAGE_REPO)/$(RELATED_IMAGE_OPENSCAP_NAME):$(OPENSCAP_DEFAULT_IMAGE_TAG)%$(RELATED_IMAGE_OPENSCAP_PATH):$(TAG)%' deploy/operator.yaml
```
mrogers950, line 247: Instead of switching to TAG here and in the `cluster-image-push` target, let's change the `openscap-image` target to use RELATED_IMAGE_OPENSCAP_TAG. 

We don't update the openscap image very frequently and when we rebuild it will still be 1.3.4, so we can keep using the existing tag.

Makefile
```diff
@@ -333,8 +337,10 @@ deploy: namespace deploy-crds ## Deploy the operator from the manifests in the d
 .PHONY: deploy-local
 deploy-local: namespace image-to-cluster deploy-crds ## Deploy the operator from the manifests in the deploy/ directory and the images from a local build
 	$(SED) 's%$(IMAGE_REPO)/$(APP_NAME):latest%$(RELATED_IMAGE_OPERATOR_PATH)%' deploy/operator.yaml
+	$(SED) 's%$(IMAGE_REPO)/$(RELATED_IMAGE_OPENSCAP_NAME):$(RELATED_IMAGE_OPENSCAP_TAG)%$(RELATED_IMAGE_OPENSCAP_PATH)%' deploy/operator.yaml
```
mrogers950, line 340: I think the tag referenced here will be incorrect, TAG is "latest" and RELATED_IMAGE_OPENSCAP_TAG is "1.3.4". So let's use RELATED_IMAGE_OPENSCAP_TAG all around.

mrogers950: Just a note: We will need a push target for the openscap image to run before the next upstream release.

Makefile
```diff
@@ -290,6 +290,8 @@ ifdef IMAGE_FORMAT
 image-to-cluster:
 	@echo "IMAGE_FORMAT variable detected. We're in a CI enviornment."
 	@echo "We're in a CI environment, skipping image-to-cluster target."
+	$(eval component = $(RELATED_IMAGE_OPENSCAP_NAME))
+	$(eval RELATED_IMAGE_OPENSCAP_PATH = $(IMAGE_FORMAT))
```
mrogers950, line 294: FYI, If https://github.com/openshift/compliance-operator/pull/704 merges before this, IMAGE_FORMAT will need to be IMAGE_FROM_CI

Vincent056: Let's merge #704 first, then I can fix this PR

Makefile
Vincent056: Sounds Good!

Makefile
Vincent056: I used tag, because in existing openscap_build, 
```
.PHONY: openscap-image
openscap-image:
	$(RUNTIME) build --no-cache -t $(RELATED_IMAGE_OPENSCAP_PATH):$(TAG) $(OPENSCAP_DOCKER_CONTEXT)
```

JAORMX:
@Vincent056 retesting won't work yet. This https://github.com/openshift/release/pull/21934 needs to merge first. It was delayed cause I messed up the configuration in the first pass of that PR.

Vincent056:
> @Vincent056 retesting won't work yet. This [openshift/release#21934](https://github.com/openshift/release/pull/21934) needs to merge first. It was delayed cause I messed up the configuration in the first pass of that PR.

Haha, thanks for letting me know

Makefile
```diff
@@ -26,7 +26,7 @@ endif
 
 # Temporary
 OPENSCAP_DEFAULT_IMAGE_TAG=1.3.4
-RELATED_IMAGE_OPENSCAP_TAG?=$(OPENSCAP_DEFAULT_IMAGE_TAG)
+RELATED_IMAGE_OPENSCAP_TAG?=latest
```
JAORMX, line 29: I think we should stop using `OPENSCAP_DEFAULT_IMAGE_TAG` and just stick to the variable `RELATED_IMAGE_OPENSCAP_TAG`.

Vincent056: We are using 1.3.4 at https://github.com/openshift/compliance-operator/blob/50cbfab8678e1164fdbea713c8f57303612f2311/deploy/operator.yaml#L48, should we change here as well then?

PR #703, opened 2021-09-14
Add instructions and check type to Rule object

JAORMX:
This enables folks browsing the rules to know what type of check will a
specific rule do and how to manually audit it without having to run a
scan.

### DIFF
--- a/deploy/crds/compliance.openshift.io_rules_crd.yaml
+++ b/deploy/crds/compliance.openshift.io_rules_crd.yaml
@@ -43,11 +43,18 @@ spec:
             nullable: true
             type: array
             x-kubernetes-list-type: atomic
+          checkType:
+            description: 'What type of check will this rule execute: Platform, Node
+              or none (represented by an empty string)'
+            type: string
           description:
             description: The description of the Rule
             type: string
           id:
             description: The XCCDF ID
+            type: string
+          instructions:
+            description: Instructions for auditing this specific rule
             type: string
           kind:
             description: 'Kind is a string value representing the REST resource this
--- a/doc/crds.md
+++ b/doc/crds.md
@@ -511,39 +511,42 @@ of rules. An example `Rule` object looks like this:
 
 ```yaml
 apiVersion: compliance.openshift.io/v1alpha1
-description: '<br /><br /><pre>$ sudo nmcli radio wifi off</pre>Configure the system
-  to disable all wireless network interfaces with the&#xA;following command:'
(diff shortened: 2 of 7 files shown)


JAORMX:
Pull-request updated, HEAD is now 1fa6af5e520d1dc993ab2ee8e3ea74b5fecdd9a0

JAORMX:
Pull-request updated, HEAD is now fb8e071364be2562b2619477f2fc5ba57f057e2b

jhrozek:
I think this PR is good and can be merged, but I wonder if you considered using a different approach for the checkType - what if we instead looked at the list of enabled rules under `xccdf-1.2:Profile` and used the profile's `<xccdf-1.2:platform idref="cpe:/o:redhat:openshift_container_platform_node:4"/>` to determine if the check is platform or node? I guess we could even use a combination of the two.

JAORMX:
> I think this PR is good and can be merged, but I wonder if you considered using a different approach for the checkType - what if we instead looked at the list of enabled rules under `xccdf-1.2:Profile` and used the profile's `<xccdf-1.2:platform idref="cpe:/o:redhat:openshift_container_platform_node:4"/>` to determine if the check is platform or node? I guess we could even use a combination of the two.

@jhrozek I thought about doing that, but then figured that the approach wouldn't encompass all rules that would potentially be part of the data stream. There might be rules that we (for x or y reason) take out of a profile, but still would like to have customers be able to use. So, we wouldn't be able to accurately detect the type for those. By instead using the custom warning we use to detect the API path, we can know if it's node or Platform and encompass the case I mentioned.

jhrozek:
> > I think this PR is good and can be merged, but I wonder if you considered using a different approach for the checkType - what if we instead looked at the list of enabled rules under `xccdf-1.2:Profile` and used the profile's `<xccdf-1.2:platform idref="cpe:/o:redhat:openshift_container_platform_node:4"/>` to determine if the check is platform or node? I guess we could even use a combination of the two.
> 
> @jhrozek I thought about doing that, but then figured that the approach wouldn't encompass all rules that would potentially be part of the data stream. There might be rules that we (for x or y reason) take out of a profile, but still would like to have customers be able to use. So, we wouldn't be able to accurately detect the type for those. By instead using the custom warning we use to detect the API path, we can know if it's node or Platform and encompass the case I mentioned.

Ah, good points. I didn't think about that, then your approach is better.

jhrozek (approved):
/lgtm

PR #706, opened 2021-09-15
e2e: Mark logging functions as helper functions

JAORMX:
This ensures that we get the relevant log lines in our test output when
using the logging functions


jhrozek (approved):
/lgtm

PR #711, opened 2021-09-17
Add flag to skip the metrics deployment

JAORMX:
This doesn't work on non-OCP environments, so this flag skips the error
that we would normally do so we can keep running the operator without
problems.


mrogers950 (approved):
/lgtm

PR #714, opened 2021-09-21
Proposal for Kubelet Config Remediation

Vincent056:
Created a proposal for  KubeletConfig Remediation

### DIFF
--- a/doc/kubeletConfig-remediations.md
+++ b/doc/kubeletConfig-remediations.md
@@ -1,0 +1,127 @@
+# KubeletConfig Remediation proposed flow design
+
+This page describes the design and implementation of the KubeletConfig 
+remediation templating support in the compliance operator.
+
+## Goals
+We want to accomplish:
+   * The remediation content can have `KubeletConfig` as payload
+   * The remediation controller will be able to merge KubeletConfig remediation into the correct KubeletConfig that is currently being used by the cluster
+
+## Add CaC `KubletConfig` remediation content 
+In CaC content, we are going to add `KubletConfig` remediation besides `MachineConfig` that we currently have.
+
+We will have remdiation that has object payload to be `KubletConfig`. As a exmaple below:
+```yaml
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+spec:
+  kubeletConfig:
+    maxPods: 1100
+```
+
+## Remediation controller handles `KubletConfig` remediation
+
+When we are going to apply one of `KubletConfig` remediation the following will happen:
+
+First, we need to get MachineConfigPool list.
+```go
+mcfgpools := &mcfgv1.MachineConfigPoolList{}
+  if err := r.client.List(context.TODO(), mcfgpools); err != nil {
+    return fmt.Errorf("couldn't list the pools for the remediation: %w", err)
+  }
+```
+
+We then find the poola that meets the remedition nodes' label. ex. worker/master 
+nodes `machineconfiguration.openshift.io/role:`
+
+We will use the nodeselector: `machineconfiguration.openshift.io/role:`  to match 
+machineConfigPoolSelector: `pools.operator.machineconfiguration.openshift.io/:`
+
+### Remediation when no `KubeletConfigs` present
+
+Once we have the machine config pool, we need to find out if an Admin has created 
+any `KubeletConfigs`, to do that we can check `pool.spec.configuration.source`, if 
+there is only one `name: 01-master-kubelet` machine config in the list.
+
+If there is no `KubeletConfigs` for that pool, we will need to create one, and also 
+add a label to the pool depends on the nodeselctor `pools.operator.machineconfiguration.openshift.io/master: ""`/ `pools.operator.machineconfiguration.openshift.io/worker: ""`.: The name of KubeletConfig will be the scan name.
+
+```yaml
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  name: 
+spec:
+  machineConfigPoolSelector:
+    matchLabels:
+      pools.operator.machineconfiguration.openshift.io/master: ""
+  kubeletConfig:
+    maxPods: 1100
+```
+
+### Remediation when there are custom `KubeletConfigs` present
+If there are more than one `KubeletConfigs`, for example `99-master-generated-kubelet`, 
+`99-master-generated-kubelet-1`, `99-master-generated-kubelet-2` ..,we need to find the 
+last one that was used to render the `MachineConfigPool`'s `MachineConfig`. Only the 
+last `KubeletConfigs` will be taken into render the final `MachineConfig`
+
+
+Therefore, we need first to find `99-master-generated-kubelet` with largest number, in 
+this case  it is `99-master-generated-kubelet-4` and then we can find which `KubletConfig` 
+was used to render this MC, in this example:
+
+```yaml
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  annotations:
+    machineconfiguration.openshift.io/generated-by-controller-version: a537783ea4a0cd3b4fe2a02626ab27887307ea51
+  creationTimestamp: "2021-09-20T06:44:07Z"
+  generation: 2
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: 99-master-generated-kubelet-4
+  ownerReferences:
+  - apiVersion: machineconfiguration.openshift.io/v1
+    blockOwnerDeletion: true
+    controller: true
+    kind: KubeletConfig
+    name: set-max-pods-3
+    uid: 8decf73a-e49e-474b-982f-ffaebb261894
+  resourceVersion: "153110"
+  uid: 3cd98e85-6e77-488e-a94b-e84f78ff4029
+...
+```
+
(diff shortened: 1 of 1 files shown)


doc/kubeletConfig-remediations.md
```diff
@@ -0,0 +1,104 @@
...
+  name: co-remediation
+spec:
+  machineConfigPoolSelector:
+    matchLabels:
+      custom-kubelet: compliance-remediaiton
+  kubeletConfig:
+    maxPods: 1100
+```
+
+## Remediation controller handles `KubletConfig` remediaiton
```
JAORMX, line 28: s/remediaiton/remediation/

doc/kubeletConfig-remediations.md
```diff
@@ -0,0 +1,104 @@
...
+  }
+```
+
+We then find the pool that meets the remediation nodes' label. ex. worker/master nodes `machineconfiguration.openshift.io/role:`
+
+### Remediation when no `KubeletConfigs` present
+
+Once we have the machine config pool, we need to find out if an Admin has created any `KubeletConfigs`, to do that we can check `pool.spec.configuration.source`, if there is only one `name: 01-master-kubelet` machine config in the list.
+
+If there is no `KubeletConfigs` for that pool, we will need to create one, and also add a label to the pool, `custom-kubelet=co-kublet-remediation`.
```
JAORMX, line 46: Will we have a default naming scheme used by the Compliance Operator for this? If so, we might want to document it here.

doc/kubeletConfig-remediations.md
```diff
@@ -0,0 +1,104 @@
...
+We will have remdiation that has object payload to be `KubletConfig`. As a exmaple below:
+```yaml
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  name: co-remediation
+spec:
+  machineConfigPoolSelector:
+    matchLabels:
+      custom-kubelet: compliance-remediaiton
```
JAORMX, line 23: will we need this selector? Isn't this something that CO will fill?

Vincent056: So in order to have `KubeletConfig` to be processed and render a machine config, it need to have some sort of  matchLabel to match machineConfigPool, if a admin didn't use any custom`KubeletConfig`, we are going to add one here.

doc/kubeletConfig-remediations.md
```diff
@@ -0,0 +1,104 @@
+# KubeletConfig Remediation proposed flow design
+
+This page describes the design and implementation of the KubeletConfig remediation templating support in the compliance operator.
+
+## Goals
+We want to accomplish:
+   * The remediation content can have `KubeletConfig` as payload
+   * We need to update the `parse_arf_result` to be able to take `KubeletConfig`
```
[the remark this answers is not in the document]
Vincent056: That's a good idea!

doc/kubeletConfig-remediations.md
JAORMX: Let's add these details to the doc

doc/kubeletConfig-remediations.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Add CaC `KubletConfig` remediation content 
+In CaC content, we are going to add `KubletConfig` remediation besides `MachineConfig` that we currently have.
+
+We will have remdiation that has object payload to be `KubletConfig`. As a exmaple below:
+```yaml
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  name: co-remediation
```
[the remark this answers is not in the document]
Vincent056: That's a good point, No I don't think we need to have metadata here, since we are just going to choose to merge with the existing one.

doc/kubeletConfig-remediations.md
Vincent056: Yeah, we might want that there! Is there any specific formatted I should follow?

doc/kubeletConfig-remediations.md
Vincent056: Thanks for that!

doc/kubeletConfig-remediations.md
JAORMX: we have one for machineconfigs: https://github.com/openshift/compliance-operator/blob/d3ac10754b556dabeff9a107ccb611775d2d28d1/pkg/apis/compliance/v1alpha1/complianceremediation_types.go#L145 which is based on the scan name. You could do something similar. Not requiring the numbering which we use for priority, but the scan name would be a good idea.

doc/kubeletConfig-remediations.md
Vincent056: Sounds good

doc/kubeletConfig-remediations.md
Vincent056: Agree, that's a good idea.

Vincent056:
> this looks good! Thanks for the proposal.
> 
> One thing I'd like to see also is... what are the requirements for CO to be able to remediate a KC? Can it just remediate a KC in every instance? or do we have some limitations?
> 
> For instance, for MachineConfigs, we have the limitation that you need to use a nodeSelector, and the nodeSelector needs to match a specific MachineConfigPool, else we skip creating the remediation as it won't be applicable. Should we have something similar for KCs?

I think we are going to use the similar logic here, but there is one limitation, the `KubeletConfig` needs machineConfigPoolSelector instead of nodeSelector, we can continue to use nodeSelector, but there can be some situations that some machineConfigPoolSelector is not equal to nodeSelector, ex. machineConfigPoolSelector that contains  both worker and master nodes, but nodeSelector here will only choose one of them

JAORMX:
> > this looks good! Thanks for the proposal.
> > One thing I'd like to see also is... what are the requirements for CO to be able to remediate a KC? Can it just remediate a KC in every instance? or do we have some limitations?
> > For instance, for MachineConfigs, we have the limitation that you need to use a nodeSelector, and the nodeSelector needs to match a specific MachineConfigPool, else we skip creating the remediation as it won't be applicable. Should we have something similar for KCs?
> 
> I think we are going to use the similar logic here, but there is one limitation, the `KubeletConfig` needs machineConfigPoolSelector instead of nodeSelector, we can continue to use nodeSelector, but there can be some situations that some machineConfigPoolSelector is not equal to nodeSelector, ex. machineConfigPoolSelector that contains both worker and master nodes, but nodeSelector here will only choose one of them

That is fine, if we cannot remediate then we raise a validation error on the ComplianceRemediation object. But we should not introduce machineConfigPoolSelectors into scans or anything of that sort. We should stick to stuff that would work in any kube distro. So, let's stick to nodeSelectors.

Vincent056:
> > > this looks good! Thanks for the proposal.
> > > One thing I'd like to see also is... what are the requirements for CO to be able to remediate a KC? Can it just remediate a KC in every instance? or do we have some limitations?
> > > For instance, for MachineConfigs, we have the limitation that you need to use a nodeSelector, and the nodeSelector needs to match a specific MachineConfigPool, else we skip creating the remediation as it won't be applicable. Should we have something similar for KCs?
> > 
> > 
> > I think we are going to use the similar logic here, but there is one limitation, the `KubeletConfig` needs machineConfigPoolSelector instead of nodeSelector, we can continue to use nodeSelector, but there can be some situations that some machineConfigPoolSelector is not equal to nodeSelector, ex. machineConfigPoolSelector that contains both worker and master nodes, but nodeSelector here will only choose one of them
> 
> That is fine, if we cannot remediate then we raise a validation error on the ComplianceRemediation object. But we should not introduce machineConfigPoolSelectors into scans or anything of that sort. We should stick to stuff that would work in any kube distro. So, let's stick to nodeSelectors.

Sounds good!, https://github.com/openshift/compliance-operator/pull/715/files#diff-37d868832c42510177fe527ffee41dc25b205168859c7a1467e2b9f592fb4ed3L306 Now I see this one, so under ideal situation

doc/kubeletConfig-remediations.md
```diff
@@ -0,0 +1,113 @@
...
+    controller: true
+    kind: KubeletConfig
+    name: set-max-pods-3
+    uid: 8decf73a-e49e-474b-982f-ffaebb261894
+  resourceVersion: "153110"
+  uid: 3cd98e85-6e77-488e-a94b-e84f78ff4029
+...
+```
+
+From here, we know `KubeletConfig` named `set-max-pods-3` is the one used to generate the final pool MC. Next, we need to patch merge the compliance remediation kubelet config remediation to this file.
```
mrogers950, line 86: How will we handle this for a profile with multiple KubeletConfig remediations? If we were to patch them individually, I think we would need to wait for node reboots with each one. So if we can batch them together to a single new KubeletConfig or patch that might be better.

Vincent056: I think that's a good point, I was thinking about pause the pool and then apply the KubeletConfig remediations does it still render a new kubelet MC?

Vincent056: Since we are already pausing the pool for the remediation

JAORMX: yeah, seems to me like pausing the pool is the only way to go, let's add that to the proposal.

mrogers950: Sounds good.

PR #715, opened 2021-09-21
aggregator: Remove MachineConfig validation

JAORMX:
we don't need to do this in the aggregator, the validation is ran anyway
on the remediation controller. This way we can actually show to the user
why a MachineConfig remediation wasn't able to be created.

With this, the aggregator is just a "simple" program that aggregated and
creates remediations, no extra magic.

### DIFF
--- a/cmd/manager/aggregator.go
+++ b/cmd/manager/aggregator.go
@@ -297,14 +297,6 @@ func canCreateRemediationObject(scan *compv1alpha1.ComplianceScan, obj *unstruct
 		return false, "The remediaiton yaml file is empty"
 	}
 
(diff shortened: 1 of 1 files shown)


Vincent056:
I think it is a good approach to leave that handling to the remediation controller

Vincent056 (approved):
/lgmt

jhrozek (approved):
/lgtm
pretty much the same code is done in `verifyAndCompleteMC` in the remediation controller and also given that we don't return an error in the aggregator, but rather try to process as much as possible, it makes sense to move the validation logic to the controller where it has a change to surface up properly.

PR #716, opened 2021-09-21
Validate that rules in tailored profile are of appropriate type

JAORMX:
This PR now validates that the rules added to a tailored profile belong
to a relevant type.

As we know, we have the following types:

* **Platform**: Runs on the platform, do k8s checks.
* **Node**: Runs on the node itself.
* **None**: Informational.

When building a tailored profile, we want there only to be rules that
are gonna be relevant and run as the user expects. Here, you can mix and
match the following:

* Platform rules and None rules
* Node rules and None rules

but nothing in between. Platform and Node rules shouldn't mix.

With this validation, users will now see an error in the tailored
profile and will not build tailored profiles that don't do what they
want.


jhrozek (approved):
/lgtm

PR #717, opened 2021-09-21
Remove Benchmark unit tests

JAORMX:
the `Measure` directive will be removed from Ginkgo v2.

Instead of migrating to something else, this removes the Benchmark tests
entirely. The coverage we were getting out of them is questionable and
we also didn't see much benefit from them from a resource measurement
perspective.


jhrozek (approved):
/lgtm

PR #720, opened 2021-09-23
Makefile: Add push-openscap-image target

mrogers950:
I just ran this target so `quay.io/compliance-operator/openscap-ocp:latest` now has the addition of the version file.
@JAORMX @jhrozek


JAORMX (approved):
/lgtm

PR #721, opened 2021-09-28
add error to the result object as comment

tjungblu:


cmd/manager/scap.go
```diff
@@ -300,6 +300,8 @@ func fetch(client *kubernetes.Clientset, objects []utils.ResourcePath) (map[stri
 				DBG("Encountered non-fatal error to be persisted in the scan: %s", err)
 				objerr := fmt.Errorf("could not fetch %s: %w", uri, err)
 				warnings = append(warnings, objerr.Error())
+				// additionally, we'll add a warning comment in the object so openSCAP can read and process it
+				results[rpath.DumpPath] = []byte("# Error=" + kerrors.ReasonForError(err))
```
JAORMX, line 304: Looks good! Let's make this look a little bit more standard by replacing `Error=` by something like `kube-api-error=` or something of the sort. Also... do we want to do this only for `404` errors? This looks like it will persist a file for every resource it couldn't fetch. What do you think?

JAORMX (changes requested):

cmd/manager/scap.go
tjungblu: yep, sounds sane. I think forbidden might be interesting for some rules as well?

JAORMX: @tjungblu we usually try to fix those in `deploy/role.yaml` here in the operator. The operator should be able to read everything it needs to.

tjungblu: right, then let's do 404 only for now?

JAORMX: sounds good to me!

JAORMX (changes requested):
Could you add a unit test for this? It could be added here https://github.com/openshift/compliance-operator/blob/master/cmd/manager/scap_test.go

JAORMX:
> I'd like to echo the comment @JAORMX had about testing, but (and maybe I just don't see the full picture), doesn't this approach of writing the ReasonForError into the body have the possibility to affect rules that just check for an existence of some yamlpath? I'm not sure we have any honestly and we should be defensive and check that the object fetched is what we intended to fetch, but writing errors into the object result makes me nervous.
> 
> A test would make things clearer.

As far as I can tell, this shouldn't really affect existing rules. But that's exactly why I asked you to take a look as well. Wanted a second opinion on this and if you could think of any rules that would be affected by this (I couldn't). I personally like the approach, as it would end up in an empty YAML file with merely a comment that contains the status.

tjungblu:
added a test @JAORMX @jhrozek

tjungblu:
> just check for an existence of some yamlpath?

do you guys have an understanding of what rules might be impacted by this? I remember @JAORMX explaining that the rules mostly test whether metadata.Name exists (which requires a proper k8s object). Those shouldn't be impacted at all.

jhrozek:
LGTM (explicitly not using the gitopsy lgtm so that Ozz also has the time to look at the test)

jhrozek:
On Tue, Sep 28, 2021 at 08:41:39AM -0700, Jakub Hrozek wrote:
> We do have a bunch of regexes that match against `.*` 

note: most of them would probably be better looking for `.+`, not that
it makes much difference in this case where we hopefully have a
yamlpath..

tjungblu:
@JAORMX wdyt? Shall we merge? Anything I can add here?

JAORMX:
@tjungblu sorry for the silence. I'm doing a little bit of research first before merging this. e.g. how does OpenSCAP's YAML probe react with empty YAML files, and how this could affect future rules. I don't think this will take me too long.

tjungblu:
@JAORMX cool no worries. I'm not in a rush, but I'll be out next week. Merge at your own leisure or ping me if anything needs to be changed :) 
:rocket:

JAORMX:
I think this change is fine. The YAML path will still issue a FAIL even if the file exists but the path isn't available.

PR #722, opened 2021-09-29
Add KubeletConfig Remediaiton Support

Vincent056:
Still a work in progress


pkg/utils/remediationutils.go
```diff
@@ -20,6 +21,16 @@ func IsMachineConfig(obj *unstructured.Unstructured) bool {
 	return "MachineConfig" == objgvk.Kind && mcfgapi.GroupName == objgvk.Group
 }
 
+func IsKubeletConfig(obj *unstructured.Unstructured) bool {
```
JAORMX, line 24: to avoid code duplication, I think we could have a common function for this. e.g.

```
func IsKind(obj *unstructured.Unstructured, kind string) bool {
	if obj == nil {
		return false
	}
	// FIXME(jaosorior): Find a more dynamic way to get
	// the MachineConfig's GVK
	objgvk := obj.GroupVersionKind()
	return kind == objgvk.Kind && mcfgapi.GroupName == objgvk.Group
}
```

So then you could wrap `IsKubeletConfig` as follows:
```
func IsKubeletConfig(obj *unstructured.Unstructured) bool {
  return IsKind(obj, "KubeletConfig")
}

File: pkg/controller/complianceremediation/complianceremediation_controller.go [hunk 2]
```diff
@@ -613,6 +619,55 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteMC(obj *unstructured.U
...
+		//string to int convertion error
+		return fmt.Errorf("string-int convertion error for KC remediation: %w", err)
+	}
+	// we need to patch the remediation if there is already a CustomKC present
+	if hasCustomKC {
+		kubeletConfig := &mcfgv1.KubeletConfig{}
+		kcKey := types.NamespacedName{Name: kubletName, Namespace: rem.Namespace}
+		if err := r.client.Get(context.TODO(), kcKey, kubeletConfig); err != nil {
+			return fmt.Errorf("couldn't get current KubeletConfig: %w", err)
+		}
```
JAORMX, line 654: Here you could set the name of the KC to patch:
```
        // Set kublet config name
	obj.SetName(kubeletConfig.GetName())
	obj.SetLabels(kubeletConfig.GetLabels())
```
And then just return. That way, when the object enters the reconcile loop, it would be handled appropriately.

JAORMX (changes requested):

pkg/utils/remediationutils.go
Vincent056: That's a good idea!

JAORMX (changes requested):
Great start!

This is missing tests and hooking KubeletConfigs to the pausing mechanism in the suite controller.

Vincent056:
> Great start!
> 
> This is missing tests and hooking KubeletConfigs to the pausing mechanism in the suite controller.

The testing is working, tho I feel the code is kinda repetitive from the machine config remediation

pkg/utils/nodeutils.go
```diff
@@ -60,13 +63,50 @@ func GetFirstNodeRole(nodeSelector map[string]string) string {
...
+			return true, &poolList.Items[i]
 		}
 	}
-	return false
+	return false, foundPool
+}
+
+// isMcfgPoolUsingKC check if a MachineConfig Pool is using a custom Kubelet Config
+// if any custom Kublet Config used, return name of generated latest KC machine config from the custom kubelet config
+func IsMcfgPoolUsingKC(pool *mcfgv1.MachineConfigPool) (bool, string, error) {
```
JAORMX, line 78: let's add unit tests for this function

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -613,6 +619,70 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteMC(obj *unstructured.U
 	return nil
 }
 
+// Process kubeletconfig remediation
+func (r *ReconcileComplianceRemediation) verifyAndCompleteKC(obj *unstructured.Unstructured, rem *compv1alpha1.ComplianceRemediation) error {
```
JAORMX, line 623: it would be good to start testing this with unit tests

JAORMX (changes requested):
Let's get e2e tests for this. You could create test content and push it to quay, then test it using a custom ProfileBundle.

pkg/utils/nodeutils.go
Vincent056: Sounds good, I will add unit tests for those as well as the e2e tests!

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: The unit test is all ready, just need the e2e tests

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -613,6 +619,72 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteMC(obj *unstructured.U
...
+	if !ok {
+		return common.NewNonRetriableCtrlError("not applying remediation that doesn't have a matching MachineconfigPool. Scan: %s", scan.Name)
+	}
+
+	// We need to check if there is exsisting kubeletConfigMC present for selected MCP
+	hasCustomKC, kubletMCName, err := utils.IsMcfgPoolUsingKC(pool)
+
+	if err != nil {
+		//string to int convertion error
+		return fmt.Errorf("string-int convertion error for KC remediation: %w", err)
```
jhrozek, line 646: General golang nit: the fact that the error is a str2int should be expressed in err. Ideally, you would just return the error, wrapping it in context that this level knows about. This is also future-proof in case the error would mean something else in the future.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -613,6 +619,72 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteMC(obj *unstructured.U
...
+	// The scans contain a nodeSelector that ultimately must match a machineConfigPool. The only way we can
+	// ensure it does is by checking if it matches any MachineConfigPool's labels.
+	// See also: https://github.com/openshift/machine-config-operator/blob/master/docs/custom-pools.md
+	ok, pool := utils.AnyMcfgPoolLabelMatches(scan.Spec.NodeSelector, mcfgpools)
+	if !ok {
+		return common.NewNonRetriableCtrlError("not applying remediation that doesn't have a matching MachineconfigPool. Scan: %s", scan.Name)
+	}
+
+	// We need to check if there is exsisting kubeletConfigMC present for selected MCP
+	hasCustomKC, kubletMCName, err := utils.IsMcfgPoolUsingKC(pool)
```
jhrozek, line 642: s/kublet/kubelet/

jhrozek (commented):
Two small nits inline, otherwise looks good. Looking forward to the e2e test.

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: > Two small nits inline, otherwise looks good. Looking forward to the e2e test.

Thanks for the review! I will fix them.

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: That's a very good point.

tests/e2e/e2e_test.go
```diff
@@ -3034,5 +3034,161 @@ func TestE2E(t *testing.T) {
...
 		},
+
+		testExecution{
+			Name:       "TestKubeletConfigRemediation",
+			IsParallel: false,
+			TestFn: func(t *testing.T, f *framework.Framework, ctx *framework.Context, mcTctx *mcTestCtx, namespace string) error {
+
+				const baselineImage = "quay.io/wenshen/co_content@sha256:2c16e3d4732bdb2e87ef2833d2f1368295d96f134df805b34710cfeb62d1d028"
+				const requiredRule = "kubelet-eviction-thresholds-set-hard-imagefs-available"
+				const pbName = "upstream-ocp4"
```
JAORMX, line 3045: This test fails as the name `upstream-ocp4` was already used by the previous test `TestVariableTemplate`. I highly suggest making this ProfileBundle have a unique name (maybe something more relevant to the test itself) as opposed to using a hardcoded and general name, as it's making this test fail. We have a function called `getObjNameFromTest` which you could use for such purposes.

Vincent056: Thanks for the advice, I will use that instead!

Vincent056:
```type KubeletConfigSpec struct {
	MachineConfigPoolSelector *metav1.LabelSelector `json:"machineConfigPoolSelector,omitempty"`
	KubeletConfig             *runtime.RawExtension `json:"kubeletConfig,omitempty"`
}``` I think we have to use MachineConfigPoolSelector insteade of node-selector here

**Comment by Vincent056** [comment]
`type KubeletConfigSpec struct {
	MachineConfigPoolSelector *metav1.LabelSelector `json:"machineConfigPoolSelector,omitempty"`
	KubeletConfig             *runtime.RawExtension `json:"kubeletConfig,omitempty"`
}`
I think we have to use MachineConfigPoolSelector insteade of node-selector here @JAORMX We will match the pool with `node-selector` first, then copy pool's `MachineConfigPoolSelector` to `KubeletConfig`

File: pkg/controller/complianceremediation/complianceremediation_controller.go [hunk 8]
```diff
@@ -613,6 +619,70 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteMC(obj *unstructured.U
...
+	}
+
+	// We will need to create a kubelet config if there is no custom KC
+	kubletName := "compliance-operator-kubelet"
+
+	// Set kublet config name
+	obj.SetName(kubletName)
+
+	// Set MachineConfigSelector
+	NodeSelector := []string{"spec", "machineConfigPoolSelector", "matchLabels"}
```
Vincent056, line 680:
```
type KubeletConfigSpec struct {
	MachineConfigPoolSelector *metav1.LabelSelector `json:"machineConfigPoolSelector,omitempty"`
	KubeletConfig             *runtime.RawExtension `json:"kubeletConfig,omitempty"`
}
```
I think we have to use MachineConfigPoolSelector insteade of node-selector here @JAORMX We will match the pool with `node-selector` first, then copy pool's `MachineConfigPoolSelector` to `KubeletConfig`

tests/e2e/e2e_test.go
```diff
@@ -3034,5 +3034,149 @@ func TestE2E(t *testing.T) {
...
+						SetValues: []compv1alpha1.VariableValueSpec{
+							{
+								Name:      prefixName(pbName, "var-kubelet-evictionhard-imagefs-available"),
+								Rationale: "Value to be set",
+								Value:     "20%",
+							},
+						},
+					},
+				}
+				mcTctx.ensureE2EPool()
```
JAORMX, line 3100: This creates a custom e2e pool. We use this so we have shorter MCO update times. It's an e2e pool with one node only. If you start going through that function and the ones it calls, you'll notice it creates a MachineConfigPool that tracks only ONE node, this way, we keep it small and fast.

tests/e2e/e2e_test.go
```diff
@@ -3034,5 +3034,149 @@ func TestE2E(t *testing.T) {
...
+						{
+							APIGroup: "compliance.openshift.io/v1alpha1",
+							Kind:     "TailoredProfile",
+							Name:     suiteName,
+						},
+					},
+					SettingsRef: &compv1alpha1.NamedObjectReference{
+						APIGroup: "compliance.openshift.io/v1alpha1",
+						Kind:     "ScanSetting",
+						Name:     "default-auto-apply",
```
JAORMX, line 3121: instead of using our default settings, you'll need to use other scan settings. We already created some for you to use:

https://github.com/openshift/compliance-operator/blob/4be2108346d71496ecd09bcbce3187b1c4acfba7/tests/e2e/helpers.go#L2137

So... you could use `e2e-default-auto-apply` instead, which will use the needed label to use the aforementioned e2e pool

Vincent056:
```
I1012 11:20:46.829255       1 kubelet_config_controller.go:304] Error syncing kubeletconfig compliance-operator-kubelet: could not find any MachineConfigPool set for KubeletConfig
E1012 11:23:14.278960       1 render_controller.go:217] error finding pools for machineconfig: could not find any MachineConfigPool set for MachineConfig 75-test-remediate-e2e-no-direct-root-logins with labels: map[compliance.openshift.io/scan-name:test-remediate-e2e compliance.openshift.io/suite:test-remediate machineconfiguration.openshift.io/role:e2e]
I1012 11:23:30.679120       1 kubelet_config_controller.go:304] Error syncing kubeletconfig compliance-operator-kubelet: could not find any MachineConfigPool set for KubeletConfig
```
I found from MCO logs, it seems like we don't have label for e2e pool?

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -613,6 +619,72 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteMC(obj *unstructured.U
...
+
+	machineConfigPoolSelector := map[string]string{"pools.operator.machineconfiguration.openshift.io/" + pool.GetName(): ""}
+	unstructured.SetNestedStringMap(obj.Object, machineConfigPoolSelector, NodeSelector...)
```
jhrozek, line 683: This function returns error and we should handle it (although the function only ever returns an error if you pass something else than map[string]interface{}, so we're good here..it's still a good practice)

jhrozek (commented):
Only one nit. Looks really good, thanks for the hard work.

pkg/controller/complianceremediation/complianceremediation_controller.go
Vincent056: Good Catch, I will fix it.

PR #727, opened 2021-10-04
Add variable support for non-urlencoded content

Vincent056:
Since not all remediation templating are URL-encoded, we need different processing for those remediation that does not have URL-encoded content. This Pr is for KubeletConfig remediation, since KubeletConfig is not URL-encoded, and we need to use variable template.  

ex.
`{{.var_some_value}}` is not URL-encoded, we just need to replace the value

### DIFF
--- a/pkg/utils/parse_arf_result.go
+++ b/pkg/utils/parse_arf_result.go
@@ -38,6 +38,9 @@ const (
 	enforcementTypeAnnotationKey    = "complianceascode.io/enforcement-type"
 	optionalAnnotationKey           = "complianceascode.io/optional"
 	valueInputRequiredAnnotationKey = "complianceascode.io/value-input-required"
+	//index to trim `{{`and`}}`
+	trimStartIndex = 2
+	trimEndIndex   = 2
 )
 
 // Constants useful for parsing warnings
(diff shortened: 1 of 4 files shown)


pkg/utils/parse_arf_result.go
```diff
@@ -724,17 +724,34 @@ func parseValues(remContent string, resultValues map[string]string) (string, []s
...
 		//take out `{{ `,' }}' from content
-		trimmedContent := content[3:][:len(content)-6]
-		decodeContent, decodeErr := url.QueryUnescape(trimmedContent)
-		if decodeErr != nil {
-			return remContent, valuesUsedList, valuesMissingList, errors.Wrap(decodeErr, "error while decode remediation context: ")
+		//content starts with `{{.`` is not urlencoded
+		//else the content needs url-decoding first
+		var preProcessedContent string
+		hasEncoding := true
+		if content[2:3] == "." {
```
JAORMX, line 737: Could you add a comment giving a better explanation of why "`.`" means that the variable doesn't have urlencoding set?

JAORMX (changes requested):

pkg/utils/parse_arf_result.go
Vincent056: Sounds good!

JAORMX: Is the only format allowed `{{.<variable name>}}` or should we allow for spaces here too?

pkg/utils/parse_arf_result.go
```diff
@@ -724,17 +724,36 @@ func parseValues(remContent string, resultValues map[string]string) (string, []s
...
+	// if the third charater start with `.` we know that the content is not url-encoded
+	// no processing needed, no {{ ... }} format found
+	//process {{ ... }} content file data one by one and replace them in fixedText
 	for _, content := range contentList {
 		//take out `{{ `,' }}' from content
-		trimmedContent := content[3:][:len(content)-6]
-		decodeContent, decodeErr := url.QueryUnescape(trimmedContent)
-		if decodeErr != nil {
-			return remContent, valuesUsedList, valuesMissingList, errors.Wrap(decodeErr, "error while decode remediation context: ")
+		//content starts with `{{.`` is not urlencoded
```
JAORMX, line 735: What if... instead of checking via if statements, you always url-decode the string? As you already do with `QueryUnescape`. If the string stays the same or there was a decoding error, it means it wasn't urlencoded. If it succeeded unencoding it and the string is shorter than the original, it was urlencoded.

pkg/utils/parse_arf_result.go
Vincent056: I think that's a good point, to be future proof, I think I should use a different way to detect if the content is URL-encoded.

pkg/utils/parse_arf_result.go
Vincent056: That's a good way of doing it! that way we can take array of values as well

pkg/utils/parse_arf_result_test.go
```diff
@@ -403,4 +419,179 @@ Server 3.fedora.pool.ntp.org`
...
+					mode: 420
+					overwrite: true
+					path: /etc/chrony.conf`
+
+			expUsedVals := []string{"the-value-1", "the-value-2", "var-version"}
+			expMissingVals := []string{"the-value-not-defined", "var-file"}
+			_, usedVals, missingVals, err := parseValues(MachineConfig, value_dic)
+
+			It("Should parse with error", func() {
+				Expect(err).To(BeNil())
```
JAORMX, line 549: This seems contradictory. The message says that "it should parse with error", yet the test checks that the error should be `nil`

pkg/utils/parse_arf_result_test.go
```diff
@@ -403,4 +419,179 @@ Server 3.fedora.pool.ntp.org`
...
+				  files: 
+				  - contents:
+					  source: data:,test1%20%3A%3D%20%7B%7B.the_value_1%7D%7D%0Atest2%20%3A%3D%20%7B%7B.the_value_2%7D%7D%0Atest_not_defined%20%3A%3D%20%7B%7B.the_value_not_defined%7D%7D
+					mode: 420
+					overwrite: true
+					path: /etc/chrony.conf`
+
+			processedContent, usedVals, missingVals, err := parseValues(MachineConfig, value_dic)
+
+			It("Should parse with error", func() {
```
JAORMX, line 579: I think you meant "without"

JAORMX (changes requested):
Thanks for adding tests! This is looking much better!

pkg/utils/parse_arf_result_test.go
Vincent056: > Thanks for adding tests! This is looking much better!

Thanks for the review, I missed that, let me correct it!

tests/e2e/e2e_test.go
```diff
@@ -2885,5 +2885,156 @@ func TestE2E(t *testing.T) {
...
+						},
+					},
+				}
+
+				createTPErr := f.Client.Create(goctx.TODO(), tp, getCleanupOpts(ctx))
+				if createTPErr != nil {
+					return createTPErr
+				}
+
+				mcTctx.ensureE2EPool()
```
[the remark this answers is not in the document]
Vincent056: That's a good catch

pkg/utils/parse_arf_result.go
```diff
@@ -717,54 +717,89 @@ func toArrayByComma(format string) []string {
...
-	//process urlencoded content file data one by one and replace them in fixedText
+	}
+	// there are two types of content we need to process, one is url-encoded machine config source ex. {{ -a%20always0-F%20di }},
+	// the other one is not url_encoded ex. {{.var_some_value}}, we are going to take care of url-encoded content first, and then
+	// feed the processed content to template again
+
 	for _, content := range contentList {
-		//take out `{{ `,' }}' from content
+		// if third character is not space, skip non url-encoded content
+		if content[2:3] != " " {
```
JAORMX, line 734: I still feel shaky about this, since it feels to me like a fragile check. Can we use a regex or something else instead?

pkg/utils/parse_arf_result.go
```diff
@@ -717,54 +717,89 @@ func toArrayByComma(format string) []string {
...
+			continue
+		}
+		// take out `{{ `,' }}' out from content
 		trimmedContent := content[3:][:len(content)-6]
-		decodeContent, decodeErr := url.QueryUnescape(trimmedContent)
+		// trimmedContent here should only contain url-encoded content, check if it contains illegall character space
```
JAORMX, line 739: now that I read this again:
```
		trimmedContent := content[3:][:len(content)-6]
```
I'm confused (again)... could we turn the `3` and `len(content)-6` into variables instead to make them more descriptive? If needed, let's add a bit more specific comments on why those numbers were chosen.

the `3` is not too hard, as it's clearly `{{ `, which is 3 characters. But the length  `- 6` is a little trickier and I don't quite remember why it was done this way. Since it's something we might easily forget (which would affect maintenance) being more descriptive is necessary. [1]

[1] https://dev.to/tbutterwith/avoid-magic-numbers-to-increase-code-readability-1da1

pkg/utils/parse_arf_result.go
JAORMX: huh... isn't this already addressed in line 740? do we need this if statement?

pkg/utils/parse_arf_result.go
```diff
@@ -774,7 +809,7 @@ func getParsedValueName(t *template.Template) []string {
 func trimToValue(listToBeTrimmed []string) []string {
 	trimmedValuesList := listToBeTrimmed[:0]
 	for _, oriVal := range listToBeTrimmed {
-		re := regexp.MustCompile("([a-zA-Z]+(_[a-zA-Z]+)+)")
+		re := regexp.MustCompile("([a-zA-Z-0-9]+(_[a-zA-Z-0-9]+)+)")
```
JAORMX, line 812: good find!

pkg/utils/parse_arf_result.go
Vincent056: That's true! But now I thought about this, someone who write content might unintentionally bring a extra space might crash the remediation

JAORMX: > That's true! But now I thought about this, someone who write content might unintentionally bring a extra space might crash the remediation

sounds to me like this would be a nice unit-test to have here!

Vincent056: > now that I read this again:
> 
> ```
> 		trimmedContent := content[3:][:len(content)-6]
> ```
> 
> I'm confused (again)... could we turn the `3` and `len(content)-6` into variables instead to make them more descriptive? If needed, let's add a bit more specific comments on why those numbers were chosen.
> 
> the `3` is not too hard, as it's clearly `{{ `, which is 3 characters. But the length `- 6` is a little trickier and I don't quite remember why it was done this way. Since it's something we might easily forget (which would affect maintenance) being more descriptive is necessary. [1]
> 
> [1] https://dev.to/tbutterwith/avoid-magic-numbers-to-increase-code-readability-1da1

- 6 is because when we use len(content), the real length already become len(content)-3 when we trimmed the first three character, tha's why we need to do -6 to trim the last three character.

PR #728, opened 2021-10-11
Add permission for checking the kubeadmin user

JAORMX:
This adds the necessary RBAC permission for checking if the kubeadmin
secret exists. The intent is to verify that it has been removed, as done
in the accompanying CaC PR [1]. This is possible thanks to the recent
addition that enables us to persist if kube returned NotFound for a
specific object [2].

[1] https://github.com/ComplianceAsCode/content/pull/7723
[2] Commit f200ae00f32ea29670cfa67ee802109b0a0709e2

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -1183,3 +1183,12 @@ rules:
   - get
   - watch
   - list
+# Necessary for checking that the kubeadmin secret has been removed
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - kubeadmin
+  verbs:
+  - get
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -387,6 +387,8 @@ rules:
 # because the OLM doesn't support adding labels to roles nor specifying
 # roleRefs
 # See: https://github.com/operator-framework/operator-lifecycle-manager/issues/732
+# As an addition, this role is also able to read the kubeadmin secret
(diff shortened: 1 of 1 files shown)


Vincent056:
> From the functional point of view the rule along with the content works and detects that I have kubeadmin on my cluster.
> 
> What I want to have a second opinion on from @mrogers950 and @Vincent056 is if it's OK to read secrets in the api-resource-collector and if some precautions would make sense.
> 
> My opinion is that we would need to add these rules sooner or later because our users and customers simply expect the checks and remediations to work out of the box. So what I'm wondering about is if there is something we can do to make this safer. I wish RBAC had an "access" permission to just check if an object exists, but it doesn't. We could emulate that by modifying api-resource-collector to throw away contents of secrets and just dump placeholder files for any secrets API resources, but I'm not sure this accomplishes much -- it would at best prevent someone with access to the namespace from just looking at the API resource dumps. But if someone has access to the namespace and thus the ability to use the api-resource-collector SA, then they automatically have the permissions to read those secrets. So I'm not sure this truncation of secret payloads accomplishes much. Also, at some point we might /want/ to inspect the secrets' bodies (check certificates for whatever..)
> 
> In the end, I would probably accept this PR and just let's make sure we always only allow specific resources by name.
> 
> Thoughts?

Can we create the kubeadmin secret with the same name, since it won't be allowed to create if there is an existing kubeadmin secret, we will know if there is kubeadmin secret without getting it

PR #734, opened 2021-10-15
Add option to make scan scheduling strict/not strict

JAORMX:
Currently, when dealing with node scans, we only allow the scan to go
forward if the node is not unschedulable (via the unschedulable flag of
the Node object). However, for public cloud cases where there might be a
lot of ephemeral nodes (e.g. when a node autoscaler is being used), this
is not ideal as it's very difficult to do scans. On the other hand, if a
node became unready while the scan was on-going, we also return a
similar error.

So, to deal with this, the PR introduces the option `strictNodeScan`,
which enables an administrator to toggle whether we want to be strict
when getting results of a scan.

The default is `true`, which is the setting we were using before.
However, when set to `false`, the scan will not fail if a node is set to
`unschedulable` and will ignore "unschedulable" errors on the scan pods.
Thus allowing us to still get a subset of scans.

### DIFF
--- a/pkg/controller/compliancescan/scantype.go
+++ b/pkg/controller/compliancescan/scantype.go
@@ -111,7 +111,8 @@ func (nh *nodeScanTypeHandler) validate() (bool, error) {
 	nodeWarning := "Not continuing scan: Node is unschedulable"
 	for idx := range nh.nodes {
 		node := &nh.nodes[idx]
-		if node.Spec.Unschedulable {
+		// Surface error if we're being strict with our node scans
+		if nh.getScan().IsStrictNodeScan() && node.Spec.Unschedulable {
 			nh.l.Info(nodeWarning, "Node.Name", node.GetName())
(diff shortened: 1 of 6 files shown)


pkg/controller/compliancescan/scantype.go
```diff
@@ -239,6 +240,14 @@ func (nh *nodeScanTypeHandler) gatherResults() (compv1alpha1.ComplianceScanStatu
 
 		// we output the last result if it was an error
 		if result == compv1alpha1.ResultError {
+			if !nh.getScan().IsStrictNodeScan() {
+				errCode := foundCM.Data["exit-code"]
+				// If the pod was unschedulable and the scan is not
+				// strict, we can skip the error
+				if errCode == common.PodUnschedulableExitCode {
+					continue
```
mrogers950, line 248: It might be helpful to have a warning that the result was skipped under the non-strict mode.

mrogers950 (approved):
/lgtm

PR #738, opened 2021-10-21
Add support for rendering variable in rule objects

Vincent056:
This PR will will render variable in description and rationale for rule objects.

### DIFF
--- a/pkg/profileparser/profileparser.go
+++ b/pkg/profileparser/profileparser.go
@@ -26,9 +26,9 @@ import (
 )
 
 const (
-	machineConfigFixType = "urn:xccdf:fix:script:ignition"
-	kubernetesFixType    = "urn:xccdf:fix:script:kubernetes"
-
+	machineConfigFixType  = "urn:xccdf:fix:script:ignition"
+	kubernetesFixType     = "urn:xccdf:fix:script:kubernetes"
+	valuePrefix           = "xccdf_org.ssgproject.content_value_"
 	controlAnnotationBase = "control.compliance.openshift.io/"
 
 	rhacmStdsAnnotationKey   = "policies.open-cluster-management.io/standards"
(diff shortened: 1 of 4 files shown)


pkg/profileparser/profileparser.go
```diff
@@ -541,6 +541,23 @@ func ParseRulesAndDo(contentDom *xmlquery.Node, stdParser *referenceParser, pb *
...
 
+	allValues := xmlquery.Find(contentDom, "//xccdf-1.2:Value")
+	valuesList := make(map[string]string)
+	for _, variable := range allValues {
+		for _, val := range variable.SelectElements("//xccdf-1.2:value") {
+			if val.SelectAttr("hidden") == "true" {
+				// this is typically used for functions
+				continue
+			}
+			if val.SelectAttr("selector") == "" {
```
jhrozek, line 552: Can you remind me what the empty selector means?

pkg/profileparser/profileparser.go
```diff
@@ -621,10 +638,11 @@ func ParseRulesAndDo(contentDom *xmlquery.Node, stdParser *referenceParser, pb *
 				},
 			}
 			if description != nil {
-				p.Description = utils.XmlNodeAsMarkdown(description)
+				p.Description, _ = utils.RenderValues(utils.XmlNodeAsMarkdownPreRender(description), valuesList)
```
jhrozek, line 641: I agree with not passing an error up, but I think we could at least log the error and not ignore it.

pkg/utils/xml2text.go
```diff
@@ -32,9 +41,14 @@ func xmlToHtml(in string) string {
 			// ignore errors and try to format as much as possible
 			continue
 		}
+
 		switch tok := t.(type) {
 		case xml.StartElement:
-			builder.WriteString(formatElement(tok.Name, "<"))
+			if preRender && tok.Name.Local == "sub" && len(tok.Attr) > 1 {
```
jhrozek, line 47: So can the subelement only ever refer to a value or should the substitution be only done if `tok.Attr[0]` has a variable prefix?

pkg/profileparser/profileparser.go
Vincent056: Not present empty selector means the value is not one of selection values, and it is a default value

Vincent056:
@JAORMX I added the variable used annotations for rules

PR #747, opened 2021-11-04
Log creation of default objects

JAORMX:
There were cases where we need to debug if our default objects have been
created. From the logs it's currently non-trivial to figure this out.

This PR adds relevant logs so we know that the code-path is getting
reached.


jhrozek (approved):
/lgtm

PR #748, opened 2021-11-11
Support deployments on all namespaces

JAORMX:
This ensures that we support that the operator listens on all
namespaces. While this is not a big nor difficult change by itself, it
does require us to make a decision on where to deploy the default
profiles/profilebundles.

In the case where the operator is deployed on all namespaces, only the
operator namespace (openshift-compliance if using the recommendation)
will get our defaults.

### DIFF
--- a/cmd/manager/operator.go
+++ b/cmd/manager/operator.go
@@ -127,23 +127,34 @@ func RunOperator(cmd *cobra.Command, args []string) {
 		log.Error(err, "Failed to get watch namespace")
 		os.Exit(1)
 	}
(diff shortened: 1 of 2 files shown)


Vincent056 (approved):
/lgtm

Vincent056:
@JAORMX I have some questions, I am wondering if the operator gets deployed on all-namespace, it will only have one instance of CO? 

And if the profiles get deployed on `openshift-compliance`, is it also accessible by the CO deployed on all-namespace?

PR #749, opened 2021-11-11
metrics: Add ComplianceSuite status gauge and alert

mrogers950:
* Add the compliance_operator_compliance_state gauge metric that switches to 1 for a ComplianceSuite with a NON-COMPLIANT result, 0 when COMPLIANT, and 2 when INCONSISTENT.
* Create a PrometheusRule warning alert for the gauge.
* Makefile: Apply Prometheus permissions from the manifest directory when running `make deploy-local`

@JAORMX @jhrozek @Vincent056

### DIFF
--- a/cmd/manager/operator.go
+++ b/cmd/manager/operator.go
@@ -87,6 +87,7 @@ var (
 	serviceMonitorBearerTokenFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 	serviceMonitorTLSCAFile       = "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt"
 	serviceMonitorTLSServerName   = "metrics.openshift-compliance.svc"
+	alertName                     = "compliance"
 )
 
 const (
@@ -264,6 +265,11 @@ func addMetrics(ctx context.Context, cfg *rest.Config, kClient *kubernetes.Clien
 
 	if err := createServiceMonitor(ctx, cfg, mClient, operatorNs, metricsService); err != nil {
 		log.Error(err, "Error creating ServiceMonitor")
+		os.Exit(1)
+	}
+
+	if err := createNonComplianceAlert(ctx, mClient, operatorNs); err != nil {
+		log.Error(err, "Error creating PrometheusRule")
 		os.Exit(1)
 	}
 }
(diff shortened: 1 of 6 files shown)


pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -626,3 +632,12 @@ func (r *ReconcileComplianceSuite) getAffectedMcfgPool(scan *compv1alpha1.Compli
 func remediationNeedsOutdatedRemoval(rem *compv1alpha1.ComplianceRemediation, suite *compv1alpha1.ComplianceSuite) bool {
 	return suite.ShouldRemoveOutdated() && rem.Status.ApplicationState == compv1alpha1.RemediationOutdated
 }
+
+func (r *ReconcileComplianceSuite) setSuiteMetric(suite *compv1alpha1.ComplianceSuite) error {
+	if suite.Status.Result == compv1alpha1.ResultCompliant {
+		r.metrics.SetComplianceStateInCompliance(suite.Name)
+	} else if suite.Status.Result == compv1alpha1.ResultNonCompliant {
+		r.metrics.SetComplianceStateOutOfCompliance(suite.Name)
+	}
```
JAORMX, line 641: let's also handle the `INCONSISTENT` state (as a `2` maybe); this is also quite worrying to find in clusters and should be alerted on.

JAORMX (changes requested):

pkg/controller/compliancesuite/compliancesuite_controller.go
mrogers950: thanks, done.

pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -626,3 +632,14 @@ func (r *ReconcileComplianceSuite) getAffectedMcfgPool(scan *compv1alpha1.Compli
...
 }
+
+func (r *ReconcileComplianceSuite) setSuiteMetric(suite *compv1alpha1.ComplianceSuite) error {
+	if suite.Status.Result == compv1alpha1.ResultCompliant {
+		r.metrics.SetComplianceStateInCompliance(suite.Name)
+	} else if suite.Status.Result == compv1alpha1.ResultNonCompliant {
+		r.metrics.SetComplianceStateOutOfCompliance(suite.Name)
+	} else if suite.Status.Result == compv1alpha1.ResultInconsistent {
+		r.metrics.SetComplianceStateInconsistent(suite.Name)
+	}
```
Vincent056, line 643: I think it looks good! Wondering if we want to handle `ResultError` as well.

Vincent056 (changes requested):

cmd/manager/operator.go
```diff
@@ -489,3 +495,39 @@ func serveCRMetrics(cfg *rest.Config, operatorNs string) error {
 	}
 	return nil
 }
+
+// createNonComplianceAlert tries to create the default PrometheusRule. Returns nil.
+func createNonComplianceAlert(ctx context.Context, client *monclientv1.MonitoringV1Client, namespace string) error {
+	rule := monitoring.Rule{
+		Alert: "NonCompliant",
+		Expr:  intstr.FromString(`compliance_operator_compliance_state{name=~".+"} == 1`),
```
JAORMX, line 503: This doesn't alert on an inconsistent state nor an error in the compliance scan. We should adjust accordingly.

JAORMX (changes requested):

cmd/manager/operator.go
mrogers950: Ah, I wasn't sure if this would have needed separate rules for the possible states. I made it > 0 to trigger on any of the non-compliant states.

JAORMX (approved):

PR #750, opened 2021-11-15
Enable runtime customization of API resource collection paths for HyperShift support

tmishina:
### Background
When Compliance Operator tries to perform OpenShift configuration check, its API resource collector pod collects check target Kubernetes API resources.   For example, many Kube API server configuration rules check the Kubernetes API resource `/api/v1/namespaces/openshift-kube-apiserver/configmaps/config`. The path (`/api...`) is hard-coded in each rule ([example](https://github.com/ComplianceAsCode/content/blob/master/applications/openshift/api-server/api_server_insecure_port/rule.yml#L51)), and therefore it is impossible to specify an API resource path at runtime.

### Issue
[HyperShift](https://github.com/openshift/hypershift) is an enabler for managed OpenShift clusters. Master-plane components of a guest cluster in the hypershift environment runs on a separate namespace (per-guest-cluster namespace) of a management cluster (see [architecture document of hypershift](https://hypershift-docs.netlify.app/reference/controller-architecture/) in details).

**Example:** In a standard installation, a Kubernetes API server pod and its configuration exists in namespace `openshift-kube-apiserver`.  However, in a hypershift environment, a Kubernetes API server pod and its configuration are located in a per-guest-cluster namespace (such as `master-guestcluster1`).  The name of the namespace varies, and therefore the hard-coded path doesn't matter.

### Solution (New Feature of the PR)
Change the API resource collector to read the name of an XCCDF variable from a rule, and collect an API resource from the path specified as an XCCDF variable.

Also, [changes on ComplianceAsCode/content](https://github.com/TMISHINA/content/compare/master...hypershift-support) is required.
(One discussion point related to the rule modification is that `filepath` is still required in the rule to build the rules while actual filepath is externalized)

### DIFF
--- a/cmd/manager/scap.go
+++ b/cmd/manager/scap.go
@@ -39,6 +39,7 @@ import (
 
 const (
 	contentFileTimeout = 3600
+	valuePrefix        = "xccdf_org.ssgproject.content_value_"
 )
 
 var (
@@ -152,9 +153,11 @@ func (c *scapContentDataStream) FigureResources(profile string) error {
 		},
 	}
 	effectiveProfile := profile
+	var valuesList map[string]string
 
 	if c.tailoring != nil {
-		selected := getResourcePaths(c.tailoring, c.dataStream, profile)
+		var selected []utils.ResourcePath
+		selected, valuesList = getResourcePaths(c.tailoring, c.dataStream, profile, nil)
 		if len(selected) == 0 {
 			fmt.Printf("no valid checks found in tailoring\n")
 		}
@@ -168,12 +171,13 @@ func (c *scapContentDataStream) FigureResources(profile string) error {
 		}
 	}
 
-	selected := getResourcePaths(c.dataStream, c.dataStream, effectiveProfile)
+	selected, _ := getResourcePaths(c.dataStream, c.dataStream, effectiveProfile, valuesList)
 	if len(selected) == 0 {
 		fmt.Printf("no valid checks found in profile\n")
 	}
 	found = append(found, selected...)
 	c.resources = found
+	DBG("c.resources: %v\n", c.resources)
 	return nil
 }
 
@@ -181,16 +185,58 @@ func (c *scapContentDataStream) FigureResources(profile string) error {
 //
 //  <warning category="general" lang="en-US"><code class="ocp-api-endpoint">/apis/config.openshift.io/v1/oauths/cluster
 //  </code></warning>
-func getPathFromWarningXML(in *xmlquery.Node) []utils.ResourcePath {
+func getPathFromWarningXML(in *xmlquery.Node, valueList map[string]string) []utils.ResourcePath {
 	DBG("Parsing warning %s", in.OutputXML(false))
-	return utils.GetPathFromWarningXML(in)
+	path, err := utils.GetPathFromWarningXML(in, valueList)
+	if err != nil {
+		LOG("Error occurred at parsing warning %s", in.OutputXML((false)))
+		LOG("Error message: %s", err)
+	}
+	return path
 }
 
 // Collect the resource paths for objects that this scan needs to obtain.
 // The profile will have a series of "selected" checks that we grab all of the path info from.
-func getResourcePaths(profileDefs *xmlquery.Node, ruleDefs *xmlquery.Node, profile string) []utils.ResourcePath {
+func getResourcePaths(profileDefs *xmlquery.Node, ruleDefs *xmlquery.Node, profile string, overrideValueList map[string]string) ([]utils.ResourcePath, map[string]string) {
 	out := []utils.ResourcePath{}
 	selectedChecks := []string{}
+
+	// Before staring process, collect all of the variables in definitions.
+	valuesList := make(map[string]string)
+	defs := [...]*xmlquery.Node{ruleDefs, profileDefs}
+	for _, def := range defs {
+		allValues := xmlquery.Find(def, "//xccdf-1.2:Value")
+
+		for _, variable := range allValues {
+			for _, val := range variable.SelectElements("//xccdf-1.2:value") {
+				if val.SelectAttr("hidden") == "true" {
+					// this is typically used for functions
+					continue
+				}
+				if val.SelectAttr("selector") == "" {
+					// It is not an enum choice, but a default value instead
+					if strings.HasPrefix(variable.SelectAttr("id"), valuePrefix) {
+						valuesList[strings.TrimPrefix(variable.SelectAttr("id"), valuePrefix)] = val.OutputXML(false)
+					}
+				}
+			}
+		}
+		allSetValues := xmlquery.Find(def, "//xccdf-1.2:set-value")
+		for _, variable := range allSetValues {
+			if strings.HasPrefix(variable.SelectAttr("idref"), valuePrefix) {
+				valuesList[strings.TrimPrefix(variable.SelectAttr("idref"), valuePrefix)] = variable.OutputXML(false)
+			}
+		}
+	}
+
+	// override variables which is defined in tailored profile
+	if overrideValueList != nil {
+		for k, v := range overrideValueList {
+			if _, exists := valuesList[k]; exists {
+				valuesList[k] = v
+			}
+		}
+	}
 
 	// First we find the Profile node, to locate the enabled checks.
 	DBG("Using profile %s", profile)
@@ -220,7 +266,7 @@ func getResourcePaths(profileDefs *xmlquery.Node, ruleDefs *xmlquery.Node, profi
 	checkDefinitions := ruleDefs.SelectElements("//xccdf-1.2:Rule")
 	if len(checkDefinitions) == 0 {
 		DBG("WARNING: No rules to query (invalid datastream)")
-		return out
+		return out, valuesList
 	}
 
 	// For each of our selected checks, collect the required path info.
@@ -245,7 +291,7 @@ func getResourcePaths(profileDefs *xmlquery.Node, ruleDefs *xmlquery.Node, profi
 			if warn == nil {
 				continue
 			}
-			apiPaths := getPathFromWarningXML(warn)
+			apiPaths := getPathFromWarningXML(warn, valuesList)
 			if len(apiPaths) == 0 {
 				continue
 			}
@@ -261,8 +307,7 @@ func getResourcePaths(profileDefs *xmlquery.Node, ruleDefs *xmlquery.Node, profi
 		}
 
 	}
-
-	return out
+	return out, valuesList
 }
 
 func (c *scapContentDataStream) getExtendedProfileFromTailoring(ds *xmlquery.Node, tailoredProfile string) string {
--- a/pkg/profileparser/profileparser.go
+++ b/pkg/profileparser/profileparser.go
@@ -658,7 +658,12 @@ func ParseRulesAndDo(contentDom *xmlquery.Node, stdParser *referenceParser, pb *
 				}
 			}
 			if warnings != nil {
-				p.Warning = strings.Join(warnings, "\n")
+				p.Warning, valueRendered, err = utils.RenderValues(utils.XmlNodeAsMarkdownPreRender(rationale, false), valuesList)
+				if err != nil {
+					log.Error(err, "couldn't render variable in rules")
+				} else if len(valueRendered) > 0 {
+					p.Annotations[cmpv1alpha1.RuleVariableAnnotationKey] = strings.ReplaceAll(strings.Join(valueRendered, ","), "_", "-")
+				}
 			}
 			if severity != "" {
 				p.Severity = severity
--- a/pkg/utils/parse_arf_result.go
+++ b/pkg/utils/parse_arf_result.go
@@ -81,15 +81,20 @@ type ResourcePath struct {
 //
 //  <warning category="general" lang="en-US"><code class="ocp-api-endpoint">/apis/config.openshift.io/v1/oauths/cluster
 //  </code></warning>
-func GetPathFromWarningXML(in *xmlquery.Node) []ResourcePath {
+func GetPathFromWarningXML(in *xmlquery.Node, valuesList map[string]string) ([]ResourcePath, error) {
 	apiPaths := []ResourcePath{}
 
 	codeNodes := in.SelectElements("//html:code")
+	errMsgs := []string{}
 
 	for _, codeNode := range codeNodes {
 		if strings.Contains(codeNode.SelectAttr("class"), endPointTag) {
-			path := codeNode.InnerText()
+			path, _, err := RenderValues(XmlNodeAsMarkdown(codeNode), valuesList)
 			if len(path) == 0 {
+				continue
+			}
+			if err != nil {
+				errMsgs = append(errMsgs, err.Error())
 				continue
 			}
 			dumpPath := path
@@ -106,8 +111,11 @@ func GetPathFromWarningXML(in *xmlquery.Node) []ResourcePath {
 			apiPaths = append(apiPaths, ResourcePath{ObjPath: path, DumpPath: dumpPath, Filter: filter})
 		}
 	}
-
-	return apiPaths
+	if len(errMsgs) > 0 {
+		return apiPaths, errors.New(strings.Join(errMsgs, "\n"))
+	} else {
+		return apiPaths, nil
+	}
 }
 
 func warningHasApiObjects(in *xmlquery.Node) bool {
--- a/cmd/manager/scap_test.go
+++ b/cmd/manager/scap_test.go
@@ -49,8 +49,114 @@ var _ = Describe("Testing SCAP parsing and storage", func() {
 					DumpPath: "/api/v1/namespaces/openshift-kube-apiserver/configmaps/config",
 				},
 			}
-			got := getResourcePaths(contentDS, contentDS, "xccdf_org.ssgproject.content_profile_platform-moderate")
+			got, _ := getResourcePaths(contentDS, contentDS, "xccdf_org.ssgproject.content_profile_platform-moderate", nil)
 			Expect(got).To(Equal(expected))
+		})
+	})
+
+	Context("Parsing SCAP Content with runtime customization for OCP API resource", func() {
+		var dataStreamFile *os.File
+		var contentDS *xmlquery.Node
+
+		It("Gets the appropriate resource URIs", func() {
+			var err error
+			dataStreamFile, err = os.Open("../../tests/data/ssg-ocp4-ds-new-warning-variable.xml")
+			Expect(err).To(BeNil())
+
+			By("parsing content without errors")
+			contentDS, err = parseContent(dataStreamFile)
+			Expect(err).To(BeNil())
+
+			By("parsing content for warnings")
+			expected := []utils.ResourcePath{
+				{
+					ObjPath:  "/apis/config.openshift.io/v1/oauths/cluster",
+					DumpPath: "/apis/config.openshift.io/v1/oauths/cluster",
+				},
+				{
+					ObjPath:  "/api/v1/namespaces/master-mycluster1/configmaps/kas-config",
+					DumpPath: "/api/v1/namespaces/master-mycluster1/configmaps/kas-config",
+				},
+			}
+			got, _ := getResourcePaths(contentDS, contentDS, "xccdf_org.ssgproject.content_profile_platform-moderate", nil)
+			Expect(got).To(Equal(expected))
+
+			dataStreamFile.Close()
+		})
+
+		It("Gets valid resource URIs even if some of the URL references contain non-existent variable reference", func() {
+			var err error
+			dataStreamFile, err = os.Open("../../tests/data/ssg-ocp4-ds-new-warning-variable-nonexistent.xml")
+			Expect(err).To(BeNil())
+
+			By("parsing content without errors")
+			contentDS, err = parseContent(dataStreamFile)
+			Expect(err).To(BeNil())
+
+			By("parsing content for warnings")
+			expected := []utils.ResourcePath{
+				{
+					ObjPath:  "/apis/config.openshift.io/v1/oauths/cluster",
+					DumpPath: "/apis/config.openshift.io/v1/oauths/cluster",
+				},
+			}
+			got, _ := getResourcePaths(contentDS, contentDS, "xccdf_org.ssgproject.content_profile_platform-moderate", nil)
+			Expect(got).To(Equal(expected))
+			dataStreamFile.Close()
+		})
+
+		It("Gets valid resource URIs even if some of the URL references contain mal-formed go template", func() {
+			var err error
+			dataStreamFile, err = os.Open("../../tests/data/ssg-ocp4-ds-new-warning-variable-malformed.xml")
+			Expect(err).To(BeNil())
+
+			By("parsing content without errors")
+			contentDS, err = parseContent(dataStreamFile)
+			Expect(err).To(BeNil())
+
+			By("parsing content for warnings")
+			expected := []utils.ResourcePath{
+				{
+					ObjPath:  "/apis/config.openshift.io/v1/oauths/cluster",
+					DumpPath: "/apis/config.openshift.io/v1/oauths/cluster",
+				},
+			}
+			got, _ := getResourcePaths(contentDS, contentDS, "xccdf_org.ssgproject.content_profile_platform-moderate", nil)
+			Expect(got).To(Equal(expected))
+			dataStreamFile.Close()
+		})
+
+		It("Gets the appropriate resource URIs customized in a tailored profile", func() {
+			var err error
+			dataStreamFile, err = os.Open("../../tests/data/ssg-ocp4-ds-new-warning-variable.xml")
+			Expect(err).To(BeNil())
+			tpDataStreamFile, err := os.Open("../../tests/data/tailored-profile.xml")
+			Expect(err).To(BeNil())
+
+			By("parsing base profile content without errors")
+			contentDS, err = parseContent(dataStreamFile)
+			Expect(err).To(BeNil())
+
+			By("parsing tailored profile content without errors")
+			tpContentDS, err := parseContent(tpDataStreamFile)
+			Expect(err).To(BeNil())
+
+			By("parsing content for warnings")
+			expected := []utils.ResourcePath{
+				{
+					ObjPath:  "/apis/config.openshift.io/v1/oauths/cluster",
+					DumpPath: "/apis/config.openshift.io/v1/oauths/cluster",
+				},
+				{
+					ObjPath:  "/api/v1/namespaces/customized/configmaps/kas-config",
+					DumpPath: "/api/v1/namespaces/customized/configmaps/kas-config",
+				},
+			}
+			_, valuesList := getResourcePaths(tpContentDS, contentDS, "xccdf_org.ssgproject.content_profile_platform-moderate", nil)
+			got, _ := getResourcePaths(contentDS, contentDS, "xccdf_org.ssgproject.content_profile_platform-moderate", valuesList)
+			Expect(got).To(Equal(expected))
+
+			dataStreamFile.Close()
 		})
 	})
 
--- a/tests/data/tailored-profile.xml
+++ b/tests/data/tailored-profile.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0"?>
+<ds:data-stream-collection xmlns:cat="urn:oasis:names:tc:entity:xmlns:xml:catalog" xmlns:cpe-dict="http://cpe.mitre.org/dictionary/2.0" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:ds="http://scap.nist.gov/schema/scap/source/1.2" xmlns:html="http://www.w3.org/1999/xhtml" xmlns:ind="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent" xmlns:linux="http://oval.mitre.org/XMLSchema/oval-definitions-5#linux" xmlns:ocil="http://scap.nist.gov/schema/ocil/2.0" xmlns:oval="http://oval.mitre.org/XMLSchema/oval-common-5" xmlns:oval-def="http://oval.mitre.org/XMLSchema/oval-definitions-5" xmlns:unix="http://oval.mitre.org/XMLSchema/oval-definitions-5#unix" xmlns:xccdf-1.2="http://checklists.nist.gov/xccdf/1.2" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="scap_org.open-scap_collection_from_xccdf_ssg-ocp4-xccdf-1.2.xml" schematron-version="1.3">
+<ds:component id="scap_org.open-scap_comp_--tailoring--tailoring.xml_tailoring" timestamp="2021-12-15T00:56:41"><xccdf-1.2:Tailoring xmlns:xccdf-1.2="http://checklists.nist.gov/xccdf/1.2" id="xccdf_compliance.openshift.io_tailoring_hypershift-profile">
+  <xccdf-1.2:benchmark href="/content/ssg-ocp4-ds.xml"/>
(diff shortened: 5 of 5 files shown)


JAORMX:
@tmishina we already pull in the `infrastructure/cluster` resource, which could be queried to determine if we're in a HyperShift cluster. There is an open PR [1] that will enable us to query the `controlPlaneTopology` key in the status, which should be `External` if we're in a managed HyperShift cluster.

We could then just add a CPE https://github.com/ComplianceAsCode/content/blob/master/shared/checks/oval/installed_app_is_ocp4.xml to determine if we should run a check. e.g. `installed_app_is_ocp4_hypershift` or something of the sort...

[1] https://github.com/openshift/hypershift/pull/539

tmishina:
Thank you for pointing the [PR](https://github.com/openshift/hypershift/pull/539). My understanding is that the feature of the PR (assigning a special label `External` to labels or annotations) is useful to automatically select rules in a guest cluster (red-colored ones in the dialog below).  However, the feature of this PR is still needed for Compliance Operator instances running on a management cluster (blue one) because the name of namespaces for master-plane components cannot be determined at the time of rule build. I think two PRs are complementary.

co-hs-architecture-diagram

JAORMX:
> Thank you for pointing the [PR](https://github.com/openshift/hypershift/pull/539). My understanding is that the feature of the PR (assigning a special label `External` to labels or annotations) is useful to automatically select rules in a guest cluster (red-colored ones in the dialog below). However, the feature of this PR is still needed for Compliance Operator instances running on a management cluster (blue one) because the name of namespaces for master-plane components cannot be determined at the time of rule build. I think two PRs are complementary.
> 
> co-hs-architecture-diagram

gotcha, I see where you're coming from. Now... wouldn't we then only want to enable variable support for a single part of the  API? e.g. have something like:

```
<html:code class="ocp-api-endpoint">/api/v1/namespaces/<xccdf-1.2:sub idref="xccdf_org.ssgproject.content_value_foo" use="legacy"/>/configmaps/config</html:code>
```

This would allow you to reference variables and even be able to set them dynamically using TailoredProfiles if needed.

tmishina:
When will an element `<xccdf-1.2:sub>` be replaced with its referring value? If my understanding is correct, an API resource check pod runs as follows, and the substitution will take at step 3, and therefore API resource collector cannot gain the benefit of using `<xccdf-1.2:sub>`.

1. init container `content-container` extracts content (data stream) from image to local (temporal) storage
2. init container `api-resource-collector` reads the data stream and collect API resources specified in `<html:code>`
3. container `scanner` performs `oscap` command
4. container `log-collector` collects create a report from the scan result

JAORMX:
> When will an element `<xccdf-1.2:sub>` be replaced with its referring value? If my understanding is correct, an API resource check pod runs as follows, and the substitution will take place at step 3, and therefore API resource collector cannot gain the benefit of using `<xccdf-1.2:sub>`.
> 
>     1. init container `content-container` extracts content (data stream) from image to local (temporal) storage
> 
>     2. init container `api-resource-collector` reads the data stream and collect API resources specified in `<html:code>`
> 
>     3. container `scanner` performs `oscap` command
> 
>     4. container `log-collector` collects create a report from the scan result

right, the intention of my comment was to say that if we want this functionality to work, we could add substitution functionality to the Compliance Operator for this kind of things. The functionality is not there yet, so it needs to be coded, and probably take part of step 2.

Vincent056:
@JAORMX Would this be similar to how we handle remediation templating that we can use a tailored profile to set a value, in this case, it is a custom API resource name?

JAORMX:
> OK, then let me summarize tasks.
> 
>     1. implement a macro function (in [macros.jinja](https://github.com/ComplianceAsCode/content/blob/master/shared/macros.jinja)) which performs:
>        
>        1. embedding `<external_variable>` to XCCDF data stream
>        2. embedding `<xccdf-1.2:sub>` to `<code class="ocp-api-endpoint">` in the XCCDF data stream
> 
>     2. implement a new functionality (in [API resource collector](https://github.com/openshift/compliance-operator/blob/master/pkg/utils/parse_arf_result.go#L84-L111)) which performs substitution specified with `<xccdf-1.2:sub>`
> 
> 
> One thing I'd like to know is how to embed `<external_variable>`. The macro will be called in `.warnings.general` in `rule.yml`, and the output of macro goes under `<xccdf-1.2:warning>` element. if my understanding is correct, the `<external_variable>` element should be located at upper level (e.g., direct child of `<xccdf-1.2:Profile>`). Is there any way to do so?

@tmishina : The warning would contain `xccdf-1.2:sub` indicating that a substitution needs to be made. The variable needs to be set somewhere... it could be on a Profile as you say. But most likely than not it'll be in a tailoring. @Vincent056 did very similar work to substitute variables on remediations, where, we take the default value of a variable if it isn't set, else, we look a the tailoring.

The idea is that we'd be able to substitute the aspects that are unique for a deployment, and we'd set them up in a tailoring when needed. And the default case would then just work for regular OpenShift.

> @JAORMX Would this be similar to how we handle remediation templating that we can use a tailored profile to set a value, in this case, it is a custom API resource name?

@Vincent056 exactly!

Vincent056:
@tmishina Here are a few PRs I think might be helpful, We are currently rendering the XCCDF variable's value in the rule's description and rational [1] section, here we get the default value list from datastream, and we then parse all the rules, we replace the place where an XCCDF variable present to the default value using the value list. 

Also, we are using the XCCDF variable in remediations,[2][3] the way we substitute value is similar to the one above, however, we only get XCCDF variable value from `result ConfigMap`[6], which is the values used to do the scanning, the values could be from datastream default value as well as tailored profile, so a little correction to what @JAORMX mentioned about tailor profile above. However, we are indeed using `tailored profile` to check if the variable is set in a `tailored profile` if a certain value is in `value required` annotation[4] , I think you can change the `isRequiredValueSet` function[5] to extract the value you needed from the tailored profile, and then parse them in warning section.

I think what you can do is to first look for value in `tailored profile`, if not found, then it will go to datastream to use the default value(ex. in this case, the default value could be `openshift-kube-apiserver`). You can create a new variable, and set the default value in the compliance as code content that is used to generate the datastream (ex. [7]), so that the implementation will still work for the normal openshift cluster if no value is provided in the tailored profile.

Link:

[1] https://github.com/openshift/compliance-operator/pull/738
[2] https://github.com/openshift/compliance-operator/pull/672
[3] https://github.com/openshift/compliance-operator/pull/727
[4] https://github.com/openshift/compliance-operator/blob/master/doc/remediation-templating.md
[5] https://github.com/openshift/compliance-operator/blob/88f0dc5103ceb2edc2594bfc040842eb667ac822/pkg/controller/complianceremediation/complianceremediation_controller.go#L458
[6] https://github.com/openshift/compliance-operator/blob/88f0dc5103ceb2edc2594bfc040842eb667ac822/pkg/utils/parse_arf_result.go#L396
[7] https://github.com/ComplianceAsCode/content/blob/master/applications/openshift/kubelet/var_kubelet_evictionhard_imagefs_available.var

tmishina:
@Vincent056 Thank you very much, I think expanding the substitution feature from `description` element  to `warning` element to enables the API resource collector pod to collect a resource specified in a variable.

Another issue is that we need to specify the check file target in a template `yamlfile_value`. To solve the issue, I tired to add a new element `filepath_xccdf_variable` to tell the name of variable to the `yamlfile_value` template.

- Example rule: https://github.com/tmishina/content/blob/hypershift-support/applications/openshift/api-server/api_server_admission_control_plugin_AlwaysAdmit/rule.yml#L45
- change in `oval.template` of `yamlfile_value` template: https://github.com/tmishina/content/blob/hypershift-support/shared/templates/yamlfile_value/oval.template#L21-L31

I had tried to reuse the variable designated in `warning` element, but for me it is impossible because the variable specified in `warning` element cannot been referred in `oval.template`.

Do you have any idea to solve the issue without using the additional variable (`filepath_xccdf_variable`)?

JAORMX:
@tmishina do you have a PR  already in ComplianceAsCode? Maybe we could discuss solutions there instead and include the maintainers of CaC. They might be able to help.

tmishina:
@JAORMX Thank you for suggestion, I've note created a PR on ComplianceAsCode/content yet. I will do that.

I have tried to implement the idea discussed on this PR.
https://github.com/openshift/compliance-operator/compare/master...tmishina:parameter-subst-in-warnings

I will write test for this (I'm learning the test of this project; scap_test.go looks like the best position to add test for this new feature).

tmishina:
@JAORMX I have implemented test, and changed the code of the source branch `custom-api-resource` from my old one to the one we are discussing now (merged sub-branch `parameter-subst-in-warnings` into the source branch).
I'd like to hear your comment on the following topics, thanks!

- Do this test implementation and test data align with the community rule? Please tell me if I need to read some document about test implementation.
- [At the text substitution](https://github.com/openshift/compliance-operator/pull/750/files#diff-711414826f224eb48348b26463adec3363b3755d57196c843c06cdd9478ba338R90-R91), simply reusing `RenderValues` caused error because the substituted text is surrounded by spaces and therefore I used `strings.Replace` to remove the spaces.  Is there better solution than using `strings.Replace`?

Vincent056:
> @Vincent056 Thank you very much, I think expanding the substitution feature from `description` element to `warning` element to enables the API resource collector pod to collect a resource specified in a variable.
> 
> Another issue is that we need to specify the check file target in a template `yamlfile_value`. To solve the issue, I tired to add a new element `filepath_xccdf_variable` to tell the name of variable to the `yamlfile_value` template.
> 
>     * Example rule: https://github.com/tmishina/content/blob/hypershift-support/applications/openshift/api-server/api_server_admission_control_plugin_AlwaysAdmit/rule.yml#L45
> 
>     * change in `oval.template` of `yamlfile_value` template: https://github.com/tmishina/content/blob/hypershift-support/shared/templates/yamlfile_value/oval.template#L21-L31
> 
> 
> I had tried to reuse the variable designated in `warning` element, but for me it is impossible because the variable specified in `warning` element cannot been referred in `oval.template`.
> 
> Do you have any idea to solve the issue without using the additional variable (`filepath_xccdf_variable`)?

Agree with @JAORMX I think it might be good to get CaC repo maintainers involved on this one.

pkg/utils/parse_arf_result.go
```diff
@@ -81,14 +80,15 @@ type ResourcePath struct {
...
+func GetPathFromWarningXML(in *xmlquery.Node, valuesList map[string]string) []ResourcePath {
 	apiPaths := []ResourcePath{}
 
 	codeNodes := in.SelectElements("//html:code")
 
 	for _, codeNode := range codeNodes {
 		if strings.Contains(codeNode.SelectAttr("class"), endPointTag) {
-			path := codeNode.InnerText()
+			s, _, _ := RenderValues(XmlNodeAsMarkdownPreRender(codeNode), valuesList)
+			path := strings.Replace(s, " ", "", -1)
```
Vincent056, line 91: I wonder will `strings.TrimSpace(s)` works in this case?

Vincent056 (commented):
I think it looks good to me overall! Let's see if there is any input from @JAORMX

pkg/utils/parse_arf_result.go
tmishina: `strings.TrimSpace(s)` just removes beginning and ending spaces, and therefore in this case it doesn't work as expected.
```
                DumpPath: "/api/v1/namespaces/ master-mycluster1 /configmaps/ kas-config",
```

For error check I misunderstand the behaviour of `RenderValues`; I just saw the following line

https://github.com/openshift/compliance-operator/blob/f5fc6e18f72b29f42a9f9b58228aab46f269be44/pkg/utils/xml2text.go#L105

but I realized that in the following lines it returns error. I will update my code to handle these errors. Thank you for pointing out my mistake.

https://github.com/openshift/compliance-operator/blob/f5fc6e18f72b29f42a9f9b58228aab46f269be44/pkg/utils/xml2text.go#L95

https://github.com/openshift/compliance-operator/blob/f5fc6e18f72b29f42a9f9b58228aab46f269be44/pkg/utils/xml2text.go#L101

tmishina: @Vincent056 I added an error case in the test code, and I found that this modification is required to detect an error case which specifies non-existent variable.

https://github.com/openshift/compliance-operator/pull/750/files#diff-ad98feb1f5a6f396117cf1436906a17cf571de01e7a82cd4389e906d179256a9L92-R92

`make test-unit` shows no fail for this modification, but I'm wondering how this modification affects.
Is this modification acceptable?

Vincent056: > 

@tmishina Thanks for the PR; sorry about the extra space, I just realized that the extra space was added by https://github.com/Vincent056/compliance-operator/blob/88e8b77ae995f93cfa484c1d39107a62f2bda3f7/pkg/utils/xml2text.go#L68, it was introduced because there aren't any spaces around variable in the rule's description and rationale, maybe we can have it removed here, and I can use a different way to add space for that maybe add in CaC before it gets parsed in Compliance Operator.

Vincent056: @tmishina as far as detecting a missing value here https://github.com/openshift/compliance-operator/pull/750/files#diff-ad98feb1f5a6f396117cf1436906a17cf571de01e7a82cd4389e906d179256a9L92-R92, If I remember correctly, I think the reason it wasn't used here is that it will stop the parsing for the reset of content altogether if there is a missing value, and we don't want to have such behavior in description and rational sections. The err here was mainly for wrongly formatted content, such as `{{.so me-var}` an extra space in the content. @JAORMX suggests that we can issue an event when there is an error. Alternative to detect missing values without using `missingkey=error` https://github.com/openshift/compliance-operator/blob/fd0ab5b03eeb70ab45101b8e65518154e11730ef/pkg/utils/parse_arf_result.go#L820-L831

tmishina: @Vincent056 sorry but let me understand your suggestion...

For the first issue, can I just revert https://github.com/Vincent056/compliance-operator/blob/88e8b77ae995f93cfa484c1d39107a62f2bda3f7/pkg/utils/xml2text.go#L68 for now? or do I need to do more in addition to the revert? We also need to sync with the change on CaC as you mentioned (`I can use a different way to add space for that maybe add in CaC before it gets parsed in Compliance Operator.`).

For the second issue, how can we detect substitution errors? In case of the example you showed, it is an error case when a parsed value is not in the result value. However, in the case of this PR, not all of the variables listed in the argument are replaced, and therefore it is difficult to detect the substitution error.

Vincent056: @tmishina Sorry for the confusion, for the first issue, I have a PR to fix the issue, https://github.com/openshift/compliance-operator/pull/753.

tmishina: @Vincent056 
For #1, I saw your PR; I will wait for its merger.

For #2, I'm still unclear. sorry for that...let me clarify how variable declaration is substituted to real value using `xccdf_value(varname)` defined in `macros.jinja` of `ComplianceAsCode/conent`.

1. user writes a statement like:

   ```yaml
   warning:
     general: {{{xccdf_value("myvar")}}}
   ```

2. at a time of datastream image build, ComplianceAsCode/content converts YaML into XML.

   ```xml
   <xccdf-1.2:warning xml:lang="en-US" category="general"><xccdf-1.2:sub idref="xccdf_org.ssgproject.content_value_myvar" use="legacy"/></xccdf-1.2:warning>
   ```

3. at a time of scan, Compliance Operator replaces `<xccdf-1.2:sub>` with an actual value

   ```xml
   <xccdf-1.2:warning xml:lang="en-US" category="general">myvarvalue</xccdf-1.2:warning>
   ```



You said `For the second issue, I think we should be able to detect if there is substitution error due to wrong formate` , but the input of `scan.go` (or `parse_arf_result.go`) is an XML data, and expressions like `{{varname}}` has already been replaced with XML elements by jinja of ComplianceAsCode/content. What type of format error are you expecting?

Vincent056: Hi, @tmishina Do you have time for a quick google meeting, I think you might not need to change anything in CaC after all. You can just use `{{.myvar}}` in the content.

```yaml
warnings:
    - general: |-
            {{{ openshift_cluster_setting("/api/v1/namespaces/{{.myvar}}/configmaps/config") | indent(8) }}}
```
will do the trick,

Vincent056: There is one step between 2 and 3,
We convert
```yaml
<xccdf-1.2:warning xml:lang="en-US" category="general"><xccdf-1.2:sub idref="xccdf_org.ssgproject.content_value_myvar" use="legacy"/></xccdf-1.2:warning>
```
into 
```yaml
<xccdf-1.2:warning xml:lang="en-US" category="general">{{.myvar}}</xccdf-1.2:warning>
```
and then we replace {{.myvar}} to the actual value in your step 3

tmishina: @Vincent056 thank you for the conversation! I revisited the code, and I realized that we need to change both `profileparser.go` and `parse_arf_result.go` because:

- Variable substitution for `description` and `rationale` is done at the time of parsing profile (by `profileparser.go`) ([code](https://github.com/openshift/compliance-operator/blob/da001d6d2fb2ace21386fb9ed7d98462bd98d324/pkg/profileparser/profileparser.go#L643)), and its output is stored as a Profile resource
- However, API resource collection is done at the time of scan by `parse_arf_result.go`, and its source is not the Profile resource but a data stream (XML) ([code](https://github.com/openshift/compliance-operator/blob/da001d6d2fb2ace21386fb9ed7d98462bd98d324/cmd/manager/scap.go#L171)), and therefore substitution at the time of parsing profile does not affect the API resource collection operation
- Hence, we need to modify:
  - `profileparser.go` to substitute variables in Profile resources
  - `parse_arf_result.go` to substitute variables in the target of API resource collection

Is my understanding correct?

Vincent056: I believe that’s the correct.

tmishina: @Vincent056 pushed the revised version which modifies `parse_arf_result.go`. Important change is the following line.

https://github.com/openshift/compliance-operator/pull/750/files#diff-711414826f224eb48348b26463adec3363b3755d57196c843c06cdd9478ba338R90

now this expression is acceptable.

```
/api/v1/namespaces/{{.openshift_kube_apiserver_config_namespace}}/configmaps/{{.openshift_kube_apiserver_config_name}}
```

I will also change `profileparser.go`.

Vincent056: @tmishina I think the code looks good!

pkg/profileparser/profileparser.go
```diff
@@ -658,7 +658,12 @@ func ParseRulesAndDo(contentDom *xmlquery.Node, stdParser *referenceParser, pb *
 				}
 			}
 			if warnings != nil {
-				p.Warning = strings.Join(warnings, "\n")
+				p.Warning, valueRendered, err = utils.RenderValues(utils.XmlNodeAsMarkdownPreRender(rationale), valuesList)
```
tmishina, line 661: @Vincent056 sorry but could you tell me where is the best position to add the test for this new feature? `profileparser_test.go` must be the target file, but in which part do I need to add the test?

Vincent056: @tmishina maybe create something like this one https://github.com/openshift/compliance-operator/blob/10f5c89c801a5f10b75068ac2da8d9b1b29120cb/pkg/profileparser/profileparser_test.go#L607-L633

tmishina: @Vincent056 I wrote test and the test failed. I check the reason and found this.

https://github.com/openshift/compliance-operator/blob/bb69581e97be73e78bb25dad2ff797de763eea70/pkg/utils/parse_arf_result.go#L517-L521

This code means: when a `xccdf-1.2:warning` contains an element which class is `ocp-api-endpoint`,  such a `warning` element will not be parsed, and therefore substitution function will never be called.

My conclusion here is that we don't need to add a functionality to  to substitute variables in `warning` element at a time of profile parsing. Is my understanding correct?

Vincent056: @JAORMX, Do you have any knowledge on this one? If we are using custom variables in warning, I think we should also parse it?

tmishina:
I found important issue at the test with my modified ComplianceAsCode/content.

The issue is that customized variables specified in a tailored profile resource does not affect the API resource collector, and it collects resources specified as default value for each variable.  It ignores customized values in the tailored profile resource. That is not an expected behaviour. 

The reason is that API resource collector checks `warnings` elements in the data stream stored in a ComplianceAsCode/content image (no substitution), not in a profile resource (substituted).
(I've not checked whether this also occurs in description/rationale elements)



As far as I know, there are two solution candidates.

1. use my first code (introduce special macro to embed variable reference in a data stream image and parse it in CO)
   - pros: [small modification](https://github.com/tmishina/compliance-operator/commit/505ebf9de27fe80d71fe7176e62f381fccb56f5a)
   - cons: substitution mechanism differ than that for description/rationale of `profile` resource
2. modify the API resource collector to use `warnings` in kube resource instead of data stream
   - pros: same substitution mechanism as description/rationale
     - cons: a number of (maybe 10-100 order) lines should be changed (mainly at [scap.go](https://github.com/openshift/compliance-operator/blob/0a7246b39ac3bffb33961341a4593bf581661543/cmd/manager/scap.go#L242-L248))

My preference is 1, but @Vincent056, @JAORMX, I'd appreciate it if I can hear your opinions, thanks.

pkg/profileparser/profileparser.go
Vincent056: > I found important issue at the test with my modified ComplianceAsCode/content.
> 
> The issue is that customized variables specified in a tailored profile resource does not affect the API resource collector, and it collects resources specified as default value for each variable. It ignores customized values in the tailored profile resource. That is not an expected behaviour.
> 
> The reason is that API resource collector checks `warnings` elements in the data stream stored in a ComplianceAsCode/content image (no substitution), not in a profile resource (substituted). (I've not checked whether this also occurs in description/rationale elements)
> 
> As far as I know, there are two solution candidates.
> 
> 1. use my first code (introduce special macro to embed variable reference in a data stream image and parse it in CO)
>    
>    * pros: [small modification](https://github.com/tmishina/compliance-operator/commit/505ebf9de27fe80d71fe7176e62f381fccb56f5a)
>    * cons: substitution mechanism differ than that for description/rationale of `profile` resource
> 2. modify the API resource collector to use `warnings` in kube resource instead of data stream
>    
>    * pros: same substitution mechanism as description/rationale
>      
>      * cons: a number of (maybe 10-100 order) lines should be changed (mainly at [scap.go](https://github.com/openshift/compliance-operator/blob/0a7246b39ac3bffb33961341a4593bf581661543/cmd/manager/scap.go#L242-L248))
> 
> My preference is 1, but @Vincent056, @JAORMX, I'd appreciate it if I can hear your opinions, thanks.

Hi @tmishina, Yes, in the warning and rational section, it only gets values from the input value list when calling `RenderValues`, but I believe you can first get the tailored profile CM, then get the value list, and feed it to the parsing function. in https://github.com/openshift/compliance-operator/blob/441faa0d3fe6d43a4803297b9615bb74bc8dcca6/cmd/manager/scap.go#L186 you can parse the content here before calling `utils.GetPathFromWarningXML(in)`

tmishina: @Vincent056 Thank you for your suggestion! I made changes.
Major update is done to `scap.go`, especially the change on the method signature of `getPathFromWarningXML`.

https://github.com/openshift/compliance-operator/pull/750/files#diff-f86e02e20295a205521712d1773d904635f825c76fff3e0408ad4decc24c578aL191

In addition, I implemented test for the new functionality (overriding variables with the ones in tailored profile).

https://github.com/openshift/compliance-operator/pull/750/files#diff-2cd82c26c0af69487c7dfc803d1b54c1091dc696c3071969c6f3d3d9cc0acf94R108-R139

Are these changes OK?

pkg/utils/parse_arf_result.go
```diff
@@ -81,14 +80,19 @@ type ResourcePath struct {
...
 
 	for _, codeNode := range codeNodes {
 		if strings.Contains(codeNode.SelectAttr("class"), endPointTag) {
-			path := codeNode.InnerText()
+			s, _, err := RenderValues(XmlNodeAsMarkdown(codeNode), valuesList)
+			if err != nil {
+				fmt.Printf(err.Error() + "\n")
+				continue
+			}
+			path := strings.Replace(s, " ", "", -1)
```
Vincent056, line 95: I guess I will reopen the PR that remove the extra space.

Vincent056: This has been merged! https://github.com/openshift/compliance-operator/pull/753

cmd/manager/scap.go
```diff
@@ -168,30 +171,71 @@ func (c *scapContentDataStream) FigureResources(profile string) error {
...
+				if val.SelectAttr("selector") == "" {
+					// It is not an enum choice, but a default value instead
+					if strings.HasPrefix(variable.SelectAttr("id"), valuePrefix) {
+						DBG("value assignment: %s -> %s\n", strings.TrimPrefix(variable.SelectAttr("id"), valuePrefix), val.OutputXML(false))
+						valuesList[strings.TrimPrefix(variable.SelectAttr("id"), valuePrefix)] = val.OutputXML(false)
+					}
+				}
+			}
+		}
+		allSetValues := xmlquery.Find(def, "//xccdf-1.2:set-value")
```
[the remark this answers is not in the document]
tmishina: Yes, `Profile` element (including tailored profiles) may contain `set-value` elements which override `Value` 
(reference: [Specification for the Extensible Configuration Checklist Description Format (XCCDF) Version 1.2](https://csrc.nist.gov/CSRC/media/Publications/nistir/7275/rev-4/final/documents/nistir-7275r4_updated-march-2012_markup.pdf) , p. 34), and value customization by tailored profile is achieved by these elements.  And therfore 
we need to override `valuesList` with this `allSetValues`.

cmd/manager/scap.go
```diff
@@ -168,30 +171,71 @@ func (c *scapContentDataStream) FigureResources(profile string) error {
 		}
 	}
 
-	selected := getResourcePaths(c.dataStream, c.dataStream, effectiveProfile)
+	selected, _ := getResourcePaths(c.dataStream, c.dataStream, effectiveProfile, valuesList)
```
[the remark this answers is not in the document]
tmishina: If `c.tailoring == nil`, then we don't need to provide values list, and therefore I make `valuesList` remain `nil`.
(I [declare](https://github.com/tmishina/compliance-operator/blob/custom-api-resource/cmd/manager/scap.go#L156) `valueList` as a map, which default value is `nil` according to the go language specification)

 Please tell me if this is not a good manner.

cmd/manager/scap.go
```diff
@@ -168,30 +171,71 @@ func (c *scapContentDataStream) FigureResources(profile string) error {
...
+func getPathFromWarningXML(in *xmlquery.Node, valueList map[string]string) []utils.ResourcePath {
 	DBG("Parsing warning %s", in.OutputXML(false))
-	return utils.GetPathFromWarningXML(in)
+	return utils.GetPathFromWarningXML(in, valueList)
 }
 
 // Collect the resource paths for objects that this scan needs to obtain.
 // The profile will have a series of "selected" checks that we grab all of the path info from.
-func getResourcePaths(profileDefs *xmlquery.Node, ruleDefs *xmlquery.Node, profile string) []utils.ResourcePath {
+func getResourcePaths(profileDefs *xmlquery.Node, ruleDefs *xmlquery.Node, profile string, overrideValueList map[string]string) ([]utils.ResourcePath, map[string]string) {
```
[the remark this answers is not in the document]
tmishina: The original code consists of following steps.

1. if tailored profile exists, process it
    1. extract resource paths from `warning` elements
    2. extract extended profile (as `effectiveProfile`)
2. process profile (or extended profile)
    1. extract resource paths from `warning` elements

I want to replace not only variables in `warning` in the tailored profile (step 1.1) but also those in `warning` in the extended profile (step 2.1).  In step 2.1, we need to override values with the ones declared in the tailored profile.  It is achieved by using the second return value of `getResourcePaths` at step 1.2 for the final parameter of `getResourcePaths` at step 2.1.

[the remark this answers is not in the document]
Vincent056: Thanks for the explanation. I think the code looks good to me, just need to remove some of `fmt.Println(err)`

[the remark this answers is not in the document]
tmishina: sorry I didn't pushed the latest code to the repo. Now most of the `fmt.Println()` and `fmt.Printf()` occurrences are replaced with `DBG()`.
(There remains several `fmt.PrintXX()` in `scap.go`, but they are not added by me; I think that they should also be replaced with `DBG()`, but  such a replacement would be done on another PR if needed)

One thing I'd like to discuss is the error handling In `parse_arf_result.go`. I can't find any error output in it. Instead, it ignores error or returns error to callers ([example](https://github.com/tmishina/compliance-operator/blob/custom-api-resource/pkg/utils/parse_arf_result.go#L407)). What is the best way to handle the error at [calling RenderValues](https://github.com/tmishina/compliance-operator/blob/custom-api-resource/pkg/utils/parse_arf_result.go#L95)?  Currently, I use `fmt.Printf` here.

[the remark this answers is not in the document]
tmishina: @Vincent056 Is there any progress on this PR? or is there anything that I need to do? thanks.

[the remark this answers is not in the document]
Vincent056: @tmishina Sorry for the late review! I think the PR looks good, for the question regarding how to handle error output, does it make sense if we use DBG() for `RenderValues` as well? I am going to think about the best approach for CaC PR.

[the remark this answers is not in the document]
tmishina: @Vincent056 `DBG()` is defined in `main` package (`cmd/manager/log.go\), and is not defined in `utils` package, to which `parse_arf_result.go` belongs.

I saw several functions of `utils` package, and I imagine that the authors of `utils` functions intentionally avoid handling error in `utils` functions.  Instead, these functions just return error messages their callers, so do functions in `parse_arf_result.go`. That's why I'm wondering what is the best way to handle errors in `GetPathFromWarningXML()` in `parse_arf_result.go`.  Can `GetPathFromWarningXML()` return error when replacement fails?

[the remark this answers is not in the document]
tmishina: now `utils.utils.GetPathFromWarningXML()` in `par_arf_result.go`  just returns error messages, and its caller, `getPathFromWarningXML()` in `scap.go`, writes log message.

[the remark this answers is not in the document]
Vincent056: Hi, @tmishina I just realized we are using a different way to log in to the example I gave above, I think DEBUG() and LOG() you had in there is good. s/remote/remove just nits on one of the commit message.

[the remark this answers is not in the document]
tmishina: @Vincent056 thank you for pointing out my typo...I've just fixed it and force-pushed the branch.

It would be better to squash commits into reasonable numbers by `git rebase -i`.  I will do that if the PR becomes ready to be merged.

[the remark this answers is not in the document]
Vincent056: Thanks, I think the PR looks good overall.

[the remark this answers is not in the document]
tmishina: Thank you! What is the remaining actions for me? `git rebase master` must be done, and what else?
The test bound with this PR is still `Some checks haven’t completed yet`; is there anything I can do to solve this? (I hope `git rebase master` and force pushing a new commit can fix this...)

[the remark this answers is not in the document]
tmishina: @Vincent056 done `git rebase master`, and `make test-unit` finishes without error.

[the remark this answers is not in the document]
Vincent056: I am wondering if we want to squash the commit?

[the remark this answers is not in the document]
tmishina: @Vincent056 I squashed the commits of this PR into one, and `rebase master` again. Now this PR is in the next of commit fc7a2579

Vincent056 (approved):
/lgtm

PR #751, opened 2021-11-17
Make showing `NotApplicable` results optional

JAORMX:
This adds the ability to optionally show or not show the `NotApplicable`
results as part of the compliance results. We've noticed that people get
concerned with the quantity of `NotApplicable` results, which will
increase as we add specific checks for different platforms.

This reduces the user's burden by skipping showing NotApplicable results
if the `showNotApplicable` is set to `false`. This also makes that
option the default.


jhrozek (approved):
/lgtm

PR #753, opened 2021-12-02
Add extra space optional around variable during parsing

Vincent056:
Since we are introducing the variable for API Resource Path for HyperShift, we don't want extra space around the variable as we do in the description, and rationale, hence we added a boolean to make it optional.

### DIFF
--- a/pkg/utils/xml2text.go
+++ b/pkg/utils/xml2text.go
@@ -12,24 +12,24 @@ import (
 	"github.com/pkg/errors"
 )
 
-func XmlNodeAsMarkdownPreRender(node *xmlquery.Node) string {
-	return xmlToMarkdown(node.OutputXML(false), true)
+func XmlNodeAsMarkdownPreRender(node *xmlquery.Node, needsSpace bool) string {
+	return xmlToMarkdown(node.OutputXML(false), true, needsSpace)
 }
 
 func XmlNodeAsMarkdown(node *xmlquery.Node) string {
-	return xmlToMarkdown(node.OutputXML(false), false)
+	return xmlToMarkdown(node.OutputXML(false), false, false)
 }
 
-func xmlToMarkdown(in string, preRender bool) string {
+func xmlToMarkdown(in string, preRender bool, needsSpace bool) string {
 
-	text, err := html2text.FromString(xmlToHtml(in, preRender), html2text.Options{PrettyTables: true, OmitLinks: false})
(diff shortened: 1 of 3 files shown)


pkg/utils/xml2text.go
```diff
@@ -12,24 +12,24 @@ import (
 	"github.com/pkg/errors"
 )
 
-func XmlNodeAsMarkdownPreRender(node *xmlquery.Node) string {
-	return xmlToMarkdown(node.OutputXML(false), true)
+func XmlNodeAsMarkdownPreRender(node *xmlquery.Node, needsSpace bool) string {
```
rhmdnd, line 15: Does this signature need to change? Or will this function always be called with `needsSpace=False`? Based on the changes in profileparser.go, it looks like this will always be called with `true`.

Vincent056: thanks for the review, yes, it will be used here: https://github.com/openshift/compliance-operator/blob/fd0ab5b03eeb70ab45101b8e65518154e11730ef/pkg/utils/parse_arf_result.go#L90

Vincent056: We will call false in the warnings section since we don't want extra space in API path

rhmdnd: Aha - got it, I didn't look at the other review.

Thanks!

jhrozek:
Not related to the code which looks good, but would you mind amending the commit message (`git commit --amend`) with a more detailed explanation? Linking to the issue might be a good idea as well as explaining why we don't want the extra space in that particular case but do want it in others.

Vincent056:
> Not related to the code which looks good, but would you mind amending the commit message (`git commit --amend`) with a more detailed explanation? Linking to the issue might be a good idea as well as explaining why we don't want the extra space in that particular case but do want it in others.

Thanks for the tip, I will do that.

Vincent056:
> Not related to the code which looks good, but would you mind amending the commit message (`git commit --amend`) with a more detailed explanation? Linking to the issue might be a good idea as well as explaining why we don't want the extra space in that particular case but do want it in others.

It's done

rhmdnd:
Looks good. Thanks, Vincent.

Vincent056:
I believe this PR is no longer needed since we are going to implement something better instead for the variable parsing. Related conversation:  https://github.com/openshift/compliance-operator/pull/750#discussion_r761585312

Vincent056:
We are going to need this PR again, for https://github.com/openshift/compliance-operator/pull/750

PR #755, opened 2021-12-08
ScanSetting: Introduce special `@all` role to match all nodes

JAORMX:
This introduces the ability to detect `@all` in the ScanSettings roles.
When used, the scansettingbinding controller will set an empty
`nodeSelector` which will then match all nodes available.

This is mostly intended for non-openshift deployments, and, as expected,
this won't work with remediations.

### DIFF
--- a/deploy/crds/compliance.openshift.io_scansettings_crd.yaml
+++ b/deploy/crds/compliance.openshift.io_scansettings_crd.yaml
@@ -155,7 +155,14 @@ spec:
               <type>'
             type: string
           roles:
-            description: The list of roles to apply node-specific checks to
+            description: "The list of roles to apply node-specific checks to. \n This
+              will be translated to the standard Kubernetes role label `node-role.kubernetes.io/<role
+              name>`. \n It's also possible to specify `@all` as a role, which will
+              run a scan on all nodes by not specifying a node selector as we normally
+              do. The usage of `@all` in OpenShift is discouraged as the operator
+              won't be able to apply remediations unless roles are specified. \n Note
+              that tolerations must still be configured for the opeartor to appropriately
+              schedule scans."
             items:
               type: string
             type: array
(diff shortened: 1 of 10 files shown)


pkg/controller/scansettingbinding/scansettingbinding_controller_test.go
```diff
@@ -735,4 +744,45 @@ var _ = Describe("Testing scansettingbinding controller", func() {
...
+	When("Validating roles", func() {
+		DescribeTable("Should pass the validation",
+			func(roles []string) {
+				ss := &compv1alpha1.ScanSetting{
+					Roles: roles,
+				}
+				err := reconciler.validateRoles(ss)
+				Expect(err).To(BeNil())
+			},
+			Entry("master & worker", []string{"master", "worker"}),
```
Vincent056, line 756: I wonder if we want to test for empty role, `[]string{""}`

Vincent056: Just see that one below.

deploy/crds/compliance.openshift.io_scansettings_crd.yaml
```diff
@@ -155,7 +155,14 @@ spec:
               <type>'
             type: string
           roles:
-            description: The list of roles to apply node-specific checks to
+            description: "The list of roles to apply node-specific checks to. \n This
+              will be translated to the standard Kubernetes role label `node-role.kubernetes.io/<role
+              name>`. \n It's also possible to specify `@all` as a role, which will
+              run a scan on all nodes by not specifying a node selector as we normally
+              do. The usage of `@all` in OpenShift is discouraged as the operator
+              won't be able to apply remediations unless roles are specified. \n Note
```
rhmdnd, line 163: Is this because remediations are role specific?

JAORMX: yes, and it's because of the way the Machine Config Operator works, which requires labels.

Vincent056 (approved):
/lgtm

PR #756, opened 2021-12-09
Inherit scheduling for workloads from Compliance Operator manager

JAORMX:
When the Compliance Operator was first coded, some assumptions were made
that were specific to Openshift. One of these assumptions was the fact
that the `node-role.kubernetes.io/` was more actively used amongst other
distros. An even worst assumption was that the `master` role was always
available. These statements hold true for OpenShift, but not for other
distros like EKS.

This PR removes that assumption by making every workload inherit the
same `nodeSelector` and `tolerations` from the controller manager. This
is done by passing this information to every controller. However, using
the information is optional.

When CO begins, the manager will query it's own pod and fetch this
information. Failure to do so will result in an error.

Since we fetch the operator name and namespace from constants, the
constants were modified to not depend any more on k8sutils. the
aforementioned library will no longer be available in further versions
of operator-sdk, so it's better to move away from it now little by
little.

### DIFF
--- a/cmd/manager/operator.go
+++ b/cmd/manager/operator.go
@@ -9,8 +9,6 @@ import (
 	"runtime"
 	"strings"
 
-	"k8s.io/apimachinery/pkg/util/intstr"
-
 	monitoring "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	monclientv1 "github.com/coreos/prometheus-operator/pkg/client/versioned/typed/monitoring/v1"
 	ocpapi "github.com/openshift/api"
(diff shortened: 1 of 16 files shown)


cmd/manager/operator.go
```diff
@@ -219,8 +220,14 @@ func RunOperator(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
+	getSIErr, si := getSchedulingInfo(ctx, mgr.GetAPIReader())
+	if getSIErr != nil {
+		log.Error(getSIErr, "Getting control plane schedling info")
```
rhmdnd, line 225: nit: scheduling*

deploy/operator.yaml
```diff
@@ -65,11 +67,3 @@ spec:
...
           operator: "Exists"
           effect: "NoSchedule"
-        - key: "node.kubernetes.io/unreachable"
-          operator: "Exists"
-          effect: "NoExecute"
-          tolerationSeconds: 120
-        - key: "node.kubernetes.io/not-ready"
-          operator: "Exists"
-          effect: "NoExecute"
-          tolerationSeconds: 120
```
rhmdnd, line 75: Ok - so by removing these here, we're inheriting a set of global tolerances somewhere?

rhmdnd (commented):
Quick initial pass, but I'd like to try this out on an EKS cluster.

deploy/operator.yaml
Vincent056: was this here to prevent pods from getting evicted on node updates? @JAORMX

Vincent056 (approved):
/lgtm

PR #757, opened 2021-12-10
Add documentation for E2E test

Vincent056:
Added some instructions on how to write the E2E test for our compliance operator.

### DIFF
--- a/doc/e2e-tests.md
+++ b/doc/e2e-tests.md
@@ -1,0 +1,216 @@
+# End to end test documentation
+
+Running e2e ensures various parts of the Compliance Operator are working correctly in a real cluster environment. And we are able to test out new rules and new features using e2e test.
+The tests file directories for e2e are under tests/e2e/
+
+`tests/e2e/e2e-tests.go` - contains all the tests we have, we can add new tests to that file.
+
+## How to add a new test
+
+First, the structure for the test is following:
+
+```go
+	testExecution{
+			Name:       "TestKubeletConfigRemediation",
+			IsParallel: false,
+			TestFn: func(t *testing.T, f *framework.Framework, ctx *framework.Context, mcTctx *mcTestCtx, namespace string) error {
+            ...
+            }
+    }
+```
+We can set `IsParallel` to instruct the test run as Parallel or Serial.
+Usually, we would set "read-only" tests as parallel and anything that modifies the cluster state as serial.
+
+### Build Content image and create `ProfileBundle`
+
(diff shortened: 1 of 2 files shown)


doc/e2e-tests.md
```diff
@@ -0,0 +1,215 @@
...
+			Name:       "TestKubeletConfigRemediation",
+			IsParallel: false,
+			TestFn: func(t *testing.T, f *framework.Framework, ctx *framework.Context, mcTctx *mcTestCtx, namespace string) error {
+            ...
+            }
+    }
+```
+We can set `IsParallel` to instruct the test run as Parallel or Serial.
+
+### Build Content image and create `ProfileBundle`
```
jhrozek, line 23: I'm fine having this documented for now, but I think we should have a Jira task to get rid of using quay images for tests (and I know that I started that dubious practice..). We should move to having all the test content in the repo and build the container images during the test..

This is not criticizing your document, just saying we should at least plan on fixing this part of our e2e testing..

doc/e2e-tests.md
```diff
@@ -0,0 +1,215 @@
...
+```go
+	testExecution{
+			Name:       "TestKubeletConfigRemediation",
+			IsParallel: false,
+			TestFn: func(t *testing.T, f *framework.Framework, ctx *framework.Context, mcTctx *mcTestCtx, namespace string) error {
+            ...
+            }
+    }
+```
+We can set `IsParallel` to instruct the test run as Parallel or Serial.
```
jhrozek, line 21: I think it would be nice to give some best practices as to when can you use parallel or serial. My rule of thumb has been to set 'read-only' tests as parallel and anything that modifies the cluster state as serial.

doc/e2e-tests.md
```diff
@@ -0,0 +1,215 @@
+# End to end test documentation
+
+Running e2e ensures various parts of the Compliance Operator are working correctly in a real cluster environment. And we are able to test out new rules and new features using e2e test.
+The tests file directories for e2e are under tests/e2e/
+
+`tests/e2e/e2e-tests.md` - contains all the tests we have, we can add new tests to that file.
```
jhrozek, line 6: I think you mean to point to the go file?

doc/e2e-tests.md
Vincent056: Thanks for the review, I will create a Jira task for that!

doc/e2e-tests.md
Vincent056: Sounds good! I will add that

JAORMX (approved):
/lgtm

PR #758, opened 2021-12-13
Remove regex checks for url-encoded content

Vincent056:
The regex checks for url-encoded content is not correct, also redundant because `url.QueryUnescape` will return an error if the url-encoded content has illegal characters in it.


jhrozek:
/retest
cluster provisioning error

jhrozek (approved):
/lgtm

PR #760, opened 2021-12-14
Enable defaults per platform

JAORMX:
This enables the Compliance Operator to specify defaults per specific
paltforms. This allows for a better UX based on a set of supported
platforms.

OpenShift
---------

for OpenShift we'll deploy what we're currently doing:

* ScanSettings that target `worker` and `master` nodes.
* ScanSettings that allow the resultserver to schedule on master nodes

EKS
---

For EKS we'll do something a little different:

* ScanSettings will target `@all` so we can schedule to all available
  nodes.

* ScanSettings will inherit the nodeSelector and tolerations from the
  Compliance Operator pod by default.

### DIFF
--- a/cmd/manager/operator.go
+++ b/cmd/manager/operator.go
@@ -60,30 +60,38 @@ func init() {
 	defineOperatorFlags(operatorCmd)
 }
 
-func defineOperatorFlags(cmd *cobra.Command) {
-	cmd.Flags().Bool("skip-metrics", false,
-		"Skips adding metrics.")
+type PlatformType string
 
-	// Add the zap logger flag set to the CLI. The flag set must
-	// be added before calling pflag.Parse().
-	flags := cmd.Flags()
-	flags.AddFlagSet(zap.FlagSet())
-
-	// Add flags registered by imported packages (e.g. glog and
-	// controller-runtime)
-	flags.AddGoFlagSet(flag.CommandLine)
-
-}
+const (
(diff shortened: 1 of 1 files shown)


JAORMX:
/retest

error provisioning cluster

cmd/manager/operator.go
```diff
@@ -60,30 +60,29 @@ func init() {
...
+		PlatformEKS: {
+			"eks",
+		},
+	}
+	defaultRolesPerPlatform = map[PlatformType][]string{
+		PlatformOpenShift: {
+			"master",
+			"worker",
+		},
+		"": {
```
jhrozek, line 83: Some questions:
1. does a generic k8s distro not have master/worker or controlplane/worker roles by default?
2. wouldn't it be better to be explicit for EKS instead of relying on the `""` magic?

JAORMX: 1. No, they don't.
2. I expected `""` to be a default that could be used for other distros too. Seems Azure and GKE don't have those either.

Vincent056: I am wondering for the second question, would `GenericPlatform` improve the readebility?

JAORMX: @Vincent056 I like that idea! Let me do that.

JAORMX: uhm... what if I re-use "UnknownPlatform" here?

Vincent056 (approved):
/lgtm

PR #761, opened 2021-12-14
Fix bugs where kubeletconfig gets deletes when unapplying

Vincent056:
Related Bugzilla link: https://bugzilla.redhat.com/show_bug.cgi?id=2032420

As for right now, let's not support unapplying kubeletconfig remediation, we need to come up with a plan later on that


jhrozek:
The code is clean enough and I can (sort of) reproduce the issue, sort of because I'm getting an INCONSISTENT result, which is even worse from admin's point of view I guess.

But what I'm missing in the PR is some explanation as per why are we hitting this codepath at all: when creating the remediation object in the aggregator, shouldn't we hit the codepath (in `handleRemediation`) that, `if remExists` would create the remediation as applied?

JAORMX (approved):
/lgtm

Thanks for adding unit tests too!

PR #763, opened 2021-12-15
Release v0.1.47

jhrozek:
Let's release a new upstream version

### DIFF
--- a/deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
@@ -159,13 +159,13 @@ metadata:
       ]
     capabilities: Seamless Upgrades
     categories: Monitoring,Security
-    olm.skipRange: '>=0.1.17 <0.1.46'
+    olm.skipRange: '>=0.1.17 <0.1.47'
     operatorframework.io/cluster-monitoring: "true"
     operatorframework.io/suggested-namespace: openshift-compliance
     operators.openshift.io/infrastructure-features: '["disconnected", "fips", "proxy-aware"]'
     repository: https://github.com/openshift/compliance-operator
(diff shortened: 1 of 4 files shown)


jhrozek:
Adding /hold right away so that the PR doesn't get merged without answering one question...

deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
```diff
@@ -1278,14 +1278,16 @@ spec:
                     fieldRef:
                       fieldPath: metadata.name
                 - name: OPERATOR_NAME
-                  value: compliance-operator
+                  valueFrom:
```
jhrozek, line 1281: @JAORMX this change comes from commit eb500977143174759bf677508da36a566e4a4c70 (I think?) but what confuses me is that the change was included in the previous release, not this one. When you did that release, did you hand-edit the CSV to exclude this change? Or do you know why the change appears in this release and wasn't already in the previous one? I'm concerned about upgrading, because locally I can't upgrade from a previous version just by `oc apply -f /path/to/deployment.yaml`:

```
The Deployment "compliance-operator" is invalid: spec.template.spec.containers[0].env[2].valueFrom: Invalid value: "": may not be specified when `value` is not empty
```

Vincent056 (approved):
/lgtm

deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
JAORMX: I didn't hand-edit the csv, just used our normal `generate` target

jhrozek: weird. But we're test the upgrades anyway, worst case we revert.

PR #765, opened 2021-12-16
Add EKS deployment manifest

JAORMX:
This adds a simple deployment manifest for EKS. The idea would be
that this would be generated by kustomize at some point. But let's
have this here while that gets worked out.


Vincent056 (approved):
/lgtm

PR #767, opened 2021-12-20
Initialize inherited selector and tolerations if null

JAORMX:
This prevents the default to kick in and instead uses the actual
inherited value.

This also changes the CRDs to stop defaulting the raw storage scheduling to the master nodes.
Instead relying on the inherited values from the main compliance-operator
deployment.


JAORMX:
Pull-request updated, HEAD is now a21dff9c2ae841857433238dba0fb4dd210ff4b2

Vincent056 (approved):
/lgtm

PR #768, opened 2022-01-10
Add labels to indicate availability for other arches

manojnkumar:
We have done the validation for the ocp4-cis and ocp4-cis-node profiles on ppc64le and are ready to for this operator to be supported on this architecture.

### DIFF
--- a/deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
@@ -169,7 +169,6 @@ metadata:
   namespace: placeholder
   labels:
     operatorframework.io/arch.amd64: supported
-    operatorframework.io/arch.arm64: supported
     operatorframework.io/arch.ppc64le: supported
     operatorframework.io/arch.s390x: supported
 spec:
--- a/deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
@@ -167,6 +167,11 @@ metadata:
     support: Red Hat Inc.
   name: compliance-operator.v0.1.47
   namespace: placeholder
+  labels:
+    operatorframework.io/arch.amd64: supported
+    operatorframework.io/arch.arm64: supported
+    operatorframework.io/arch.ppc64le: supported
+    operatorframework.io/arch.s390x: supported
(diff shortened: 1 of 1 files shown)


mrogers950:
Thanks @manojnkumar ! 
It looks good, but I'm going to put this on hold for the moment because the CSV file will sync to the downstream repos. My concern is that we'll advertise other arches before the downstream arch bundles are ready. I'll unhold once I can confirm that.
/hold

deploy/olm-catalog/compliance-operator/manifests/compliance-operator.clusterserviceversion.yaml
```diff
@@ -167,6 +167,11 @@ metadata:
     support: Red Hat Inc.
   name: compliance-operator.v0.1.47
   namespace: placeholder
+  labels:
+    operatorframework.io/arch.amd64: supported
+    operatorframework.io/arch.arm64: supported
```
yselkowitz, line 172: We'd love to have compliance operator for ARM too, but currently it's not being built (downstream) so it MUST NOT be included here.

mrogers950: Oh good catch, this should be corrected.

manojnkumar: OK, Dropped arm64.

yselkowitz:
The change looks good for downstream, but will this also affect the upstream (community) version as well?  The community build is still amd64-only so it shouldn't have these flags.  Can these be separated somehow?

mrogers950:
> The change looks good for downstream, but will this also affect the upstream (community) version as well? The community build is still amd64-only so it shouldn't have these flags. Can these be separated somehow?

It could be separated. We'd have to code the labels to the CSV when the downstream bundle repo's `render_templates` script runs. The downside is that it's another special piece of the midstream pipeline that we have to maintain. 

I'm not sure how common it is to use OLM to install the upstream version, as opposed to using deploy/ directly. It's provided as a courtesy for upstream, but In general, we treat the CSV as a downstream component (some of the elements like the variable names "RELATED_IMAGE_" are downstream specific).

Overall I'm inclined to have the labels here upstream so they are sourced directly downstream, and to avoid confusion, Add a comment / document that only amd64 builds are provided upstream, even though the labels are there.

gquillar:
[PR-795 ](https://github.com/openshift/compliance-operator/pull/795) to be able to build the `ppc64le/s390x` images has been merged. I think that this PR can be merged now.

PR #769, opened 2022-01-10
Use consistent markdown headings in README.md

rhmdnd:
The README contained a mix of heading syntax. Some using `#` and others
using the underline notation.

This commit updates the headings to use the same format for
consistency.


jhrozek (approved):
/lgtm

jhrozek:
Merging manually, we really don't need a valid bugzilla to fix markdown headers..

PR #772, opened 2022-01-12
Update the TailoredProfile example to include a description

rhmdnd:
A previous commit (68a9a0b73d229663ecea2dd334f92b15d3919938) required
`TailoredProfiles` to have `description` and `title` attributes, but we
forgot to update the example custom resource in deploy/crds. If you try
and apply the custom resource to a deployment, it will fail validation.

This commit updates the example custom resource so that you can apply it
as is.


Vincent056 (approved):
Thanks for fixing the example tailored profile It looks good!

jhrozek (approved):
/lgtm

PR #774, opened 2022-01-15
Bug 2040282: Fix missing checkType on some rules

Vincent056:
Bugzilla link: https://bugzilla.redhat.com/show_bug.cgi?id=2040282

We had issues where some rules have checkType of None instead of the correct type.

We only set checkType to None if the rule does not have any OVAL test, and we check if the rule has OVAL check by looking up the test definition tables. When we created the rule test definition table, those test definitions that have extended definitions are not included. For rules that have multiple checks, they usually have extended definitions, and therefore the checkType of those rules were set to None instead of Node/Platform


jhrozek (commented):
The code looks good and thank you for the quick reply, but could you also extend the existing e2e test with coverage of some rule that was previously broken?

Vincent056:
> The code looks good and thank you for the quick reply, but could you also extend the existing e2e test with coverage of some rule that was previously broken?

Sorry for the late reply, just added the e2e test to it

jhrozek (approved):
/lgtm

Vincent056:
@jhrozek I set the expected checkType wrong, just corrected to Node instead.

jhrozek (approved):
/lgtm

PR #775, opened 2022-01-17
Bug 2040401: fix can't find owner of 98-<pool>-generated-kubelet

Vincent056:
the vsphere-upi installer created some kubelet configuration files `98-<pool>-generated-kubelet`, and they were recognized as `kubeletconfig` generated mc by CO, to fix that we added an additional check to check if the prefix of kubelet mc have `99-`. since the custom `kubeletconfig` will always generate mc with name scheme `99-<pool>-generated-kubelet-<num>`

Related Bugzilla Link: https://bugzilla.redhat.com/show_bug.cgi?id=2040401

### DIFF
--- a/pkg/controller/complianceremediation/complianceremediation_controller.go
+++ b/pkg/controller/complianceremediation/complianceremediation_controller.go
@@ -653,34 +653,35 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteKC(obj *unstructured.U
 	// we need to patch the remediation if there is already a CustomKC present
 	if hasCustomKC {
 
-		kubletMC := &mcfgv1.MachineConfig{}
+		kubeletMC := &mcfgv1.MachineConfig{}
 		kMCKey := types.NamespacedName{Name: kubeletMCName}
 
-		if err := r.client.Get(context.TODO(), kMCKey, kubletMC); err != nil {
+		if err := r.client.Get(context.TODO(), kMCKey, kubeletMC); err != nil {
 			return fmt.Errorf("couldn't get current generated KubeletConfig MC: %w", err)
 		}
-		// WE need to get name of original kubelet config that used to generate this kubeletconfig machine config
-		if len(kubletMC.GetOwnerReferences()) == 0 {
-			return fmt.Errorf("the generated kubelet machine config does not have a owner")
+		// We need to get name of original kubelet config that used to generate this kubeletconfig machine config
+		// if we can't find owner of generated mc, we will create custom kubeletconfig instead
+		if len(kubeletMC.GetOwnerReferences()) != 0 {
+			if kubeletMC.GetOwnerReferences()[0].Kind == "KubeletConfig" {
+				kubeletName := kubeletMC.GetOwnerReferences()[0].Name
+				kubeletConfig := &mcfgv1.KubeletConfig{}
+				kcKey := types.NamespacedName{Name: kubeletName}
+				if err := r.client.Get(context.TODO(), kcKey, kubeletConfig); err != nil {
+					return fmt.Errorf("couldn't get current KubeletConfig: %w", err)
+				}
+				// Set kubelet config name
+				obj.SetName(kubeletConfig.GetName())
+				obj.SetLabels(kubeletConfig.GetLabels())
+				return nil
+			}
 		}
-		kubletName := kubletMC.GetOwnerReferences()[0].Name
-
-		kubeletConfig := &mcfgv1.KubeletConfig{}
-		kcKey := types.NamespacedName{Name: kubletName}
-		if err := r.client.Get(context.TODO(), kcKey, kubeletConfig); err != nil {
-			return fmt.Errorf("couldn't get current KubeletConfig: %w", err)
-		}
-		// Set kublet config name
-		obj.SetName(kubeletConfig.GetName())
-		obj.SetLabels(kubeletConfig.GetLabels())
-		return nil
 	}
 
 	// We will need to create a kubelet config if there is no custom KC
-	kubletName := "compliance-operator-kubelet-" + pool.GetName()
+	kubeletName := "compliance-operator-kubelet-" + pool.GetName()
 
-	// Set kublet config name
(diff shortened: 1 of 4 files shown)


pkg/utils/nodeutils_test.go
```diff
@@ -100,6 +100,54 @@ var _ = Describe("Nodeutils", func() {
 
 		})
 
+		FContext("MachineConfig Pool with no custom KubeletConfig", func() {
```
jhrozek, line 103: I think you want to remove the F

pkg/utils/nodeutils.go
```diff
@@ -84,7 +84,8 @@ func IsMcfgPoolUsingKC(pool *mcfgv1.MachineConfigPool) (bool, string, error) {
 	var currentKCMC string
 	for i := range pool.Spec.Configuration.Source {
 		kcName := pool.Spec.Configuration.Source[i].Name
-		if strings.Contains(kcName, generatedKubelet) {
+		// The prefix has to start with 99 since the kubeletconfig generated machine config will always start with 99
```
jhrozek, line 87: Question: Is there no label or annotation to use instead?

pkg/utils/nodeutils_test.go
Vincent056: Good catch!

pkg/utils/nodeutils.go
Vincent056: No, unfortunately I don't think there are annotation/label we can use to identify if mc is generated by kubeletconfig object, but we can use owner GetOwnerReferences to find out if it is owned by Kubeletconfig object 

```
[vincent@fedora-ws compliance-operator]$ oc get mc 99-master-generated-kubelet -o yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  annotations:
    machineconfiguration.openshift.io/generated-by-controller-version: 5da6b0129a14cb44b9d7218235992933180546bf
  creationTimestamp: "2022-01-17T17:49:25Z"
  generation: 1
  labels:
    machineconfiguration.openshift.io/role: master
  name: 99-master-generated-kubelet
  ownerReferences:
  - apiVersion: machineconfiguration.openshift.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: KubeletConfig
    name: kubelet-config-test-master
    uid: 4377ed1e-e71a-4d5c-83ef-324bd3a302d7
  resourceVersion: "135039"
  uid: da08d8ab-8d48-4eb5-bcc6-146b7c52661a
spec:
```

Vincent056: I updated the code to check if mc is created by Kubeletconfig in complianceremediation_controller.go

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -659,21 +659,24 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteKC(obj *unstructured.U
 		if err := r.client.Get(context.TODO(), kMCKey, kubletMC); err != nil {
 			return fmt.Errorf("couldn't get current generated KubeletConfig MC: %w", err)
 		}
-		// WE need to get name of original kubelet config that used to generate this kubeletconfig machine config
-		if len(kubletMC.GetOwnerReferences()) == 0 {
-			return fmt.Errorf("the generated kubelet machine config does not have a owner")
-		}
-		kubletName := kubletMC.GetOwnerReferences()[0].Name
+		// W need to get name of original kubelet config that used to generate this kubeletconfig machine config
```
jhrozek, line 662: typo: W->We

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -659,21 +659,24 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteKC(obj *unstructured.U
...
+		// if we can't find owner of generated mc, we will create custom kubeletconfig instead
+		if len(kubletMC.GetOwnerReferences()) != 0 {
```
jhrozek, line 664: typo: kublet: kubelet

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -659,21 +659,24 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteKC(obj *unstructured.U
...
+			kubletName := kubletMC.GetOwnerReferences()[0].Name
```
jhrozek, line 665: can we move the variable to the block below where it's used?

pkg/utils/nodeutils.go
jhrozek: thanks, this works for me

jhrozek (commented):
just some very small nits.

rhmdnd (commented):
I'm still piecing together how all this works, but based on the unit tests I think this looks good.

Thanks @Vincent056

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -653,34 +653,35 @@ func (r *ReconcileComplianceRemediation) verifyAndCompleteKC(obj *unstructured.U
...
+		kubeletMC := &mcfgv1.MachineConfig{}
 		kMCKey := types.NamespacedName{Name: kubeletMCName}
 
-		if err := r.client.Get(context.TODO(), kMCKey, kubletMC); err != nil {
+		if err := r.client.Get(context.TODO(), kMCKey, kubeletMC); err != nil {
 			return fmt.Errorf("couldn't get current generated KubeletConfig MC: %w", err)
 		}
-		// WE need to get name of original kubelet config that used to generate this kubeletconfig machine config
-		if len(kubletMC.GetOwnerReferences()) == 0 {
-			return fmt.Errorf("the generated kubelet machine config does not have a owner")
```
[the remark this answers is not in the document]
Vincent056: That's a good question, I removed this error return because, if the mc does not have an owner, it will assume `99-` is not generated by kubeletconfig, and create a kubeletconfig object instead.

PR #784, opened 2022-01-31
Add status descriptors back into CSV

rhmdnd:
When we go to release the compliance-operator, the operator-sdk removes
the `statusDescriptors` in the CSV because we're not annotating them
properly. This commit adds them back. We will need to manually do this
each time we release.


Vincent056 (approved):
Thanks for adding it back in!

Vincent056:
> So the addition is fine, but why are they removed? Is this an operator-sdk bug or something operator-sdk only does in more recent versions?

It is the later, It is hard coded for now, so we are planning to change that after the operator-sdk gets upgraded.

PR #785, opened 2022-01-31
Add network to api-resources we always fetch

Vincent056:
The network resource helps us determine the network type of the platform we're running on. 

We add it to the list of resources we always fetch to be able to write a CPE that will verify if the cluster has an OVN/SDN network type

This PR works in conjunction with: https://github.com/ComplianceAsCode/content/pull/8134

### DIFF
--- a/cmd/manager/scap.go
+++ b/cmd/manager/scap.go
@@ -151,6 +151,10 @@ func (c *scapContentDataStream) FigureResources(profile string) error {
 			ObjPath:  "/apis/config.openshift.io/v1/infrastructures/cluster",
 			DumpPath: "/apis/config.openshift.io/v1/infrastructures/cluster",
 		},
+		{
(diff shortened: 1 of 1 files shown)


JAORMX (approved):
/lgtm
/approve

At some point this should be made configurable, so the container doesn't always try to fetch this. e.g. for EKS or potentially Vanilla Kube (hint hint)

Vincent056:
> /lgtm /approve
> 
> At some point this should be made configurable, so the container doesn't always try to fetch this. e.g. for EKS or potentially Vanilla Kube (hint hint)

Hi Ozz, good to hear back from you! How’s thing going? That’s a good suggestion, thanks for the hints!

jhrozek (approved):
/lgtm

jhrozek (changes requested):
Actually, can you add a changelog blob? I'll merge the other PR so that yours doesn't conflict

PR #786, opened 2022-02-01
Break the release process apart

### LINKED ISSUE #783: operator-sdk is affecting the build process and tagging
We use `make` to carry out our build process, which contains steps that use operator-sdk to update CRDs in the compliance-operator.

During the build process the operator-sdk is removing the `statusDescriptor` portions of the generated CSV. This is because we're not annotating the `statusDescriptor` objects appropriately in the source.

This issue is compounded at release time when the `make release` target builds the updated CSV with the `statusDescriptors` removed, commits the change, and then tags it before pushing the change. We have to go back and re-add `statusDescriptors` to the commit, causing the tag during the release process to differ from the commit with the correct CSV.

You can determine this by looking at tag 0.1.47:

```
$ git show -s v0.1.47                                                                                                                                                                                          commit d1afdfe23cc29ac85c810d82abe6c309cdecd6d3 (tag: v0.1.47, upstream/release-v0.1.47)
Author: Jakub Hrozek <jhrozek@redhat.com>
Date:   Wed Dec 15 17:37:08 2021 +0100
    Release v0.1.47
$ git branch --contains d1afdfe23cc29ac85c810d82abe6c309cdecd6d3
$
```

This will affect people who are checking out releases based on tags and potentially break them.

rhmdnd:
The `make release` target lumped the entire release process together,
which was nice for simplicity, but it doesn't give you, the person
proposing the release, an opportunity to review the changes before
they're proposed.

In the event something is incorrect (e.g., if the operator-sdk did
something we didn't expect), we should have the ability to catch that
and fix it before proposing the release. Otherwise, we will need to fix
upstream branches and update outdated tags.

Fixes #783

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -490,16 +490,22 @@ git-release: package-version-to-tag changelog
 	git add "CHANGELOG.md"
 	git commit -m "Release v$(TAG)"
 	git tag "v$(TAG)"
-	git push origin "v$(TAG)"
-	git push origin "release-v$(TAG)"
 
 .PHONY: fetch-git-tags
 fetch-git-tags:
 	# Make sure we are caught up with tags
 	git fetch -t
 
-.PHONY: release
-release: fetch-git-tags release-tag-image bundle push push-index undo-deploy-tag-image git-release ## Do an official release (Requires permissions)
+.PHONY: prepare-release
+prepare-release: bundle
+
+.PHONY: push-release
+push-release: ## Do an official release (Requires permissions)
+	git push origin "v$(TAG)"
+	git push origin "release-v$(TAG)"
+
+.PHONY: release-images
+release-images:release-tag-image push push-index undo-deploy-tag-image
 	# This will ensure that we also push to the latest tag
 	$(MAKE) push TAG=latest
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Internal Changes
 
--
+- Implemented a multi-step release process to prevent unintentional changes
+  from making their way into the release
+  ([bug](https://github.com/openshift/compliance-operator/issues/783))
 
 ### Deprecations
 
--- a/README.md
+++ b/README.md
@@ -523,3 +523,34 @@ The following is an example release note for a feature with a security note.
     deployment, its infrastructure, or applications. Make sure you send raw
     results to trusted nodes.
 ```
+
+### Proposing Releases
+
+The release process is separated into three phases, with dedicated `make`
+targets. All targets require that you supply the `OPERATOR_VERSION` prior to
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -498,8 +498,14 @@ fetch-git-tags:
 	# Make sure we are caught up with tags
 	git fetch -t
 
-.PHONY: release
-release: fetch-git-tags release-tag-image bundle push push-index undo-deploy-tag-image git-release ## Do an official release (Requires permissions)
```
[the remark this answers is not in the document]
rhmdnd: Allowing for another step where the maintainer can review the git commit/content and the tag before pushing the release?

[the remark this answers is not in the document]
Vincent056: Yes, that's what I was thinking on, maybe a good idea to do that. Reviewing all the changes before the push.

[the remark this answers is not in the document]
rhmdnd: How about I pull the git commit bits up to the prepare-release step and keep the git push commands in the push-release step? That way the process is still three steps.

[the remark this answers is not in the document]
Vincent056: I think that will be great!

Vincent056 (approved):
/lgtm

Makefile
```diff
@@ -490,16 +490,22 @@ git-release: package-version-to-tag changelog
...
 
 .PHONY: fetch-git-tags
 fetch-git-tags:
 	# Make sure we are caught up with tags
 	git fetch -t
 
-.PHONY: release
-release: fetch-git-tags release-tag-image bundle push push-index undo-deploy-tag-image git-release ## Do an official release (Requires permissions)
+.PHONY: prepare-release
+prepare-release: bundle
```
mrogers950, line 500: We should still need to do release-tag-image as a prereq to bundle, instead of in release-images.

mrogers950: We also need to have git-release there too. I think it would suffice for this target to be

`prepare-release: release-tag-image bundle git-release`

mrogers950: but yet `git-release` makes the tag at the end... if you have to fix up a commit, you would need to replace the tag.

I suggest moving

	git commit -m "Release v$(TAG)"
	git tag "v$(TAG)"

to go above the other pushes in `push-release`

Makefile
```diff
@@ -490,16 +490,22 @@ git-release: package-version-to-tag changelog
...
+
+.PHONY: push-release
+push-release: ## Do an official release (Requires permissions)
+	git push origin "v$(TAG)"
+	git push origin "release-v$(TAG)"
+
+.PHONY: release-images
+release-images:release-tag-image push push-index undo-deploy-tag-image
```
mrogers950, line 508: you'll probably need a space after `:`, make is fussy about format

Makefile
rhmdnd: Ok - good points. Pushing a follow up to fix these.

PR #787, opened 2022-02-01
Remove the deploy/eks.yaml deployment

### LINKED ISSUE #781: The deploy/eks.yaml file adds conflicting deployments to the OLM CSV file during a release
If you make a release, the operator-sdk will pickup a recently added deployment (deploy/eks.yaml) and incorporate that deployment into the CSV for the release.

This isn't intended since OLM is an OpenShift-specific concept, and the eks.yaml deployment file is meant for running the operator on AWS EKS.

We should find a way to work around this so that we don't accidentally add a deployment to the OLM CSV when it will break on OpenShift deployments.

rhmdnd:
This deployment was added recently so that we could deploy the
compliance-operator on EKS. Shortly after, we start implementing a Helm
chart for deploying the operator on both OpenShift and EKS, depending on
the values passed into Helm.

Removing this file since it's existence is redundant with what Helm
provides and it breaks `make deploy-local` as well as the OLM artifacts
during the release process.

Fixes #781


JAORMX (approved):

Vincent056:
Thanks! That fixes the issues

Vincent056 (approved):

jhrozek (approved):
/lgtm

PR #788, opened 2022-02-02
use default value for DumpPath to enable runtime resource path customization without modification of ComplianceAsCode rule schema and macros

tmishina:
### Background
Runtime customization expands the capability of compliance operator from standard installation to customized installation such as [HyperShift](https://github.com/openshift/hypershift). In [another PR](https://github.com/openshift/compliance-operator/pull/750), we have enabled customization of API resource collection.  In addition to that, we also need to customize the scan target filepath.

### Solution
There are two solution candidates. One solution is introducing new variable in `yamlfile_value` template of ComplianceAsCode/content as I created in [a PR](https://github.com/ComplianceAsCode/content/pull/7928). However, it duplicates the variable definition in two positions (`.warnings` and `.template.vars`) and it may cause *update anomarly* in future.

In this PR, we leverage the difference of `ObjPath` and `DumpPath` to trick `oscap` command. The API resource collector accesses `ObjPath`, which is locked to a default value. And then it saves the resource to `DumpPath`, which is specified as custom value.

You can find an example in [an issue](https://github.com/tmishina/compliance-operator/issues/2).

### Discussion Point
- I modified `valuesList` from `map[string]string` to `map[string]map[string]string`, but is this good solution? Do we need to introduce some well-defined type for our porpose?
- I will also check the behavior for rules with `jqfilter`.

### DIFF
--- a/cmd/manager/scap.go
+++ b/cmd/manager/scap.go
@@ -153,7 +153,7 @@ func (c *scapContentDataStream) FigureResources(profile string) error {
 		},
 	}
 	effectiveProfile := profile
-	var valuesList map[string]string
+	var valuesList map[string]map[string]string
 
 	if c.tailoring != nil {
 		var selected []utils.ResourcePath
@@ -185,7 +185,7 @@ func (c *scapContentDataStream) FigureResources(profile string) error {
 //
 //  <warning category="general" lang="en-US"><code class="ocp-api-endpoint">/apis/config.openshift.io/v1/oauths/cluster
 //  </code></warning>
-func getPathFromWarningXML(in *xmlquery.Node, valueList map[string]string) []utils.ResourcePath {
+func getPathFromWarningXML(in *xmlquery.Node, valueList map[string]map[string]string) []utils.ResourcePath {
 	DBG("Parsing warning %s", in.OutputXML(false))
 	path, err := utils.GetPathFromWarningXML(in, valueList)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


Vincent056:
I am wondering what is your thought on following solution:
Content Sample Rule:
```
...

{{% set jqfilter = '[.status.history[] | .verified]' %}}
{{% set apipath = '/apis/{{.ns_var}}config.openshift.io/v1/clusterversions/version' %}}
{{% set apipathdefault = '/apis/config.openshift.io/v1/clusterversions/version' %}}

warnings:
- general: |-
    {{{ openshift_filtered_cluster_setting({apipath: jqfilter}) | indent(4) }}}

template:
  name: yamlfile_value
  vars:
    ocp_data: "true"
    filepath: |-
      {{{ openshift_filtered_path(apipath, apipathdefault, jqfilter) }}}
    yamlpath: "[:]"
    entity_check: "all"
    values:
      - value: "true"
        operation: "equals"
```
Then we can modify openshift_filtered_cluster_setting(), and openshift_filtered_path(path, filter) in CaC REPO to take extra parameter openshift_filtered_cluster_setting(apiendpointpath, dumplocation, jqfilter), openshift_filtered_path(path, dumplocation, filter):
```
				ObjPath:  "/apis/customized-ns/config.openshift.io/v1/clusterversions/version",
				DumpPath: "/apis/config.openshift.io/v1/clusterversions/version",
```

Vincent056 (commented):
left a comment, wondering what you think on that

tmishina:
@Vincent056 Thank you for your suggestion, I understood your's and let me try to implement it in ComplianceAsCode.

tmishina:
@Vincent056 I created [another PR on CaC repo](https://github.com/ComplianceAsCode/content/pull/8231). Hope we can continue discussion on the new PR.

Under my trial implementation, I didn't need to modify `openshift_filtered_path()` and therefore I'm wondering wether my understanding for your sample code is correct. I'd appreciate it if you give me comments on the new implementation. Thanks in advance.

tmishina:
Let me close this PR because we have moved to https://github.com/ComplianceAsCode/content/pull/8231 .

Issue #791, opened 2022-02-08
tailored profile state is Error if profile that supposed to be extended does not exist


maratsal:
I am using Argo CD for deployment of compliance operator and for creation of `scansettingbinding` and `tailoredprofile` resources.

When I run argo cd sync I use following annotation with sync retry option, so that sync of application will go through and argo cd will retry sync and create `scansettingbinding` and `tailoredprofile` resources when CRDs are created.
```
    annotations:
      argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
```

However, if tailored profile is set to Error state, then compliance operator never tries to reconcile it and the whole setup doesn't work as scansettingbinding is waiting for tailored profile, but tailored profile is never fixed.

To fix the issue I need to delete tailored profile and scan setting binding and after that run argo cd application sync one more time.

I can play with the delay of argo cd application sync retry, but I think operator should be retrying reconciliation of the tailored profile.

Status of scansettingbinding:
```
status:
  conditions:
  - lastTransitionTime: "2022-02-07T23:27:16Z"
    message: The TailoredProfile referenced has an error and is not usable
    reason: Invalid
    status: "False"
    type: Ready
```

Status of tailored profile:
```
status:
  errorMessage: 'fetching profile to be extended: Profile.compliance.openshift.io
    "ocp4-cis" not found'
  outputRef:
    name: ""
    namespace: ""
  state: ERROR
```

jhrozek:
hmm, maybe we could reconcile on tailored profiles with a custom reconcile map function that would allow us to only reconcile on those TPs that are linked to a binding..let me check if that's possible. Retrying, I don't really like tbh, we'd be potentially flooding the API server undefinitely with requests.

jhrozek:
Yeah, that seems to work, let me polish the code and submit..
/assign

maratsal:
thanks @jhrozek 

I am not a big expert in operators, but I was thinking that operator has to check all resources time to time and try to reconcile them if something is wrong (set to desired state) without any dependency on other resources. As an example in current case operator keeps trying to reconcile scansettingbinding. Please see logs sample.

```
[...]

PR #792, opened 2022-02-08
Reconcile ScanSettingBindings when a TailoredProfile changes

jhrozek:
Modifies the ScanSettingBinding controller to also watch for
TailoredProfile CRs which link to a TP. This allows to reconcile
ScanSettingBindings in case a TailoredProfile is modified, typically
after fixing issues in a TailoredProfile without having to recreate the
TailoredProfile.

Fixes: #791

NOTE: I haven't had the chance to test the e2e addition at all. Feedback is welcome
in the meantime and maybe the tests will even pass :-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixes
 
--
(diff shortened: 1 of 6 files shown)


Vincent056 (approved):
I think the code looks good! Let's wait and see if e2e test passes

PR #795, opened 2022-02-11
Other arch images

gquillar:
Build the images according to the architecture to allow the support of other arches.
The change has been tested to generate ppc64le and s390x images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,13 @@ else ifeq ($(OS_NAME), Darwin)
     SED=sed -i ''
 endif
 
+ARCH?=$(shell uname -m)
+ifeq ($(ARCH), x86_64)
+    OPM_ARCH?=amd64
+else
+    OPM_ARCH?=$(ARCH)
+endif
+
 ifeq ($(RUNTIME), podman)
     LOGIN_PUSH_OPTS="--tls-verify=false"
 else ifeq ($(RUNTIME), docker)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -75,14 +82,14 @@ export OPERATOR_NAMESPACE?=openshift-compliance
...
 SDK_VERSION?=v0.18.2
 ifeq ($(OS_NAME), Linux)
-    OPERATOR_SDK_URL=https://github.com/operator-framework/operator-sdk/releases/download/$(SDK_VERSION)/operator-sdk-$(SDK_VERSION)-x86_64-linux-gnu
+    OPERATOR_SDK_URL=https://github.com/operator-framework/operator-sdk/releases/download/$(SDK_VERSION)/operator-sdk-$(SDK_VERSION)-$(ARCH)-linux-gnu
 else ifeq ($(OS_NAME), Darwin)
     OPERATOR_SDK_URL=https://github.com/operator-framework/operator-sdk/releases/download/$(SDK_VERSION)/operator-sdk-$(SDK_VERSION)-x86_64-apple-darwin
 endif
 
-OPM_VERSION=v1.15.2
+OPM_VERSION=v1.20.0
```
jhrozek, line 90: @mrogers950 do you suppose this change could bite us while we are using the super outdated operator-sdk as we are?

jhrozek (commented):
Mostly looks harmless, unsure if bumping OPM would work well.
That said, I tested `make index-image` and `make test-index-image` and those seem to work fine.

Makefile
mrogers950: I think this will be OK.

jhrozek: LGTM, then

gquillar:
@jhrozek still pending though your LGTM. What is expected?

jhrozek (approved):
/lgtm

skuethe:
looking forward for this PR to hit the (next) release, so we can start using it. Any idea when that will be?

PR #798, opened 2022-02-25
e2e: Fix TestInconsistentResult

gquillar:
The TestInconsistentResult e2e test is relying on an annotation to determine the test result. In the case of 2 worker nodes, the expected annotation is not correctly built, so the test fails.

This patch fixes the expected annotation.


jhrozek (approved):
/lgtm
oops, nice catch.

PR #800, opened 2022-03-01
feat: variable substitution for jqfilter directive

tmishina:
This pull request provides one more customization capability to Compliance Operator in addition to the previous PR #750.

#750 supports customization of API resource path, and this PR supports customization of jqfilter.
When this PR has been merged, anyone can write a  jqfilter string with variables in a [ComplianceAsCode](https://github.com/ComplianceAsCode/content) rule like below.

```
{{% set custom_jqfilter = '[.data."{{.var_openshift_kube_apiserver_config_data_name}}" | fromjson]' %}}
{{% set default_jqfilter = '[.data."config.yaml" | fromjson]' %}}
{{% set custom_api_path = '/api/v1/namespaces/{{.var_openshift_kube_apiserver_namespace}}/configmaps/{{.var_openshift_kube_apiserver_config}}' %}}
{{% set default_api_path = '/api/v1/namespaces/openshift-kube-apiserver/configmaps/config' %}}
{{% set dump_path = default_api_path ~ ',' ~ default_jqfilter ~ ',' ~ custom_jqfilter %}}
```
In this example, `{{.var_openshift_kube_apiserver_config_data_name}}` in the first line will be substituted with the value of variable `var_openshift_kube_apiserver_config_data_name`. 

This is a concrete example rule for [HyperShift](https://github.com/openshift/hypershift). Configuration data of OpenShift API Server in a HyperShift environment is stored in `.data."config.json"` while that of standard OpenShift API Server is in `.data."config.yaml"`.

By using this feature, we can support more platforms like HyperShift and other managed environments in addition to standard OpenShift installations.

### DIFF
--- a/cmd/manager/scap.go
+++ b/cmd/manager/scap.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"html"
 	"io"
 	"io/ioutil"
 	"os"
@@ -220,7 +221,7 @@ func getResourcePaths(profileDefs *xmlquery.Node, ruleDefs *xmlquery.Node, profi
 				if val.SelectAttr("selector") == "" {
 					// It is not an enum choice, but a default value instead
 					if strings.HasPrefix(variable.SelectAttr("id"), valuePrefix) {
-						valuesList[strings.TrimPrefix(variable.SelectAttr("id"), valuePrefix)] = val.OutputXML(false)
+						valuesList[strings.TrimPrefix(variable.SelectAttr("id"), valuePrefix)] = html.UnescapeString(val.OutputXML(false))
 					}
 				}
 			}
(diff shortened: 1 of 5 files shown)


tmishina:
@Vincent056 sorry but could you wait for a while when I complete revising the code?
I found that current code causes issue when a jqfilter string contains double quote characters. I have already fixed the issue and added unit test, and I'm now on the final check.

tmishina:
@Vincent056 I fixed the error.
1. use `html.UnescapeString` to unescape double quote characters
2. use `text/template` instead of `html/template` to suppress  encoding twice (`html/template` encodes `"` to `&amp;&#34`, not just to `&#34`)

The second change may break backward compatibility (while no error happens at `make test-unit`), so please review my change, thanks.

Vincent056:
> /retest

@tmishina We have some issue with the CI

Vincent056:
@tmishina the CI issue is fixed, could you rebase it? Thanks

Vincent056 (approved):
/lgtm

PR #801, opened 2022-03-03
bug #2059611: scan: Only scan Linux nodes with node scans

jhrozek:
The node scans can only be scheduled on Linux nodes because the OpenScap
scanner is compiled as a Linux binary. Remediations work fine because
even if a pool consists of multiple OSs, the machineConfigPool only
contains Linux nodes.

Most containers (e.g. aggregator, resultserver), when running on OCP, are
already scheduled on master nodes by default anyway, which only run Linux.

JIRA: [OCPBUGSM-41212](https://issues.redhat.com/browse/OCPBUGSM-41212)

### DIFF
--- a/pkg/controller/compliancescan/scantype.go
+++ b/pkg/controller/compliancescan/scantype.go
@@ -85,8 +85,10 @@ func (nh *nodeScanTypeHandler) getTargetNodes() ([]corev1.Node, error) {
 	case compv1alpha1.ScanTypePlatform:
 		return nodes.Items, nil // Nodes are only relevant to the node scan type. Return the empty node list otherwise.
 	case compv1alpha1.ScanTypeNode:
+		// we only scan Linux nodes
+		nodeScanSelector := map[string]string{"kubernetes.io/os": "linux"}
 		listOpts := client.ListOptions{
-			LabelSelector: labels.SelectorFromSet(nh.scan.Spec.NodeSelector),
+			LabelSelector: labels.SelectorFromSet(labels.Merge(nh.scan.Spec.NodeSelector, nodeScanSelector)),
 		}
 
 		if err := nh.r.client.List(context.TODO(), &nodes, &listOpts); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controller/compliancescan/scantype.go
```diff
@@ -85,8 +85,10 @@ func (nh *nodeScanTypeHandler) getTargetNodes() ([]corev1.Node, error) {
 	case compv1alpha1.ScanTypePlatform:
 		return nodes.Items, nil // Nodes are only relevant to the node scan type. Return the empty node list otherwise.
 	case compv1alpha1.ScanTypeNode:
+		// we only scan Linux nodes
+		nodeScanSelector := map[string]string{"kubernetes.io/os": "linux"}
```
jhrozek, line 89: @rhmdnd does EKS contain these node labels?

jhrozek: If not, we can pass down the node selector up from the operator level which detects the platform. This is just much simpler.

jhrozek: btw upstream kube does have those labels (verified with SPO suite which just installs upstream precompiled binaries)

jhrozek:
@pdhamdhe for your testing pleasure..
marking as hold so that QE can test the PR out and we don't merge it accidentally
/hold

JAORMX (approved):
lgtm

Both Docker Desktop and `kubeadm` use these labels. They're standard

pkg/controller/compliancescan/scantype.go
Vincent056: It seems like `"kubernetes.io/os": "linux"` are commonly used

Vincent056: https://aws.github.io/aws-eks-best-practices/windows/docs/scheduling/

Vincent056 (approved):
I saw similar CI failing on `TestSuiteWithContentThatDoesNotMatch`

jhrozek:
/retest
now we should have a new broken content image..

jhrozek:
Probably a race, but the test that had been failing earlier is now succeeding
/retest

pkg/controller/compliancescan/scantype.go
rhmdnd: Thanks for following up here @Vincent056

PR #803, opened 2022-03-05
Increase e2e timeout

Vincent056:
We added one more parallel test, and some tests like `TestSuiteWithContentThatDoesNotMatch` are getting timeout because of this, let's increase the timeout to 30 mins instead.


jhrozek:
/retest
now we should have a new broken content image..

PR #810, opened 2022-03-22
Make sure we convert the release version to tag in all release targets

rhmdnd:
We recently broke apart the release process, but we forgot to add the
conversion of the release version to the tag in the last two steps.

This commit adds that step to the last two parts of the release process
so it doesn't fail or assume 'latest' for the release version.


jhrozek (approved):
/lgtm

jhrozek:
/retest
can't install cluster

PR #814, opened 2022-04-05
BUG 2071854: Fix unpausing machine config pool too soon

Vincent056:
Fix issues of unpausing machine config pool too soon after applying remediations,
added a check to check if kubeletconfig has been fully rendered into machine config
before unpause affected pool. [1]
([1]https://bugzilla.redhat.com/show_bug.cgi?id=2071854).


pkg/controller/compliancesuite/compliancesuite_controller.go
```diff
@@ -513,10 +513,18 @@ func (r *ReconcileComplianceSuite) reconcileRemediations(suite *compv1alpha1.Com
 			return reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
 		}
 	}
-
```
rhmdnd, line 516: Unrelated whitespace change?

pkg/utils/nodeutils.go
```diff
@@ -115,6 +121,80 @@ func IsMcfgPoolUsingKC(pool *mcfgv1.MachineConfigPool) (bool, string, error) {
...
+		return false, fmt.Errorf("kubeletconfig %s is not subset of %s", currentKCMCName, decodedKC)
+	}
+}
+
+func GetKCFromMC(mc *mcfgv1.MachineConfig, client runtimeclient.Client) (*mcfgv1.KubeletConfig, error) {
+	if mc == nil {
+		return nil, fmt.Errorf("machine config is nil")
+	}
+	if len(mc.GetOwnerReferences()) != 0 {
+		if mc.GetOwnerReferences()[0].Kind == "KubeletConfig" {
```
rhmdnd, line 185: Of this list of references, we only check the first? Or does this mean it's safe to assume there is only ever one element in the list?

pkg/utils/compare_json.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	table.Render()
+}
+
+// JSONIsSubset checks if a is a subset json of b
+func JSONIsSubset(a, b []byte) (bool, *JSONDiff, error) {
+	return jsonIsSubsetR(a, b, nil, nil)
+}
+
+func jsonIsSubsetR(a, b []byte, diff *JSONDiff, prefix interface{}) (bool, *JSONDiff, error) {
```
rhmdnd, line 44: There are a lot of cases in here. Can we add some unit tests or is it covered in an integration test?

pkg/utils/nodeutils.go
```diff
@@ -115,6 +121,80 @@ func IsMcfgPoolUsingKC(pool *mcfgv1.MachineConfigPool) (bool, string, error) {
 	return true, currentKCMC, nil
 }
 
+func AreKubeletConfigsRendered(pool *mcfgv1.MachineConfigPool, client runtimeclient.Client) (bool, error) {
```
rhmdnd, line 124: Similar comment as above about testing.

CHANGELOG.md
```diff
@@ -31,6 +31,22 @@ Versioning](https://semver.org/spec/v2.0.0.html).
...
+
+### Enhancements
+
+- 
+
+### Fixes
+
+- Fix issues of unpausing machine config pool too soon after apply remediations,
+  added a check to check if kubelet has been fully rendered into machine config
+  before unpause affected pool. [1]
```
rhmdnd, line 44: nit: I'd try and include the outcome here if possible (helps justify why we made the change).

For example:

The compliance-operator now waits until the kubelet configuration is fully rendered before unpausing the pool to avoid kubelets getting stuck in a `NotReady` state.

pkg/utils/nodeutils.go
Vincent056: It is safe to assume there is only ever one element in the list as only one KC corresponds to one MC

pkg/utils/compare_json.go
Vincent056: @rhmdnd I just added integration test here: https://github.com/openshift/compliance-operator/pull/814/files#diff-0a9f36cc9b36d2e583d93013174fdc0bfedafad46ef1e2075effcbf8bbb87e8dR604

I am not sure if we could add an e2e test  for this one since it will be hard to mimic the delay for rendering KubeletConfig, but I think having this test should be enough. This fix will greatly improve the stability of our compliance operator.

pkg/controller/complianceremediation/complianceremediation_controller.go
```diff
@@ -240,6 +240,7 @@ func (r *ReconcileComplianceRemediation) reconcileRemediation(instance *compv1al
 	}
 
 	if instance.Spec.Apply {
+		fmt.Println("something")
```
jhrozek, line 243: this leftover should be removed

pkg/utils/nodeutils.go
```diff
@@ -115,6 +121,79 @@ func IsMcfgPoolUsingKC(pool *mcfgv1.MachineConfigPool) (bool, string, error) {
...
+	isUsingKC, currentKCMCName, err := IsMcfgPoolUsingKC(pool)
+	if err != nil {
+		return false, fmt.Errorf("failed to check if pool %s is using a custom kubelet config: %w", pool.Name, err)
+	}
+	if !isUsingKC || currentKCMCName == "" {
+		return true, nil
+	}
+	// if the pool is using a custom kubelet config, check if the kubelet config is rendered
+	kcmcfg := &mcfgv1.MachineConfig{}
+	err = client.Get(context.TODO(), types.NamespacedName{Name: currentKCMCName, Namespace: pool.Namespace}, kcmcfg)
```
jhrozek, line 135: This is a nit, but a pool is not namespaced, but cluster-wide, so you can just remove the `Namespace` parameter and not use the `pool.Namespace` attribute, it will always be an empty string anyway.

pkg/utils/nodeutils.go
```diff
@@ -115,6 +121,79 @@ func IsMcfgPoolUsingKC(pool *mcfgv1.MachineConfigPool) (bool, string, error) {
...
+
+	encodedKC, err := jsonpath.Get("storage.files[0].contents.source", obj)
+	if err != nil {
+		return false, fmt.Errorf("failed to get encoded kubelet config from machine config %s: %w", currentKCMCName, err)
+	}
+	encodedKCStr := encodedKC.(string)
+	if encodedKCStr == "" {
+		return false, fmt.Errorf("encoded kubeletconfig %s is empty", currentKCMCName)
+	}
+	mcPayloadPrefix := `data:text/plain,`
```
jhrozek, line 159: can you make this a const?

pkg/utils/nodeutils.go
```diff
@@ -115,6 +121,79 @@ func IsMcfgPoolUsingKC(pool *mcfgv1.MachineConfigPool) (bool, string, error) {
...
+	encodedKCStr = strings.TrimPrefix(encodedKCStr, mcPayloadPrefix)
```
jhrozek, line 160: If we are expecting the prefix, shouldn't we be more defensive and error out if the prefix is not there?

pkg/utils/nodeutils.go
```diff
@@ -115,6 +121,79 @@ func IsMcfgPoolUsingKC(pool *mcfgv1.MachineConfigPool) (bool, string, error) {
...
+	}
+
+	isSubset, diff, err := JSONIsSubset(kc.Spec.KubeletConfig.Raw, []byte(decodedKC))
+	if err != nil {
+		return false, fmt.Errorf("failed to check if kubeletconfig %s is subset of %s: %w", currentKCMCName, decodedKC, err)
+	}
+	if isSubset {
+		return true, nil
+	} else {
+		PrintJSONDiff(diff)
```
jhrozek, line 174: hmm, on one hand I like printing the diff, on the other hand shouldn't we go through the logger interface rather than printing the diff directly to stdout? If you don't like to add logging to this function, returning the diff up or writing it into a io.Writer or smh would both work better IMO than an utility function printing to stdout..

jhrozek (commented):
See a couple of nits inline, mostly looks good!

pkg/utils/nodeutils.go
Vincent056: that's a good catch!

jhrozek:
You need to rebase atop origin/master to get rid of the merge conflict or else CI won't run..

jhrozek (approved):
/lgtm
thank you, all my changes have been addressed

PR #815, opened 2022-04-08
Modify the api-resource-collector to fetch network operator resources

rhmdnd:
The CIS benchmarks recommends using a CNI that supports network
policies. Previously, this was a manual check in the profile, but it's
actually something we can check by querying the K8S operator API.

This commit adds the network resource to the operator.openshift.io API
group so that the api-resource container can fetch that resource. This
change is required to automate this check and used in:

  https://github.com/ComplianceAsCode/content/pull/8524

Partial-Fix: https://bugzilla.redhat.com/show_bug.cgi?id=2072431

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,12 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Enhancements
 
--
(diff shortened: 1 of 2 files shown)


jhrozek (approved):
/lgtm

PR #817, opened 2022-04-12
e2e: Only check for the coreos-vsyscall-kernel-argument on x86

jhrozek:
We have a check that makes sure that even an INFO-level check generates
a remediation. However, the only INFO-level check that is available in
the content is x86-specific.

This patch makes that test only run on x86 only.


Vincent056 (approved):
/lgtm

PR #818, opened 2022-04-13
Fetch node resources

Vincent056:
Added node resource to the list of resources we always fetch so that arch CPEs will be evaluated appropriately.


jhrozek (approved):
/lgtm

PR #819, opened 2022-04-14
Update api-resource-collector cluster role to check if GitOps operator is installed

Vincent056:
Added permission for `api-resource-collector` to fetch network resources for the `gitopsservices.pipelines.openshift.io` API group, so that we can if the cluster is using GitOps operator


jhrozek (approved):
/lgtm

PR #820, opened 2022-04-21
Bug 2075029: scans: Delete scan pods and aggregator when they're done

jhrozek:
We didn't delete scan pods and the aggregator after a scan was done
unless the delete was forced by the scan object being deleted. This
appears to be breaking the cluster-autoscaler which can't drain pods
from nodes that are not backed by a replicating controller.

Since there appears to be no reason to keep the pods around unless the
debug=true option is set, let's remove those pods when the scan reaches
done and only keep the other resources around that are needed for
eventually re-running the scan.

Jira: [OCPBUGSM-43262](https://issues.redhat.com//browse/OCPBUGSM-43262)

### DIFF
--- a/pkg/controller/compliancescan/compliancescan_controller.go
+++ b/pkg/controller/compliancescan/compliancescan_controller.go
@@ -454,24 +454,30 @@ func (r *ReconcileComplianceScan) phaseDoneHandler(h scanTypeHandler, instance *
 	var err error
 	logger.Info("Phase: Done")
 
-	// We need to remove resources before doing a re-scan
-	if doDelete || instance.NeedsRescan() {
-		logger.Info("Cleaning up scan's resources")
+	// the scan pods and the aggregator are done at this point and can be cleaned up regardless
+	// of the doDelete flag, unless we are running in debug mode and thus requested them to stay
+	// around for later inspection
+	if instance.Spec.Debug == false || instance.NeedsRescan() {
 		// Don't try to clean up scan-type specific resources
 		// if it was an unknown scan type
 		if h != nil {
 			if err := h.cleanup(); err != nil {
+				logger.Error(err, "Cannot clean up scan pods")
 				return reconcile.Result{}, err
 			}
 		}
 
-		if err := r.deleteResultServer(instance, logger); err != nil {
-			logger.Error(err, "Cannot delete result server")
-			return reconcile.Result{}, err
-		}
-
 		if err := r.deleteAggregator(instance, logger); err != nil {
 			logger.Error(err, "Cannot delete aggregator")
+			return reconcile.Result{}, err
+		}
+	}
+
+	// We need to remove resources before doing a re-scan
+	if doDelete || instance.NeedsRescan() {
+		logger.Info("Cleaning up scan's resources")
+		if err := r.deleteResultServer(instance, logger); err != nil {
+			logger.Error(err, "Cannot delete result server")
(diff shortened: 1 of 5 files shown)


Vincent056 (approved):
/lgtm
sorry for the late review, thanks for the PR

pkg/controller/compliancescan/compliancescan_controller.go
```diff
@@ -454,24 +454,30 @@ func (r *ReconcileComplianceScan) phaseDoneHandler(h scanTypeHandler, instance *
...
-		logger.Info("Cleaning up scan's resources")
+	// the scan pods and the aggregator are done at this point and can be cleaned up regardless
+	// of the doDelete flag, unless we are running in debug mode and thus requested them to stay
+	// around for later inspection
+	if instance.Spec.Debug == false || instance.NeedsRescan() {
 		// Don't try to clean up scan-type specific resources
 		// if it was an unknown scan type
 		if h != nil {
 			if err := h.cleanup(); err != nil {
+				logger.Error(err, "Cannot scan pods")
```
rhmdnd, line 465: For my understanding, we're saying we can't scan the pods because we weren't able to cleanup after ourselves, right?

pkg/controller/compliancescan/compliancescan_controller_test.go
```diff
@@ -320,17 +320,35 @@ var _ = Describe("Testing compliancescan controller phases", func() {
...
+				Expect(pods.Items).To(BeEmpty())
+
+				// but other resources should be preserved
+				var secrets corev1.SecretList
+				reconciler.client.List(context.TODO(), &secrets)
+				Expect(err).To(BeNil())
+				Expect(secrets.Items).ToNot(BeEmpty())
 			})
 		})
 		Context("with delete flag on", func() {
```
rhmdnd, line 354: Should we add a test for the case where debug is on?

pkg/controller/compliancescan/compliancescan_controller.go
mrogers950: This error should probably be "Cannot clean up"

jhrozek:
Thanks for the reviews. I'll add the debug test and fix the other small issues tomorrow.
Removing lgtm until then

pkg/controller/compliancescan/compliancescan_controller.go
jhrozek: Copy-paste or something. Oops.

jhrozek:
/test e2e-aws
metrics flake

pkg/controller/compliancescan/compliancescan_controller_test.go
```diff
@@ -303,66 +324,83 @@ var _ = Describe("Testing compliancescan controller phases", func() {
...
+				createFakeScanPods(reconciler, compliancescaninstance.Name, nodeinstance1.Name, nodeinstance2.Name)
+				createFakeRsSecret(reconciler, compliancescaninstance.Name)
 
 				// Set state to DONE
 				compliancescaninstance.Status.Phase = compv1alpha1.PhaseDone
 				reconciler.client.Status().Update(context.TODO(), compliancescaninstance)
 			})
 			It("Should return success & clean up resources", func() {
-				var pods corev1.PodList
+				fmt.Println("SHOULD DELETE")
```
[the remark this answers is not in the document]
jhrozek: lol yeah

PR #822, opened 2022-04-23
Fix memory pointer crash loop

Vincent056:
Fix compliance operator crashloop issue when there is `MachineConfigPool` that does not have `NodeSelector`, this pr will check if there is `NodeSelector` in a `MachineConfigPool` before comparing the match.

Related customer case: https://access.redhat.com/support/cases/#/case/03199627

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   before unpause affected pool. [1]
   ([1]https://bugzilla.redhat.com/show_bug.cgi?id=2071854).
 
+- Fixes memory pointer crashloop when there is invalid `MachineConfigPool` in the
+  cluster. 
+  [Cusomter Case](https://access.redhat.com/support/cases/#/case/03199627) for more
+  information
+
 ### Internal Changes
 
 - Added node resource to the list of resources we always fetch so that arch CPEs will
--- a/pkg/utils/nodeutils.go
+++ b/pkg/utils/nodeutils.go
@@ -208,7 +208,15 @@ func McfgPoolLabelMatches(nodeSelector map[string]string, pool *mcfgv1.MachineCo
 	if nodeSelector == nil {
 		return false
 	}
+
+	if pool.Spec.NodeSelector == nil {
+		return false
+	}
 	// TODO(jaosorior): Make this work with MatchExpression
+	if pool.Spec.NodeSelector.MatchLabels == nil {
+		return false
+	}
+
 	return reflect.DeepEqual(nodeSelector, pool.Spec.NodeSelector.MatchLabels)
 }
 
--- a/pkg/utils/nodeutils_test.go
+++ b/pkg/utils/nodeutils_test.go
@@ -227,4 +227,49 @@ var _ = Describe("Nodeutils", func() {
 			Entry("@all", "@all", map[string]string{}),
 		)
 	})
+
+	Context("MachineConfig Pool with no node selector", func() {
+		targetNodeSelector := map[string]string{
+			"test-node-role": "",
+		}
+
+		mcp := &mcfgv1.MachineConfigPool{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "MachineConfigPool",
+				APIVersion: "v1",
+			},
+			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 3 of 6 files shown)


pkg/utils/nodeutils.go
```diff
@@ -121,6 +121,14 @@ func McfgPoolLabelMatches(nodeSelector map[string]string, pool *mcfgv1.MachineCo
...
+	if pool.Spec.NodeSelector == nil {
+		return false
+	}
+
+	if pool.Spec.NodeSelector.MatchLabels == nil {
+		return false
+	}
+
 	return reflect.DeepEqual(nodeSelector, pool.Spec.NodeSelector.MatchLabels)
 }
```
JAORMX, line 133: It would be ideal to come up with a unit test for this to ensure we don't regress

JAORMX (changes requested):

pkg/utils/nodeutils.go
```diff
@@ -209,6 +209,14 @@ func McfgPoolLabelMatches(nodeSelector map[string]string, pool *mcfgv1.MachineCo
 		return false
 	}
 	// TODO(jaosorior): Make this work with MatchExpression
```
rhmdnd, line 211: nit: I would move this down to line 220 so that it's closest to the original code it was written for.

tests/e2e/helpers.go
```diff
@@ -203,6 +225,15 @@ func (c *mcTestCtx) ensureE2EPool() {
 	c.trackPool(pool)
 }
 
+// This will creat a new machine config sub pool with one random node from worker pool to speed up the test.
```
rhmdnd, line 228: nit: create*

Looks like the actual spelling error came from line 219 above though (we can fix that separately).

pkg/utils/nodeutils_test.go
```diff
@@ -227,4 +227,29 @@ var _ = Describe("Nodeutils", func() {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "invalidpool",
+			},
+			Spec: mcfgv1.MachineConfigPoolSpec{
+				Paused: false,
+			},
+		}
+		It("Get correct custom KC name", func() {
+			isMatchingPool := utils.McfgPoolLabelMatches(targetNodeSelector, mcp)
+			Expect(isMatchingPool).To(BeFalse())
```
rhmdnd, line 250: This evaluates to `false` because the `MachineConfigPool` we're defining on line 236 doesn't have a `NodeSelector` key, correct?

Do we also want to have a case where we ensure there is a `NodeSelector` key but `MatchLabels` is `nil`?

rhmdnd (changes requested):

CHANGELOG.md
```diff
@@ -32,6 +32,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   before unpause affected pool. [1]
   ([1]https://bugzilla.redhat.com/show_bug.cgi?id=2071854).
 
+- Fixes memory pointer crashloop when there is invalid `MachineConfigPool` in the
+  cluster. 
+  [Cusomter Case](https://access.redhat.com/support/cases/#/case/03199627) for more
```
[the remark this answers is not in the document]
Vincent056: I should create a BZ, it doesn't have one only the customer case right now

tests/e2e/helpers.go
```diff
@@ -184,6 +185,17 @@ func (c *mcTestCtx) cleanupTrackedPools() {
 	}
 }
 
+func (c *mcTestCtx) cleanupTrackedInvalidPools() {
+	for _, p := range c.invalidPools {
+		// delete the pool
+		E2ELogf(c.t, "Removing pool %s\n", p.Name)
```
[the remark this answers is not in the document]
Vincent056: I think if the test fails, it will show in the log

CHANGELOG.md
```diff
@@ -32,6 +32,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   before unpause affected pool. [1]
   ([1]https://bugzilla.redhat.com/show_bug.cgi?id=2071854).
 
+- Fixes memory pointer crashloop when there is invalid `MachineConfigPool` in the
+  cluster.
```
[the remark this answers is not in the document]
rhmdnd: Ack - thanks!

PR #836, opened 2023-10-12
Correct "ScanSettings" typo

gojeaqui:


rhmdnd:
Thanks for the patch.

This repository was superseded by https://github.com/ComplianceAsCode/compliance-operator/tree/master and it looks like this is applicable there, too.

https://github.com/ComplianceAsCode/compliance-operator/blob/master/doc/usage.md?plain=1#L54

Do you mind opening a fix there instead?

JAORMX (approved):

PR #841, opened 2026-02-23
fix: typo in update event

stehessel:
Fix typo `updatd` -> `updated`.


pkg/controller/scansettingbinding/scansettingbinding_controller.go
guzalv, line 1: This repo is not maintained anymore, nowadays development is done in https://github.com/ComplianceAsCode/compliance-operator. Should the fix be applied there instead? https://github.com/ComplianceAsCode/compliance-operator/blob/30241bee7956b5e2cac904517e6364516be513aa/pkg/controller/scansettingbinding/scansettingbinding_controller.go#L282

stehessel: yes, thanks for the hint.

stehessel: https://github.com/ComplianceAsCode/compliance-operator/pull/1090

